PR #24, opened 2013-11-26
Adding FILE_APPEND to Local->update()

dragoonis:


prisis:
@frankdejonge can we add this setting to the update config?

PR #345, opened 2014-12-11
Added support for glob patterns on the filesystems that support it

Markcial:
Just an idea about giving more flexibility to the filesystems if they support different kind of operations by replacing the boolean parameter in the `listContents` by another one that uses bit mask operations, this way you can add glob behaviour to the local filesystem for example, and exceptions can be thrown if the mode is not supported by the filesystem if needed.

Feel free to review or debate about this, i think that could be a nice addition.

Regards

### DIFF
--- a/src/Adapter/AbstractFtpAdapter.php
+++ b/src/Adapter/AbstractFtpAdapter.php
@@ -217,8 +217,10 @@ abstract class AbstractFtpAdapter extends AbstractAdapter
     /**
      * {@inheritdoc}
      */
-    public function listContents($directory = '', $recursive = false)
+    public function listContents($directory = '', $mode = 0)
     {
+        $recursive = $mode & self::MODE_RECURSIVE;
+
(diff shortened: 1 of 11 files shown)


Markcial:
do we are the only ones that use the glob finding in filesystems? then feel free to close this PR, we will use the patch in our library only.

stevleibelt:
@Markcial you've provided a fix for the issue https://github.com/thephpleague/flysystem/issues/503 before it was created :-).

I like the idea as well as the implementation since it should fit all general needs (at least I could remove my filter in some projects :-)).

Markcial:
Hi @stevleibelt i ended closing this issue because seemed like nobody cared about this.
As a solution on several projects we ended creating a custom FilesystemManager with a custom glob behaviour. I don't know if that could work for you, anyways i fell like i don't care anymore about this. If you want to re-fork this changes and create PR i will help you with anything, but for anything else i feel powerless. Thanks for your interest

PR #346, opened 2014-12-11
close #338 mount manager adds reference to the filesystem that listed contents

### LINKED ISSUE #338: Files listed with listContents doesnt provide information about from which filesystem belong to
Hi, when you list files from a single filesystem is ok to not know where does it belong to, but when you do operations between filesystems would be nice to know from which had come.

This way when you do operations between filesystems you might have a better way to know where and from copy or move.

This could be a optional field in the array returned in the listcontents, like `filesystem => "alias"` or `virtualPath => fs://relative/path`

I am currently working on this in my fork. If i get a good solution i will push the branch so you can check if all of you are agree with the solution.

Thanks

Markcial:
mount manager decorates the results with a parameter detailing from which filesystem the file belongs to

### DIFF
--- a/src/MountManager.php
+++ b/src/MountManager.php
@@ -29,7 +29,6 @@ use LogicException;
  * @method bool delete($path)
  * @method bool deleteDir($dirname)
  * @method bool createDir($dirname, $config = [])
- * @method array listContents($directory = '', $recursive = false)
  * @method array listFiles($directory = '', $recursive = false)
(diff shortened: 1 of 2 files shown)


Markcial:
ping

frankdejonge:
@Markcial looks good, can you rebase? There's a conflict. Ping me and I'll merge it :+1:

Markcial:
Ok, you will have the rebase done before noon.

PR #355, opened 2014-12-31
Add $arguments for after event

GrahamCampbell:
Replaces #353.

### DIFF
--- a/src/Event/After.php
+++ b/src/Event/After.php
@@ -21,7 +21,7 @@ class After extends AbstractEvent
      * @var mixed
      */
     protected $result;
-    
+
     /**
      * @var array
      */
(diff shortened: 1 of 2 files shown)


src/Event/After.php
```diff
@@ -89,4 +96,60 @@ public function setResult($result)
...
+     *
+     * @param string $key     argument key
+     * @param mixed  $default default return value
+     *
+     * @return mixed
+     */
+    public function getArgument($key, $default = null)
+    {
+        if (! array_key_exists($key, $this->arguments)) {
+            return $default;
```
frankdejonge: @GrahamCampbell ping!

GrahamCampbell: Yeh, probably. Ping @Gounlaf.

Gounlaf: Instead of a default value, you think it would be better to throw an exception ?

frankdejonge: @Gounlaf correct. The event should not be abused to transport parameters that don't affect the call.

PR #356, opened 2015-01-01
Throw a RootViolationException if the root folder does not exist and it'...

andris-sevcenko:
...s not possible to create it in Local Adapter.

### DIFF
--- a/src/Adapter/Local.php
+++ b/src/Adapter/Local.php
@@ -53,7 +53,7 @@ class Local extends AbstractAdapter
             if (is_dir($root) === false) {
                 throw new RootViolationException("The root directory does not exist and cannot be created.");
             }
-		}
+        }
 
         return realpath($root);
     }
--- a/src/Adapter/Local.php
+++ b/src/Adapter/Local.php
@@ -7,6 +7,7 @@ use FilesystemIterator;
 use Finfo;
 use League\Flysystem\AdapterInterface;
 use League\Flysystem\Config;
+use League\Flysystem\RootViolationException;
 use League\Flysystem\Util;
 use RecursiveDirectoryIterator;
 use RecursiveIteratorIterator;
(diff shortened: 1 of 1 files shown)


GrahamCampbell:
Please indent with spaces.

src/Adapter/Local.php
```diff
@@ -40,12 +41,18 @@ public function __construct($root)
      *
      * @param string $root root directory path
      *
-     * @return string real path to root
+	 * @throws RootViolationException
```
staabm: Indentation

frankdejonge:
Does is_dir return false when a dir exists and it's writable? Because that's what this implies. If not (which I think is the case), the writable check should be moved to after the first if statement.

andris-sevcenko:
Since is_dir would return false on non-existing paths, say, "/this/path/does/not/exist", the code would then try to create the non-existing path. If the path would not be writable, it would result in a warning from mkdir. Throwing an exception there, if the folder does not exist and the path is not writable, allows for a more graceful handling of it.

frankdejonge:
The problem is that you're checking for the wrong folder. A is_writable on a non-existing folder, which is what you're checking for, always results into `false`. So mkdir can, in this setup, never be called.

andris-sevcenko:
@frankdejonge Ah I see. You're right. The cleanest way to perform this check then, in my opinion, would be to suppress the errors when creating the folder and then checking if the folder exists afterwards. If not, then throw the exception so it can be handled. Does that sound OK?

frankdejonge:
@andris-sevcenko iirc, the mkdir function doesn't emit an error. Throwing an exception what it fails seems like a good idea though! :+1:

andris-sevcenko:
Suppressing errors suppresses warnings as well :)

GrahamCampbell:
Indentation is still incorrect. Also, this should be squashed to one commit.

andris-sevcenko:
Right. Will resubmit a clean pull request.

PR #357, opened 2015-01-01
Pull request/dont cache empty objects

andris-sevcenko:
Sometimes AWS S3 would return a bogus object with path set to false and no dirname key in the array, which would make AbstractCache::has() choke, because $path was set to false and, subsequently, Util::pathinfo would choke, because the array returned by pathinfo() would not have the "dirname" key set in the response array.

### DIFF
--- a/src/Cache/AbstractCache.php
+++ b/src/Cache/AbstractCache.php
@@ -70,11 +70,11 @@ abstract class AbstractCache implements CacheInterface
         $directories = [$directory];
 
         foreach ($contents as $index => $object) {
-			if (empty($object['path'])) {
-				continue;
-			}
+            if (empty($object['path'])) {
+               continue;
+            }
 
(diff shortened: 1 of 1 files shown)


frankdejonge:
@andris-sevcenko if it's AWS that does this, it should be fixed in the S3 Adapter. Also, is there more information on this? Something for me to verify? I've never had this personally.

andris-sevcenko:
I think I found the underlying issue - if you have a folder, say, "subfolder" on your S3 bucket and you use the "subfolder/" prefix, S3 returns all the files, including the containing "subfolder/". After AbstractAdapter::removePathPrefix() strips the path prefix, the object for the containing folder now contains false for "path" value which causes all the issues later on.

I agree that handing it in the Cache layer is far from optimal. I'll submit a new pull request.

PR #358, opened 2015-01-01
Check if it was possible to create the root folder (in case it was missi...

andris-sevcenko:
...ng) in Local Adapter. If not, throw an Exception.

### DIFF
--- a/src/Adapter/Local.php
+++ b/src/Adapter/Local.php
@@ -7,6 +7,7 @@ use FilesystemIterator;
 use Finfo;
 use League\Flysystem\AdapterInterface;
 use League\Flysystem\Config;
+use League\Flysystem\RootViolationException;
 use League\Flysystem\Util;
 use RecursiveDirectoryIterator;
 use RecursiveIteratorIterator;
@@ -40,12 +41,18 @@ class Local extends AbstractAdapter
      *
      * @param string $root root directory path
      *
+     * @throws RootViolationException
+     *
      * @return string real path to root
      */
     protected function ensureDirectory($root)
     {
         if (is_dir($root) === false) {
-            mkdir($root, 0755, true);
+            @mkdir($root, 0755, true);
+
+            if (is_dir($root) === false) {
+                throw new RootViolationException("The root directory does not exist and cannot be created.");
+            }
         }
 
         return realpath($root);


src/Adapter/Local.php
```diff
@@ -40,12 +41,18 @@ public function __construct($root)
...
      * @return string real path to root
      */
     protected function ensureDirectory($root)
     {
         if (is_dir($root) === false) {
-            mkdir($root, 0755, true);
+            @mkdir($root, 0755, true);
+
+            if (is_dir($root) === false) {
+                throw new RootViolationException("The root directory does not exist and cannot be created.");
```
staabm: $root should be part of the message

andris-sevcenko: I think this is a matter of opinion, as it could potentially expose information about the server.

frankdejonge:
@andris-sevcenko I've fixed it a little differently. Also, the exception class/type was wrong. The RootViolationException is only used for invalid access ('root/../../something' stuff).

andris-sevcenko:
As per http://php.net/manual/en/function.mkdir.php, this will still emit a warning if it's not possible to create a folder at the $root path. 

For example, on servers with display_errors set to on, this could botch up a JSON response, hence my suppressing errors/warnings of mkdir with @mkdir.

andris-sevcenko:
@frankdejonge any thoughts on supressing the warning generated my mkdir?

diimpp:
What if it's readonly fs? Or I just don't want it to be writable?

paq85:
Hi.
I would also like to use that adapter with a read only filesystem and this change made it impossible (?).

frankdejonge:
@andris-sevcenko sorry for the late response, been quite hectic here. Errors should be handled by the consumer of the package. In production they shouldn't be displayed, in dev they should be handled. If there's a case when a directory shouldn't be created, this should either result in a check before the adapter is created, or it should fail like it does now.

@paq85 read-only adapters are already possible. Returning false from any adapter function is valid and doesn't break the interface.

paq85:
@frankdejonge but I would have to create my own read-only Local Adapter, right?
It's not possible to use the existing one?

frankdejonge:
@paq85 I would opt not doing that but just wrapping the filesystem class and only exposing what you want, so only the "getters" and read/readStream? Should be a 10 min job including tests ;)

PR #360, opened 2015-01-02
Remove S3 objects with no path at all, that appear when using path prefi...

andris-sevcenko:
...x and listing objects.

When getting the list of objects for an S3 in a scenario where you want to list the contents of a folder/, S3 returns an object for the containing folder/ as well, resulting in an object with path set to false after AbstractAdapter::removePathPrefix() is called, which trips up the caching logic, not to mention that the containing object should not be there at all in the first place.


frankdejonge:
@andris-sevcenko which version of the AWS SDK are you using?

andris-sevcenko:
@frankdejonge 2.7.12

frankdejonge:
@andris-sevcenko reproduces and merged, thanks! :+1:

PR #361, opened 2015-01-02
Prevent duplicate listing of directories.

andris-sevcenko:
If folders are created using AWS console in an S3 bucket, they will have their own listing there and will appear in the original $listing array. This patch ensures that the folders are not listed twice, if that is the case.

### DIFF
--- a/src/Util.php
+++ b/src/Util.php
@@ -195,7 +195,11 @@ class Util
                 $parent = static::dirname($parent);
             }
 
+<<<<<<< HEAD
             if ($object['type'] == 'dir') {
+=======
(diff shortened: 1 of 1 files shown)


frankdejonge:
@andris-sevcenko sounds useful. The tests are failing, can you look into that? Since it's failing on all version, I take it you might have not run the tests locally before submitting?

andris-sevcenko:
I was in a rush and didn't run tests on this one, yeah - I'll look into that once I get home early next week.

andris-sevcenko:
Clearly my head is still jetlagged. I'll just submit a new pull request with a clean push.

PR #363, opened 2015-01-07
Http guzzle adapter

duncan3dc:
I'm creating this PR to hopefully get the great work done by @hgraca into a release of flysystem.

I've taken the feature branch from the main repo, rebased it from master, and updated a couple of broken things.

What blockers are there to getting this merged?

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -26,6 +26,7 @@
         "dropbox/dropbox-sdk": "~1.1.1",
         "rackspace/php-opencloud": "~1.11",
         "barracuda/copy": "~1.1.4",
+        "guzzlehttp/guzzle": "~5.0",
         "phpseclib/phpseclib": "~0.3.5",
         "tedivm/stash": "~0.12.0",
         "league/phpunit-coverage-listener" : "~1.1",
--- a/readme.md
+++ b/readme.md
@@ -67,6 +67,7 @@ Want to get started quickly? Check out some of these integrations:
 * Dropbox
 * Copy
 * Ftp
+* Http (through guzzle 5)
 * Sftp (through phpseclib)
 * Zip (through ZipArchive)
 * WebDAV (through SabreDAV)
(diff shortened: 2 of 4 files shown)


GrahamCampbell:
This should be closed. If this were to be implemented, then this must go into another repo, not into this one.

hgraca:
Why not, if one may ask?



Herberto Graça
_________________________________
<http://pt.linkedin.com/pub/herberto-gra%C3%A7a/8/10/b32>

2015-01-12 20:10 GMT+01:00 Graham Campbell <notifications@github.com>:

GrahamCampbell:
> Why not, if one may ask?

I already answered that.

> If this were to be implemented, then this must go into another repo, not into this one.

All the adapters are getting moved out in this way.

frankdejonge:
@duncan3dc Hi, sorry for the late reply. What @GrahamCampbell is correct. When this is released it should be done in a separate repository. The reason for this that the package will have it's own dependencies which are not directly related to the core of Flysystem. In order for this PR to be complete a couple of things should be done, namely: Test coverage (there are untested parts), complete API coverage (also writing using POST with the file as the body) and exception handling.

duncan3dc:
Thanks @frankdejonge

barryvdh:
Writing a file with POST isn't really the use-case of this adapter, right? Wasn't the purpose of this to provide a read-only HTTP adapter (for example with Glide).
But @duncan3dc you don't really need it to be accepted by Flysystem, you could just release your own package that provides this adapter, so people can still use it :)

Also, you could think about using listContents to find all links in a html document and convert them to relative paths, for example when you have a directory listing.

And you could probably simplify a bit by using the Guzzle base_url options: http://guzzle.readthedocs.org/en/latest/clients.html#creating-a-client

frankdejonge:
@barryvdh there are Flysystem implementations with StackPHP, these can support writes too. Listing contents is indeed a tricky one. Open to suggestions/advice about this.

barryvdh:
But how many websites have something like that implemented? If you want to use Flysystem for a random website, 99% of the time there will be no rest handler to upload stuff, so it's a bit pointless for a 'general' Http adapter. Probably a specific REST handler (or even specific handler for https://github.com/h4cc/stack-flysystem) would be better.
(And you did suggest just returning false for this specific case in https://github.com/thephpleague/glide/issues/16)

But I also agree that a partial filesystem doesn't really make sense to be included in the core.

frankdejonge:
@barryvdh that's true. Just saying that there _is_ a use-case ;) I'll happily accept it as read-only too. The other points still stand though.

hgraca:
Exactly, there are use cases.
I created this adapter exactly because the REST application I was working
on, needed to get some files that a client would simply dump in an http
location.

;)



Herberto Graça
_________________________________
<http://pt.linkedin.com/pub/herberto-gra%C3%A7a/8/10/b32>

2015-01-13 12:24 GMT+01:00 Frank de Jonge <notifications@github.com>:

PR #365, opened 2015-01-08
Suppress warnings generated by mkdir.

andris-sevcenko:
When having an illegal path, mkdir will generate a warning which, on some servers, can botch up a JSON response. Suppressing it allows these scenarios to be handled using Exceptions.


frankdejonge:
@andris-sevcenko these errors are meaningful though and should lead to a developer action. The error can also be suppressed globally, which is what you want to do in production anyway. If the root dir can't be created, it's a real issue, but not something Flysystem is reponsible for. The encapsulating code should take care of this and prevent the error from happening in the first place.

PR #372, opened 2015-01-17
Memcached constructor signature optimization, removed excess use

creocoder:


frankdejonge:
@creocoder you say `optimisation`, but I only see a preference in coding style here. Is there any benefit of using this way?

GrahamCampbell:
I don't see the point in this.

creocoder:
> Is there any benefit of using this way?

> I don't see the point in this.

I can ask same question :) Import class to namespace just to not use FQN one time. Benefits? **Less code, same result**.

GrahamCampbell:
It's a preferred code style.

GrahamCampbell:
:-1: on merging this.

frankdejonge:
I'd like to keep the same style across the project, which is basically: import all the things.

creocoder:
I agree with code stype in all other cases. But this one looks like exception (collision between names, need aliasing) because its really make code hard to read. Another possible solutions is to rename main class and use Memcached without aliasing it.

frankdejonge:
@creocoder in the community at large it's very common to use aliases like this to prevent naming collisions. For instance, in the symfony community when creating a console application, the Application class is often aliased to BaseApplication for the same reasons.

creocoder:
@frankdejonge Ok, thanks for explaination. I think Memcached class name is good from practical point of view and seems it was the only way to support codestyle. I mean original code.

frankdejonge:
@creocoder you're welcome. Thank you for showing an interest and for wanting to contribute!

PR #402, opened 2015-02-02
[WIP] Test suite for adapters

### LINKED ISSUE #401: Base Test for adapters
As all adapters implement the same interface and are supposed to behave the same way, I think it would make a lot of sense to have one or more base tests that ensure the correct behavior. The adapter tests could extend those base tests and do the bootstrapping.

If we can agree on that, i could try to do a PR for that and update the included tests to use the test suite.

I would propose to create a Test namespace inside src - it is code that is useful to 3rd party adapters too. I would split into several tests: ReadTestCase, ReadMetadataTestCase, WriteTestCase, WriteDirTestCase, DeleteTestCase, VisibilityTestCase to make the size of each test more manageable. Adapter specific tests can then extend the base cases that make sense for that adapter, but still overwrite test methods or skip them if they do not apply.

dbu:
fix #401

i started to work on this. i would love to discuss before i invest more time, so that we can agree if this is the right way to go.

TODO
* [ ] Add some documentation
* [ ] Test case for write (split that test in some way or all in one?)


src/Test/Adapter/ReadTestCase.php
```diff
@@ -0,0 +1,197 @@
...
+    }
+
+    public function testGetMetadata()
+    {
+        $result = $this->adapter->getMetadata('file.txt');
+        $this->assertType('file', $result);
+        $this->assertPath('file.txt', $result);
+        $this->assertSize(strlen(static::FILE_CONTENT), $result);
+        $this->assertHasTimestamp($result);
+        // no mimetype in default metadata set
```
dbu: is this on purpose?

src/Test/Adapter/ReadTestCase.php
```diff
@@ -0,0 +1,197 @@
...
+        $result = $this->adapter->getTimestamp('file.txt');
+        $this->assertHasTimestamp($result);
+    }
+
+    public function testGetMimetype()
+    {
+        $result = $this->adapter->getMimetype('file.txt');
+        $this->assertMimeType('text/plain', $result);
+    }
+}
```
dbu: we miss a bunch of edge cases: trying to read a directory, get metadata of a directory.

also, i miss the test for getVisibility - what should happen if this is not supported by the underlying filesystem? always return true? or throw an exception?

dbu:
ping @frankdejonge

src/Test/Adapter/ReadTestCase.php
frankdejonge: The getMetadata call is basically for basic metadata (type, path[info], size). Mimetype is a more specialised call in the majority of FS implementations (like local for instance) which makes it more expensive to have in the set of common metadata. In theory only path and type are expected to be returned in this call taking into account the various adapter types.

src/Test/Adapter/ReadTestCase.php
frankdejonge: When not supported it should throw an exception, same for setVisibility.

src/Test/Adapter/ReadTestCase.php
dbu: okay, makes sense to me. should the return of this method be limited to only those two, to guarantee portability between adapters? i guess its a bit late now already, would be a BC break to drop the timestamp, or wdyt?

in general, is this method allowed to return any other keys?

src/Test/Adapter/ReadTestCase.php
dbu: right. i will check the adapter for the NotSupportingVisibilityTrait and expect the exception or otherwise expect that visibility works as expected.

dbu: fixed

src/Test/Adapter/ReadTestCase.php
dbu: i would love to make the adapter interface and the Filesystem phpdoc more specific in this regard.

frankdejonge:
@dbu Hey, I've had some time to think about it. Thanks for the twitter reminder though :+1:. The one thing that seems off for me in this approach is the fixtures stuff. It might be that I've not used this kind of fs stubbing before, but it makes me wonder how this would help, for example, the S3 adapter. Would we just say that the state of the FS should reflect X during testing? In an ideal situation the same approach/solution would be to the adapters running tests with phpspec. Which makes this a tad bit harder. For example, call assertions can then not be nicely made. Perhaps instead of these fixtures we can introduce some helpers (like I've done in some cases with some adapters). Which would look something like this:

```php
protected function ensureDirectoryContainsListing($dirname, $listing);
protected function ensureFileExistsAtLocation($location, $contents);
protected function ensureDirectoryExistsAtLocation($location);
protected function ensurePrivateFileExistsAtLocation($location, $contents);
```

Something along these lines? This would be easily translatable into phpspec test helpers and removes some of the hidden context. 

What do you think, @dbu?

dbu:
this sounds like a good idea, yes. should we then still implement those methods using the adapter, so that users only need to write their own thing if mocking is needed?

we could even use that to test error conditions then. adding new ensure methods will be tricky for adapters, as they might need to overwrite them.

for ensureDirectoryContainsListing - is ensureFileExists('folder/file', $content) not enough to assume that listing folder/ returns "file"?

frankdejonge:
@dbu the ensureDirectoryContainsListing would create different call expectations. For instance, in S3, the test should set expectations to receive a listObjects call instead of a getObject call. I'd say having those methods declared as abstract or with a body which throws an exception would be the way to go. This way the test suite for a given adapter can make their own call what's best for it. For local it might be good enough to just write the actual file, while one that uses an SDK will not. So the only case that's actually helped by already implementing this in an abstract adapter test is Local. Since it's the only one, I'd say just have it there instead of having inheritance pollution?

dbu:
well, i am sure there is other adapters that can write to their storage in a test. at least the phpcr adapter can. maybe offer a trait for that then?

frankdejonge:
@dbu a trait for that would be ideal indeed :+1:

dbu:
there you go, now with trait. if this looks good, i will continue with
* [] the write part 
* [] try to port all adapter tests in the core repo to the new format
* [] write some documentation (that should go into the gh-pages, right?)

dbu:
i just pushed the write test case and adapted the local test to use it.

for now, i keep all the edge case testing specific to the adapter. not sure if we should do something more here to share those tests among all adapter tests as well. wdyt?

can you give some feedback on this, before i change all adapter tests to the new test suite?

tests/Adapter/LocalWriteTest.php
```diff
@@ -0,0 +1,149 @@
...
+            }
+        }
+    }
+
+    /**** Additional edge cases tests ****/
+
+    public function testFailingStreamCalls()
+    {
+        $this->assertFalse($this->adapter->writeStream('false', tmpfile(), new Config()));
+        $this->assertFalse($this->adapter->writeStream('dummy', tmpfile(), new Config()));
```
frankdejonge: These tmpfile's need to be closed, windows users have problems when they aren't.

dbu:
sure, will see to close files then.

what do you think about tests for edge cases or not? should we just recommend driver implementors to do them or should we enforce consistent behaviour? see for example #440 - i guess we should not do this in this PR, or it will never finish...

do you agree that whats missing to merge this is to convert the other adapter tests to the new thing and then have some documentation?

frankdejonge:
@dbu I think we can role this out for the Local adapter first, merge and migrate the other adapters later, since most of them would require a main package update for this. I'd leave the #440 for later for the same reason you mentioned. Also one other remark, it would probably be good to reparate those test helpers from the main src directory, create an `autoload-dev` clause for them, add them to the .gitattributes for installation ignore. Argee?

tests/Adapter/LocalWriteTest.php
dbu: i only moved this around. but closing now

Potherca:
I see this issue is closed, but what is the status? Have the changes been merged through other commits or have the changes been rejected?

dbu:
i failed to ever wrap it up. if somebody wants to pick this up, please do.

frankdejonge:
@Potherca @dbu I'm going to be merging this in in parts, may not be the original commits but the work will be included.

PR #407, opened 2015-02-06
add information for phpcr to github pages

dbu:


frankdejonge:
@dbu can you add it to the _meny.yml too?

dbu:
added to _data/menu.yml - correct like this?

PR #408, opened 2015-02-06
Link to cache repo

barryvdh:


barryvdh:
Ah wait it is in the docs, but only in the upgrade part. Perhaps better fix it there?

PR #410, opened 2015-02-06
Fix timestamp from FTP adapter

cramiro:
What does this Pull Request Do?
-------------------------------
The way the timestamp is calculated
```php
$timestamp = strtotime($month.' '.$day.' '.$time);
```

forgets that the FTP server don't show the year if the date is +-6 months. So, was calculating a wrong timestamp somethimes.

What I did was ask for the timestamp to the FTP server using
```php
$timestamp = ftp_mdtm($this->getConnection(), $name);
```
This solves the problem. Now you can do things like get the last file in a `listContents()` safely.

**Changelog**
`listContents()` no longer returns the `timestamp` because was inaccurate and using `ftp_mdtm()` could led to a huge performance penalty.  To get the timestamp you have to explicitly ask for it using `getTimestamp()`.

How should this be manually tested?
-----------------------------------
I added test to this.

Any background context?
-----------------------
I need for a project to search for the last file in many FTP accounts. This week I started using the library(which is awesome, thanks!!) and started to have issues about the files I was downloading and processing. Finally saw the issue, the `timestamp`, and here one way I propose to fix it.

Any extra info?
---------------
This fix has a performance issue if the directory have lots of files, becouse, instead of using only one LIST, we'll be adding one MODTIME call to the server for each file. So, we'll end with `N` roundtrips more (`N` number of files in the directory).

Options:
Another option would be to check the date and see if it's +-6 month, and fix it if necessary. It's easy if we get the current date of the server we are running, but to do it properly we should use the date of the FTP server we are in. We can do this by writing a file to check the FTP server's date, but (1) we could be breaking read-only implementa and (2) if we do it do it in the `normalizeObject()` method is as bad for performance (3) if we do it before normalizing, we have to carry the date with us, changing the parameters we actually have for all the other methods until we get to `normalizeObject()`.

Hope you find this useful! ;)

### DIFF
--- a/src/Adapter/AbstractFtpAdapter.php
+++ b/src/Adapter/AbstractFtpAdapter.php
@@ -288,18 +288,17 @@ abstract class AbstractFtpAdapter extends AbstractAdapter
         $item = preg_replace('#\s+#', ' ', trim($item));
         list($permissions, /* $number */, /* $owner */, /* $group */, $size, $month, $day, $time, $name) = explode(' ', $item, 9);
         $type = $this->detectType($permissions);
-        $timestamp = ftp_mdtm($this->getConnection(), $name);
         $path = empty($base) ? $name : $base.$this->separator.$name;
 
         if ($type === 'dir') {
-            return compact('type', 'path', 'timestamp');
+            return compact('type', 'path');
         }
 
         $permissions = $this->normalizePermissions($permissions);
         $visibility = $permissions & 0044 ? AdapterInterface::VISIBILITY_PUBLIC : AdapterInterface::VISIBILITY_PRIVATE;
         $size = (int) $size;
 
-        return compact('type', 'path', 'visibility', 'size', 'timestamp');
+        return compact('type', 'path', 'visibility', 'size');
     }
 
     /**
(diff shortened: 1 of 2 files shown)


frankdejonge:
@cramiro good call, although the placing of this solution doesn't feel right to me. Like you've already stated, this will result in quite the performance penalty. I'd suggest replacing the `getTimestamp` method with the `ftp_mdtm` call and removing the timestamp from the listing / getMimetype call. The value given back from the current implementation was faulty, so unreliable anyway.

cramiro:
@frankdejonge I agree with you, will do that today and send you the PR. My concern is what about those that are currently using the timestamp from the listing?

cramiro:
@frankdejonge what do think about this approach? Please let me know...thanks in advance!

cramiro:
@frankdejonge please let me know if you have any suggestions here...

frankdejonge:
@cramiro there seem to be some unused variables now, could you remove those?

cramiro:
@frankdejonge you can give it a look now...

cramiro:
@frankdejonge ping

frankdejonge:
Merged manually.

PR #412, opened 2015-02-08
`getAdapter()` support CachedAdapter

nao-pon:
ref. https://github.com/thephpleague/flysystem-cached-adapter/pull/8


frankdejonge:
The other PR is fine, this however won't get accepted because it's magical. Since there could be another decorator on top or below this one, it makes no sense to make the assumption that this is what the user wants. It only fits one use-case.

PR #413, opened 2015-02-08
`getAdapter()` support stacked type adapter

nao-pon:
Like a CachedAdapter.

ref. #412

### DIFF
--- a/src/Filesystem.php
+++ b/src/Filesystem.php
@@ -46,7 +46,11 @@ class Filesystem implements FilesystemInterface
      */
     public function getAdapter()
     {
-        return $this->adapter;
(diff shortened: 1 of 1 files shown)


frankdejonge:
Same reason as before. The Filesystem will only get the underlying adapter, it's the user's own responsibility to fetch an underlying resource if needed. This doesn't fix the issue, nor is it the responsibility of the Filesystem class to do so.

nao-pon:
Thank you, I understand.

frankdejonge:
I'll look into writing tests for the other PR soon.

nao-pon:
I really appreciate it. :-)

PR #424, opened 2015-02-13
Make the suggest messages consistent

dwightwatson:
I know this is small, but it was annoying me a little.


frankdejonge:
@dwightwatson all contributions of this sort are very welcome.

PR #427, opened 2015-02-18
Use dataprovider in testOutsideRootPath

mapyo:
@frankdejonge

Thank you for this great library!

I found  a fix leakage of the following commit.
And, I have fixed it.
https://github.com/thephpleague/flysystem/commit/3108d21339406e71126f209dccf695de01bf5130#diff-0a84506104c31f88ac00dbab7372dd6aR86

In addition, because the test of the following locations fail, I have fixed it.
https://github.com/thephpleague/flysystem/blob/master/tests/UtilTests.php#L79

Thank you.

### DIFF
--- a/src/Util.php
+++ b/src/Util.php
@@ -89,7 +89,7 @@ class Util
 
         $normalized = static::normalizeRelativePath($normalized);
 
-        if (preg_match('#/\.{2}|^\.{2}/#', $normalized)) {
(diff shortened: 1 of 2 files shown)


frankdejonge:
@mapyo thanks for bringing this to the attention. I've replaced the PR, it missed one edge case. Thanks for the initial implementation.

mapyo:
:octocat:

PR #428, opened 2015-02-19
Added details for enabling S3 reduced redundancy storage

alexbilbie:


atrauzzi:
What if you only want reduced redundancy for certain files, not all?

frankdejonge:
@atrauzzi then you'd pass this in the options array when you write the file.

atrauzzi:
Awesome!  Worth mentioning in the docs as well, or is it already there?

alexbilbie:
Set up multiple adapters.




I have an adapter for my thumbnails S3 bucket which has reduced redundancy, an another for my original images bucket which has standard redundancy.



—

atrauzzi:
Oh, so you wouldn't do it when you write the file?  Multiple adapters seems tricky in a dependency injection scenario.

alexbilbie:
Not at all, I have a thumbnail filesystem and an image filesystem in my container. And I just use the appropriate one. Easy.



—

atrauzzi:
Are you accessing the root of those graphs in your container using `'string.names'`?

alexbilbie:
Yes



—

atrauzzi:
That makes things somewhat less ideal then as the "entrypoint" in your graph then isn't dependency injection, but instead service location.

I don't use string names in my dependency injection setup because I don't use facades and because they can't be type-inferred.

Based on this, I would suggest the functionality be made available (and documented) when storing individual files.

alexbilbie:
Uh what?

You're conflating two completely different things here.

Service location comes from injecting the container itself into your classes. I'm not doing that.

I'm using `league/container` and my code looks like this:

```php
$container->add('ImageFS', function () use ($container) {
    $adapter = new AwsS3Adapter(
        $container->get('S3Client'),
        $container->get('config')->get('aws.s3.imagebucket')
    );
    return new Filesystem($adapter);
});

$container->add('ThumbnailFS', function () use ($container) {
    $adapter = new AwsS3Adapter(
        $container->get('S3Client'),
        $container->get('config')->get('aws.s3.thumbnailbucket'),
        null,
        [
            'StorageClass'  =>  'REDUCED_REDUNDANCY'
        ]
    );
    return new Filesystem($adapter);
});
```

and then injecting it into my classes like so:

```php
$container->add('ImageService', function () use ($container) {
    return new ImageService(
        $container->get('CommandBus'),
        $container->get('EventEmitter'),
        $container->get('ImageModel'),
        $container->get('ImageFS'),
        $container->get('ThumbnailFS')
    );
});
```

If you're not using named (aliased) definitions in your container then just subclass the Adapter and use that for your reduced redundancy bucket, and the normal `AwsS3Adapter` for your standard redundancy bucket.

atrauzzi:
Configuring two buckets for the same service when the only thing varying is a header when writing to the source seems like a workaround more than a solution.

The implication is that until I know whether I'm storing something in reduced redundancy, I'm going to have to make sure I'm getting both injected everywhere.  There's definite code smell going on here.

atrauzzi:
Also, FWIW - I haven't confused anything.

Service location makes no assumptions about whether the locator itself is being obtained via DI or a global mechanism.

alexbilbie:
So write an AwsS3ReducedRedundancyAdapter, send a PR and I'm sure Frank will accept it



—

atrauzzi:
There should only be one S3 adapter, and then if what @frankdejonge said is possible, those options could be passed through to the storage layer to conditionally include in the requests it sends out.

PR #429, opened 2015-02-19
Handle leading and trailing spaces in ftp file names

icewind1991:
Currently the FTP adapter truncates any extra whitespace from the result of `ftp_rawlist` which breaks `listContents` when a file name contains a leading or trailing slash


frankdejonge:
@icewind1991 I fixed this in another way. Your fix wasn't bad, but that regex is very easy to misinterpret later on. I think I've fixed it in a slightly nicer way by only replacing all the whitespace before the last (which is always one space). Thanks for bringing this to the attention.

icewind1991:
Forgot that you could limit preg_replace like that, thanks for the fix

PR #436, opened 2015-02-27
the root directory can now be a link to a directory

damko:


frankdejonge:
@damko could you add a test for this as well?

frankdejonge:
@damko it's already possible to use a link as the root. When the root is resolved, the realpath is used as the actual root. I've added the test you supplied to the test suite (slightly modified) which shows you this works. Therefor I'm closing this PR.

damko:
You are right. 

There is something unclear though: I wrote the patch *because* I had troubles with links but I tried now and I clearly see that it works even without readlink() as you say.

I don't understand if I can't reproduce my previous situation or if I was wrong somewhere at the time of writing the patch ... I've no idea.

Sorry anyway

frankdejonge:
@damko no worries :) just glad it's sorted, and we have another good test case for the future.

PR #450, opened 2015-03-20
Use getter in removePathPrefix

Anahkiasen:
Currently if you override `getPathPrefix` it won't be used in `removePathPrefix`


frankdejonge:
Why are you overwriting getPathPrefix?

frankdejonge:
I saw why you needed it in the flysystem-vfs thing. So I've added this, I've replaced the PR due to a double function call. Thanks for the PR!

PR #453, opened 2015-03-26
Small improvements to Local object constructor

damko:


levacic:
I just spent far more time than necessary debugging an issue where my symlinked root path was throwing errors with `mkdir()`, because the directory that the symlink points to was moved (so `$root` actually existed, but `is_dir()` was returning `false`).

It would be a really good idea to rebase and merge this PR, since it offers some very useful checks.

frankdejonge:
The issues addressed here have been addressed a different way which also handles race conditions.

PR #457, opened 2015-04-03
Check if method exists on setting configuration values

danut007ro:
Check if method exists on setting configuration values. Related to https://github.com/helios-ag/FMElfinderBundle/pull/133 and https://github.com/thephpleague/flysystem-sftp/pull/6

### DIFF
--- a/src/Adapter/AbstractFtpAdapter.php
+++ b/src/Adapter/AbstractFtpAdapter.php
@@ -47,6 +47,7 @@ abstract class AbstractFtpAdapter extends AbstractAdapter
             }
 
             $method = 'set'.ucfirst($setting);
+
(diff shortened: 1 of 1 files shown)


src/Adapter/AbstractFtpAdapter.php
```diff
@@ -45,7 +45,11 @@ public function setConfig(array $config)
             if (! isset($config[$setting])) {
                 continue;
             }
-            $this->{'set'.ucfirst($setting)}($config[$setting]);
+
+            $method = 'set'.ucfirst($setting);
+            if (method_exists($this, $method)) {
```
frankdejonge: There needs to be a blank line before the `if`.

PR #458, opened 2015-04-04
Changes in the way some paths are normalized

rado-h:
The current regex that handles the removal of self referring parts in the path doesn't seem to work properly in some cases:

Examples:
```php
$path = Util::normalizePath("dirname./");
// Expected:    
$path == "dirname.";
// Actual:           
$path == "dirname";

$path = Util::normalizePath("dirname/.");
// Expected:    
$path == "dirname";
// Actual:          
$path == "dirname/.";

$path = Util::normalizePath("dirname/../");
// Expected:    
$path == "";
// Actual:          
$path == "dirname/.";
```

With this change the regex will replace "." at the end of the path only if it's after a "/ " and it's followed by either a "/" or the end of the string

### DIFF
--- a/src/Util.php
+++ b/src/Util.php
@@ -108,7 +108,7 @@ class Util
     public static function normalizeRelativePath($path)
     {
         // Path remove self referring paths ("/./").
-        $path = preg_replace('#/\.(?=/)|^\./|\./$#', '', $path);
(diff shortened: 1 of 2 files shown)


frankdejonge:
@rado-h thank you very much :+1:

PR #460, opened 2015-04-07
Require ext-mbstring in composer

meadsteve:
```Util::contentSize``` makes use of an mb_* function so this library should depend on it.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -14,7 +14,8 @@
         }
     ],
     "require": {
-        "php": ">=5.4.0"
+        "php": ">=5.4.0",
+        "ext-mbstring": "*"
     },
     "require-dev": {
         "ext-fileinfo": "*",

### CODE REVIEW

**Comment by barryvdh** [comment]
There are userland implementations of the mb_* functions that people might be using instead of ext-mbstring, but they would be blocked by Composer with this requirement right?

**Comment by meadsteve** [comment]
@barryvdh they would yes. Which is definitely a downside. The question is how common are people using userland implementations?

I just spent a bit of time scratching my head over the s3 adapter not working but it turned out to be that mbstring wasn't enabled in my environment.  It would have been nice to be able to detect this earlier.

**Comment by frankdejonge** [comment]
@meadsteve didn't you get an error saying the mb_* function didn't exist?

**Comment by barryvdh** [comment]
I was thinking about Laravel, but I see since 5.0 they require ext-mbstring. Laravel 4 required https://github.com/tchwork/utf8 as alternative. But flysystem is only installed by default with Laravel since 5.0 so not sure how many people actually don't have it installed.

**Comment by meadsteve** [comment]
@frankdejonge I did when I eventually found the error but unfortunately I'm using this as part of tidying up a legacy code base and the error log is far from empty.

**Comment by meadsteve** [comment]
Basically what I think we want to avoid is a situation where there's an implicit/hidden dependency.

**Comment by barryvdh** [comment]
If it really is needed, perhaps best to require it indeed.

But how many mb_* functions are used? Just the mb_strlen() function for the content size? And how critical is it, are there alternatives for when the method doesn't exist?

**Comment by frankdejonge** [comment]
@meadsteve could you move the requirement to the normal "require" section? I'll merge it then. I've just researched how to effectively do this. Since the package relied on ext-mbstring it should be defined as a dependency. If a user wants to use a polyfill, that's still possible. You're able to work around this restriction if you include the polyfill and add a "replace" section to the composer.json file.

```
{
  "replace": {
    "ext-mbstring": "*"
  }
}
```

I'm guessing this will be resolved correctly too if the polyfill specifies this in it's composer.json. In other words, there's no reason for us _not_ to require this.

**Comment by meadsteve** [comment]
@frankdejonge good idea!

I've updated the commit so this is correctly in the require section.

**Comment by rado-h** [comment]
Doesn't mb_strlen($contents, '8bit') behave like strlen($contents) if mbstring.func_overload is off? If that's the case, this dependency might be easily removed with a check if mb_strlen exists (if it doesn't, then mbstring is not installed so we can be sure mbstring.func_overload is not turned on and strlen returns bytes):
```php
if (function_exists("mb_strlen")) {
    return mb_strlen($contents, '8bit');
 }
        
 return strlen($contents);
```

twistor:
I think this was a bad idea also, #491

Issue #464, opened 2015-04-20
Memory Limit Exceeded or Request Timeout when handle large file.


khanhicetea:
When handle a large file (> 7MB - in Heroku), it seem terrible when trying detect MIME Type by content

``` php
public static function guessMimeType($path, $content)
    {
        $mimeType = MimeType::detectByContent($content); // <=== Problem here

        if (empty($mimeType) || $mimeType === 'text/plain') {
            $extension = pathinfo($path, PATHINFO_EXTENSION);

            if ($extension) {
                $mimeType = MimeType::detectByFileExtension($extension) ?: 'text/plain';
            }
        }

        return $mimeType;
    }
```

I think we should detect its size before call that function like

``` php
if (filesize($path) < 1024 * 1024) { // Smaller than 1 MB
    $mimeType = MimeType::detectByContent($content);
}
```

frankdejonge:
@khanhicetea Could you supply some more info? Like which adapter are you using? Which Flysystem version and are you using streams? If not, why not?

khanhicetea:
@frankdejonge , these is more info about my issue 

- Version : 1.0.3
- Adapter : Local
- Framework : [Lumen](http://lumen.laravel.com)

Error when use : 
``` php
Storage::disk('local')->append($path, $content); // File is larger than 7MB
```
Error log from Heroku
```
[20-Apr-2015 08:31:02 UTC] PHP Fatal error:  Maximum execution time of 30 seconds exceeded in 
/app/vendor/league/flysystem/src/Util/MimeType.php on line 23
```

frankdejonge:
@khanhicetea appending is not supported by Flysystem in this version. It's on the roadmap for 2.0 which is very far from seeing the light of day. Laravel simulates this behaviour by reading the file into a string, appending the contents and storing that. This causes the memory consumption.

PR #465, opened 2015-04-20
Fixed guessMimeType exceeded memory or request timeout when handle large...

khanhicetea:
Try to fix issue #464 .


frankdejonge:
This PR doesn't fix the issue. The issue relies in how append is implemented in the illuminate package.

PR #468, opened 2015-04-21
Allow prefixes like URI schemes in MountManager

graste:
Allow prefixes that are valid schemes according to RFC3986. That is, a prefix must start with a letter followed by any combination of letters, digits, plus (`+`), period (`.`), or hyphen (`-`). This change allows
case insensitive usage of prefixes as the internal key for the filesystems array is lowercased.

See http://tools.ietf.org/html/rfc3986#section-3.1

This change is a breaking change as prior to this the mounting of filesystems never checked for valid prefixes and prior to this change all number prefixes like `1` or `1337` were valid.

The main reason for me to create this pull-request is to get support for prefixes like `user-assets` or `article-images` in my project's code.

PS: I tried to adhere to the given coding style. May I suggest adding a contributing document to the project to tell people what to test and check when creating changes, what coding style to adhere to and most importantly against what branches you want to have pull requests created?


shrink0r:
Hi, 
this would be great to have.
I have a project where I would like to address mount managers by convention using compound prefixes, e.g. category.files://foobar.jpg, article.texts://foobar.md etc.
Are there plans to merge this pull request?

frankdejonge:
@graste since this doesn't add any new behaviour, I'm pretty sure this can be released as a patch update. Since the prefix is "wrongly" handled now and this change keeps BC. Could you change that, I'll merge soon.

graste:
Well it is up to you and what you deem a problem or not. From my perspective this has break potential for users with prefixes that are now invalid. Could you elaborate on ```Could you change that, I'll merge soon.```?

frankdejonge:
@graste I've loosened the restrictions for the mount name. Basically allowing anything. There's no actual reason to restrict it in the first place. This will also allow your type of prefixes and many others while keeping BC.

shrink0r:
Awesome. Thank you.

PR #480, opened 2015-05-10
Change adapter variable lookups to property calls, allow better extending of filesystem class

NinoFloris:
I am working on a saas platform in Symfony that has adapters (credentials, buckets etc) per client. They do share the filesystems (profile pictures, page file and image uploads etc). However the OneUp FlysystemBundle only allows adapter configuration through the symfony config file.

My plan was to fork that bundle, add an adapter registry, inherit the filesystem class, override its constructor method to take the registry and the adapter name to look up instead of the concrete adapter. Finally with this pull request I only have to override the getAdapter method to add the lookup in the registry, a null check and a nice exception to it.

When all the methods use the internal property this requires me to inherit but also copy paste the whole class to make the required changes.

### DIFF
--- a/src/Filesystem.php
+++ b/src/Filesystem.php
@@ -66,7 +66,7 @@ class Filesystem implements FilesystemInterface
     {
         $path = Util::normalizePath($path);
 
-        return (bool) $this->adapter->has($path);
+        return (bool) $this->getAdapter()->has($path);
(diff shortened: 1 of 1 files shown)


NinoFloris:
Ehh your scrutinizer config seems to have some problems

PR #483, opened 2015-05-19
Minor typography fixes

driesvints:


frankdejonge:
Hehe, recipie. Bet it's tasty.

driesvints:
Recipie. Like pie but only the recipe :confused:

PR #484, opened 2015-05-19
Cleanup documentation

driesvints:
Cleaned up the docs a bit after reading through them earlier today. Just some small things.

- Added the same titles to all the different adapter docs
- Converted arrays to short syntax (since Flysystem requires PHP >=5.4)
- Fixed some titles
- Removed the `<?php` opening tag from all code blocks except for the two where you were displaying an entire file
- Added some missing code for 2 recipes
- Tweaked 2 lists with links (this was a personal choice, I can always change it back if you want)

One suggestion: I'd add links to the main docs in each adapter's readme instead of keeping install/usage notes in both main docs and each readme. Things can get outdated real fast currently.

Hope you like :)

### DIFF
--- a/adapter/aws-s3-v2.md
+++ b/adapter/aws-s3-v2.md
@@ -12,16 +12,18 @@ title: Aws S3 Adapter V2
 composer require league/flysystem-aws-s3-v2
 ~~~
 
+## Usage
+
 ~~~ php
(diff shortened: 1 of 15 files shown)


frankdejonge:
@driesvints this is awesome, thanks a bunch!

Issue #451, opened 2015-03-20
FTP Adapter WINDOWS


kinncj:
It doesn't seems to work properly with `windows` systems, since the `$item` pattern is different from `unix`.

Here's the solution that I applied for a project and you probably want to fix in the same way.

```PHP
use League\Flysystem\Adapter\Ftp as BaseFtp;
use League\Flysystem\AdapterInterface;

class Ftp extends BaseFtp
{
    protected function normalizeObject($item, $base)
    {
        if (stripos(ftp_systype($this->connection), 'windows') !== false) {
            return $this->normalizeWindowsObject($item, $base);
        }

        return parent::normalizeObject($item, $base);
    }

    protected function normalizeWindowsObject($item, $base)
    {
        $type = "file";

        if (strpos($item, '<DIR>') !== false) {
            $type = "dir";

            str_replace('<DIR>', '', $item);
        }

        $item = preg_replace('#\s+#', ' ', trim($item));

        list ($date, $time, $size, $name) = explode(' ', $item, 4);
        list($month, $day, $year) = explode('-', $date);

        $timestamp = strtotime($month . ' ' . $day . ' ' . $time);
        $path = empty($base) ? $name : $base . $this->separator . $name;

        if ($type === 'dir') {
            return compact('type', 'path', 'timestamp');
        }

        $visibility = AdapterInterface::VISIBILITY_PUBLIC;
        $size = (int) $size;

        return compact('type', 'path', 'visibility', 'size', 'timestamp');
    }
}
```

frankdejonge:
This issue is a duplicate.

frankdejonge:
This is a

frankdejonge:
@kinncj could you supply some raw responses, I could create a windows ftp class (extended from the normal ftp) and perhaps eliminate this issue.

kinncj:
` 2015-01-01 09:09:09  112233 file.extension`

 `2015-01-01 09:09:09 <DIR> 112233 directory_name`

where `112233` is the size

frankdejonge:
@kinncj I see you're ignoring the year and default to public visibility. Any reasons for that? Is visibility not supported by windows ftp at all?

kinncj:
not supported @frankdejonge 

it's

```
mm-dd-yy hh:mm  type  size file/dir
```

Be aware that: if it's a directory it sends <DIR>, otherwise it's empty.

That's why I check for <DIR> in the code above, cuz then I don't need to worry about empty spaces.

[...]

PR #487, opened 2015-05-23
[FTP] Support windows

barryvdh:
This is an attempt at support for Windows FTP systems, by trying multiple parsers. (See #451)

 - First try to parse as Unix
 - If fails, parse as Windows
 - If fails, empty array (or in future, more parsers)

This modifies the Unix (current) parser by adding a check to see if the format is as expected, to prevent an exception.

Windows parser is a modified version, based on the one from @kinncj in the mentioned PR. This adds support for short+long date format. 

I've added tests for the windows lines, which only proves that it doesn't throw an exception (like before), but should probably check the desired output also.

Haven't actually tested this on a 'live' Windows server, so could use some testing.

/cc @kinncj @frederikbosch @jelrikvh @JLeitchSCO

### DIFF
--- a/src/Adapter/AbstractFtpAdapter.php
+++ b/src/Adapter/AbstractFtpAdapter.php
@@ -2,8 +2,10 @@
 
 namespace League\Flysystem\Adapter;
 
+use DateTime;
 use League\Flysystem\AdapterInterface;
 use League\Flysystem\Config;
+use League\Flysystem\NotSupportedException;
 use Net_SFTP;
 
 abstract class AbstractFtpAdapter extends AbstractAdapter
@@ -21,6 +23,7 @@ abstract class AbstractFtpAdapter extends AbstractAdapter
     protected $permPublic = 0744;
     protected $permPrivate = 0700;
     protected $configurable = [];
+    protected $systemType;
 
     /**
      * Constructor.
@@ -229,6 +232,30 @@ abstract class AbstractFtpAdapter extends AbstractAdapter
     }
 
     /**
+     * Return the FTP system type
+     *
+     * @return string
+     */
+    public function getSystemType()
+    {
+        return $this->systemType;
+    }
+
+    /**
+     * Set the FTP system type (windows or unix)
+     *
+     * @param string $systemType
+     *
+     * @return $this
+     */
+    public function setSystemType($systemType)
+    {
+        $this->systemType = strtolower($systemType);
+
+        return $this;
+    }
+
+    /**
      * {@inheritdoc}
      */
     public function listContents($directory = '', $recursive = false)
@@ -290,6 +317,27 @@ abstract class AbstractFtpAdapter extends AbstractAdapter
      */
     protected function normalizeObject($item, $base)
     {
+        $systemType = $this->detectSystemType($item);
+
+        if ($systemType === 'unix') {
+            return $this->normalizeUnixObject($item, $base);
+        } elseif ($systemType === 'windows'){
+            return $this->normalizeWindowsObject($item, $base);
+        }
+
+        throw NotSupportedException::forFtpSystemType($systemType);
+    }
+
+    /**
+     * Normalize a Unix file entry.
+     *
+     * @param string $item
+     * @param string $base
+     *
+     * @return array normalized file array
+     */
+    protected function normalizeUnixObject($item, $base)
+    {
         $item = preg_replace('#\s+#', ' ', trim($item), 7);
         list($permissions, /* $number */, /* $owner */, /* $group */, $size, /* $month */, /* $day */, /* $time*/, $name) = explode(' ', $item, 9);
         $type = $this->detectType($permissions);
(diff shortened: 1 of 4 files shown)


barryvdh:
I've added some tests to check the line + dir parser, for type/visibility/size/timestamp. @frankdejonge is this good enough?

Btw, reason I didn't use the [ftp_sytype()](http://php.net/manual/en/function.ftp-systype.php) is that I'm not really sure what the values could be. Probably not 100% safe to rely on them..

kinncj:
:+1:

frankdejonge:
@jelrikvh, I'd like your feedback on this if possible.

jelrikvh:
@frankdejonge I'll give it a go for our usecase later tonight.

barryvdh:
@frankdejonge If you want to test with a public Windows FTP, ftp.microsoft.com is open for anonymous access, using the shortdate format.

    $filesystem = new Filesystem(new FtpAdapter([
        'host' => 'ftp.microsoft.com',
        'username' => 'anonymous',
    ]));
    print_r($filesystem->listContents('deskapps/ACCESS'));

These are the listings:

	04-28-10  07:22PM       <DIR>          KB
	07-02-02  11:25AM                 2107 ReadMe1.txt

Before this PR, it would fail:

> ErrorException (E_NOTICE): Undefined offset: 8 in src/Adapter/AbstractFtpAdapter.php line 294

After the PR, this is the result:

	Array
	(
		[0] => Array
			(
				[type] => dir
				[path] => deskapps/ACCESS/KB
				[timestamp] => 1272475320
				[dirname] => deskapps/ACCESS
				[basename] => KB
				[filename] => KB
			)

		[1] => Array
			(
				[type] => file
				[path] => deskapps/ACCESS/ReadMe1.txt
				[visibility] => public
				[size] => 2107
				[timestamp] => 1025601900
				[dirname] => deskapps/ACCESS
				[basename] => ReadMe1.txt
				[extension] => txt
				[filename] => ReadMe1
			)
	)

barryvdh:
And additionally, there are other formats that might not be detected (see http://filezilla.cvs.sourceforge.net/filezilla/FileZilla/source/FtpListResult.cpp). Should those be ignored or throw an Exception?

I've returned null for now and didn't put them in the $result array, otherwise 1 failed result would make the sort fail..

src/Adapter/AbstractFtpAdapter.php
```diff
@@ -290,8 +291,32 @@ protected function sortListing(array $result)
      */
     protected function normalizeObject($item, $base)
     {
+        if ($result = $this->normalizeUnixObject($item, $base)) {
+            return $result;
+        } elseif ($result = $this->normalizeWindowsObject($item, $base)) {
+            return $result;
+        }
```
jelrikvh: I don't really like the assignment/boolean expression crossovers here. It makes stuff quite difficult to read.

barryvdh: So something like this?

        $result = $this->normalizeUnixObject($item, $base);
        
        if ( ! $result) {
            $result = $this->normalizeWindowsObject($item, $base);
        }
        
        return $result;

src/Adapter/AbstractFtpAdapter.php
```diff
@@ -290,8 +291,32 @@ protected function sortListing(array $result)
...
+     * @return array normalized file array
+     */
+    protected function normalizeUnixObject($item, $base)
+    {
         $item = preg_replace('#\s+#', ' ', trim($item), 7);
-        list($permissions, /* $number */, /* $owner */, /* $group */, $size, /* $month */, /* $day */, /* $time*/, $name) = explode(' ', $item, 9);
+        $item = explode(' ', $item, 9);
+        if (count($item) < 9) {
+            return null;
+        }
```
jelrikvh: This, like the regex check line in the ``normalizeWindowsObject``, method introduces a second responsibility to this method (namely checking if the response is eligible for parsing). I would favour a separate class with two methods, or simply checking the connection type (which is less flexible) with something like ```if (stripos(ftp_systype($this->connection), 'windows') !== false) {```

src/Adapter/AbstractFtpAdapter.php
jelrikvh: Well, in combination with my second comment below, I would opt for something like this

```php
protected function normalizeObject($item, $base)
{
    $responseType = $this->guessResponseType($item, $base);
    if (false === $responseType) {
        throw new \SomethingSomethingException("the server's FTP file list response format is not supported");
    }
    $normalizer = FtpNormalizerFactory::createNormalizerForType($responseType);
    
    return $normalizer->normalize($item, $base);
}

protected function guessResponseType($item, $base)
{
    // Ideally, this list of checks should be put away into separate classes, so we don't have to "keep a list" of all possible formats, but well, here it is.
    if ([windows format check]) {
        return 'windows';
    }

    if ([unix format check]) {
        return 'unix';
    }

    return false;
}
```

jelrikvh:
Your solution for checking which format to expect is way more elegant than mine. I just used, for the time being, a check whether the ftp server is a windows machine and acted accordingly. You, however, are checking the actual format which makes much more sense.

@frankdejonge I left some structural code-based comments but overall it's looking good. I'll now start an actual functional walkthrough.

barryvdh:
Well, ideally we just have to check the type once, instead of checking every message.
And right now, the check is just 'does it have at least 8 spaces. Ideally it would do some actual checking.

A different option would be to make it semi-configurable:

- Add 'type' config option.
- If not set, use ftp_systype() to determine if it has something like windows in it, otherwise just fall back to Unix like before.
- Add try/catch to prevent weird errors, and let the user know that the format is either undetected or unsupported.

You could use multiple classes/detectors etc, but if we assume that 90% is Unix, 9% is Windows and < 1% some exotic format (?, jst random guesses), just supporting those 2 would be good enough for now?

barryvdh:
But for the ftp_systype() approach, we would really need a lot more user data. I'm not sure what the results could be and how reliable they are.

jelrikvh:
Maybe we could use convention over configuration. We assume unix, unless something else is configured at Adapter's construction-time? I think that would make more sense than checking it at every single response coming in. Other option is introducing a property of the adapter that holds ``null`` until the first response comes in. The adapter then checks for the response type and caches it so it doesn't have to check every time.

This is, I think, a choice up to @frankdejonge. It depends on which level of automated type-guessing he would want to have.

barryvdh:
I've added it as a config option + detect on the first string, then remember. Throw exception on unsupported type.
I only detect windows and otherwise just assume unix.. Do you like this better?

src/Adapter/AbstractFtpAdapter.php
```diff
@@ -290,7 +317,31 @@ protected function sortListing(array $result)
      */
     protected function normalizeObject($item, $base)
     {
+        $systemType = $this->detectSystemType($item);
+
+        if ($systemType === 'unix') {
+            $result = $this->normalizeUnixObject($item, $base);
+        } elseif ($systemType === 'windows'){
+            $result = $this->normalizeWindowsObject($item, $base);
```
jelrikvh: I'd say "return early", but that's code style and up to @frankdejonge

jelrikvh:
Yes, I like this much better. Let me update my local usecase with the new code.

src/Adapter/AbstractFtpAdapter.php
barryvdh: Changed it to early return

jelrikvh:
It works for me, so with @frankdejonge looking into the architectural thingies mentioned above, I would wholeheartedly say: :+1:

barryvdh:
Okay cool. I've squashed the commits and let @frankdejonge decide if he wants to merge or change some things :) (I'm not 100% if I did the config thing right. I forgot to add it to configurables, but should be good now)

barryvdh:
@frankdejonge do you need anything else for this?

frankdejonge:
@barryvdh I'll review the thoroughly as soon as I get back from my holidays :+1:

barryvdh:
Ok good, no rush. Have a happy holiday!

frankdejonge:
@barryvdh could you give this PR a rebase?

barryvdh:
THis better?

PR #488, opened 2015-05-27
Doc string fixes.

twistor:
This moves all of of the docs back into FilesystemInterface from Filesystem where appropriate. It also fixes a bunch of docs that were just wrong, or copy-pasted incorrectly. Lastly, it documents all of the exceptions thrown at the moment.

### DIFF
--- a/src/FilesystemInterface.php
+++ b/src/FilesystemInterface.php
@@ -16,137 +16,165 @@ interface FilesystemInterface
     /**
      * Read a file.
      *
-     * @param string $path
+     * @param string $path The path to the file.
      *
-     * @return false|string
+     * @throws FileNotFoundException
+     *
+     * @return string|false The file contents or false on failure.
      */
     public function read($path);
 
     /**
-     * Read a file as a stream.
+     * Retrieves a read-stream for a path.
      *
-     * @param string $path
+     * @param string $path The path to the file.
      *
-     * @return false|resource
+     * @throws FileNotFoundException
+     *
+     * @return resource|false The path resource or false on failure.
      */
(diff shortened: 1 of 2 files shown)


twistor:
I discovered that Filesystem::getSize() is the only metadata method that doesn't call assertPresent(), I didn't change it as this is only documentation fixes.

src/FilesystemInterface.php
```diff
@@ -16,137 +16,165 @@ public function has($path);
...
     public function write($path, $contents, array $config = []);
 
     /**
      * Write a new file using a stream.
      *
-     * @param string   $path
-     * @param resource $resource
-     * @param array    $config   config array
+     * @param string   $path     The path of the new file.
+     * @param resource $contents The file handle.
```
frankdejonge: This should be $resource.

src/FilesystemInterface.php
```diff
@@ -16,137 +16,165 @@ public function has($path);
...
 
     /**
-     * Update a file using a stream.
+     * Update an existing file using a stream.
      *
-     * @param string   $path
-     * @param resource $resource
-     * @param array    $config   config array
+     * @param string   $path     The path of the existing file.
+     * @param resource $contents The file handle.
```
frankdejonge: This should be $resource.

frankdejonge:
@twistor this is awesome. I've added two notes of incorrect param names, when these are fixed ping me and I'll merge this. Thanks for the effort, much appreciated.

frankdejonge:
I'll just merge this and correct the param names myself, less hassle :+1:

PR #491, opened 2015-05-28
Remove dependency on mbstring.

twistor:
Sorry to be inundating you with pull requests. Thanks for taking them!

As far as I can tell, the only reason there's a dependency on mbstring is because of the mb_strlen($contents, '8bit') call in Util.

I would prefer just documenting that 'mbstring.func_overload' is unsupported, as it's only for legacy apps, but this would work and be backwards compatible.

Are there other adapters that require mbstring?

We could cache the results of ini_get() as well, but that might be overkill.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -14,8 +14,7 @@
         }
     ],
     "require": {
-        "php": ">=5.4.0",
-        "ext-mbstring": "*"
+        "php": ">=5.4.0"
     },
     "require-dev": {
         "ext-fileinfo": "*",
--- a/src/Util.php
+++ b/src/Util.php
@@ -142,7 +142,7 @@ class Util
      */
     public static function contentSize($contents)
     {
-        return mb_strlen($contents, '8bit');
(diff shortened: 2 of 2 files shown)


twistor:
For background: http://php.net/manual/en/mbstring.overload.php

The only reason to ever use mb_string( '8bit'), is if mbstring.func_overload is used. Symfony decided that it's not worth it, which I agree with. https://github.com/symfony/symfony/issues/6043

GrahamCampbell:
:-1:

GrahamCampbell:
Don't we have this extension because we use it in other places too?

twistor:
Nope, that's it.

twistor:
Turns out there's a constant for this.

http://php.net/manual/en/mbstring.constants.php

twistor:
I did a quick check through all the adapters I could find, and none of them use mbstring for anything. This makes sense. Flysystem should only be dealing with bytes, there shouldn't be any concern for the file/path content.

@frankdejonge I've squashed the branch. Thoughts?

PR #499, opened 2015-06-04
Allow for absolute paths to be used.

willrowe:
This PR does two things:
1 - Fixes the `League\Flysystem\Adapter\AbstractAdapter@removePathPrefix` method. Previously if you passed a path that did not actually contain the prefix then it would just strip the length of the prefix from the beginning of the path. In order to accomplish this the method `hasPathPrefix` was added.
2 - Allows the `League\Flysystem\Adapter\AbstractAdapter@applyPathPrefix` to accept absolute paths. This also employs the new `hasPathPrefix` method to check if it already has the prefix so that it doesn't add it again.
Below I have included an example of something I was working on today which made it somewhat hacky to get an absolute path from a Faker method and pipe it into a Filesystem method (this is in a Laravel seeder).
```php
$this->files->move($this->files->getDriver()->getAdapter()->removePathPrefix($faker->image($avatarDir)), $this->files->getDriver()->getAdapter()->removePathPrefix($avatarDir) . '/' . $user->id . '.jpg');
```
As you can see I have to get the driver from Laravel, then get the adapter from Flysystem and then finally call `removePathPrefix` (which I found out wasn't even checking if it had the prefix already). I then have to do it again on the directory path I got from the config. This is quite a lot to try to do something simple. Below is how it would look after this PR.
```php
$this->files->move($faker->image($avatarDir), $avatarDir . '/' . $user->id . '.jpg');
```
The move method will try to add the prefix, but will skip it once detecting that it's already there. This is much cleaner and what I expected the functionality to be.

### DIFF
--- a/src/Adapter/AbstractAdapter.php
+++ b/src/Adapter/AbstractAdapter.php
@@ -45,6 +45,24 @@ abstract class AbstractAdapter implements AdapterInterface
     }
 
     /**
+     * Determine if the path contains the prefix.
+     *
+     * @param string  $path
+     *
+     * @return boolean
+     */
+    public function hasPathPrefix($path)
+    {
+        $pathPrefix = $this->getPathPrefix();
+
+        if ($pathPrefix === null) {
+            return true;
+        }
+
+        return substr($path, 0, strlen($pathPrefix)) === $pathPrefix;
+    }
+
+    /**
      * Prefix a path.
      *
      * @param string $path
@@ -53,6 +71,10 @@ abstract class AbstractAdapter implements AdapterInterface
      */
     public function applyPathPrefix($path)
     {
+        if ($this->hasPathPrefix($path)) {
+            return $path;
+        }
+
         $path = ltrim($path, '\\/');
 
         if (strlen($path) === 0) {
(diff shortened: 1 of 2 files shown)


frankdejonge:
@willrowe Hi, the relative path is a design decision. It's possible to use absolute paths when you set `/` as the root path. Other parts of flysystem also rely on this relative concept, for instance the adapter response caching decorator. This change is very likely to introduce unexpected side-effects because it alters behaviour that is currently valid. The choice to work with relative paths only is not something I plan to move away from because it has many benefits and with a root of `/` it functions just the same. Therefor I won't be merging this PR. Hope you understand.

willrowe:
Is it expected behavior for the `removePathPrefix` method to not check that it's actually removing the set prefix? My intention was not to remove the use of relative paths, just to prevent the path prefix from being added twice or parts of a path being mistakenly removed.

zaak:
@willrowe I think the proposed change for checking path prefix is not valid.
What if I have path prefix set as `/foo`, and then I'm accessing relative path `/foo/bar`, so in effect I want to access `/foo/foo/bar`. With your change the prefix won't be added, and `/foo/bar` path will be used, which is incorrect.

frankdejonge:
@willoller ah I see. Since all the paths which are normalised will come from inside the adapter this should never happen, so this would be a fix for a non-existing situation.

willrowe:
Thanks for the explanation, I have a better understanding of how it's supposed to be used now.

frankdejonge:
@willoller welcome :) thanks for your your willingness to improve the package! :+1:

PR #506, opened 2015-06-12
NotSupportingVisibilityTrait should not throw exceptions.

twistor:
This might be considered an API break, while it shouldn't break any existing code, it could break someone's tests.

The fact that `NotSupportingVisibilityTrait` throws exceptions makes things more difficult to work with than necessary. For one, it's not documented on the AdapterInterface that it's possible to throw an exception. 
Another is that it ruins the abstraction of the API. To code for all adapters, I have to catch this exception, but this isn't exceptional behavior. Does this logic extend to all methods on the AdapterInterface? Theoretically, I would have to defend against a LogicException with every method call.

LogicExceptions imply that I have to change my code, but to keep it adapter agnostic, I can't.

I think a better approach is to treat all files as public. We just assume that if there's no visibility setting, then it's always public.

### DIFF
--- a/src/Adapter/Polyfill/NotSupportingVisibilityTrait.php
+++ b/src/Adapter/Polyfill/NotSupportingVisibilityTrait.php
@@ -2,7 +2,7 @@
 
 namespace League\Flysystem\Adapter\Polyfill;
 
-use LogicException;
+use League\Flysystem\AdapterInterface;
 
 trait NotSupportingVisibilityTrait
 {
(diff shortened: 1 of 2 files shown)


fruitl00p:
Logically speaking: if an adapter doesn't have the support for visibility defaulting to public _or_ private would both be wrong... ? Thus logically speaking the exception would be the only remaining choice right?

twistor:
I disagree.

The adapter is irrelevant. The file system the adapter represents still has some concept of visibility. Either people can access it, or they can't.

What visibility you default to doesn't really matter. Making it default to public is better from a security perspective.

Another option would be to always return false from setVisibility(), but getVisibility() should always return a value.

fruitl00p:
@twistor I agree on always returning false on the setter, but still returning the exception on the getter seems a wiser choice... (or just plain NULL as its undefined since it's not supported... but essentially the same meaning as the actual exception..)

Oh well... @frankdejonge whats your thoughts?

twistor:
I guess my real problem here was throwing `LogicException` since there's no way to handle adapters generically, so it's not a problem with the code.

Changing that would be a BC break.

PR #507, opened 2015-06-12
Added config for file_put_contents flags

dogiedog:
For a project we encountered a problem with the local adapter, our mounted storage (nfs share) didn't support exclusive locking. 

In the pull request you can set the flags for the file_put_contents in the config.

### DIFF
--- a/src/Adapter/Local.php
+++ b/src/Adapter/Local.php
@@ -26,11 +26,17 @@ class Local extends AbstractAdapter
     protected $pathSeparator = DIRECTORY_SEPARATOR;
 
     /**
+     * @var int
+     */
+    protected $writeFlags;
+
+    /**
      * Constructor.
      *
      * @param string $root
+     * @param int $writeFlags
      */
-    public function __construct($root)
+    public function __construct($root, $writeFlags = LOCK_EX)
     {
         $realRoot = $this->ensureDirectory($root);
 
@@ -39,6 +45,7 @@ class Local extends AbstractAdapter
         }
 
         $this->setPathPrefix($realRoot);
+        $this->writeFlags = $writeFlags;
     }
 
     /**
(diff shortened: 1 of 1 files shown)


rdrenth:
+1

GrahamCampbell:
:-1: as this config isn't available in all adapters. It's WRONG to call the write method with config specific to the adapter in question.

GrahamCampbell:
Maybe we need to make this locking configurable for the entire adapter instance?

dogiedog:
I pushed a new approach for the flags in the local adapter.

frankdejonge:
I'd say making it a constructor arguments would make this better. Agree @GrahamCampbell ?

GrahamCampbell:
Yeh, that's what I was meaning @frankdejonge. :)

src/Adapter/Local.php
```diff
@@ -402,4 +407,15 @@ protected function getDirectoryIterator($path)
 
         return $iterator;
     }
+
+    /**
+     * @param int $flags
+     * @return $this
+     */
+    public function setFlags($flags)
```
GrahamCampbell: :-1: I think this should be immutable. Should be configured on adapter instantiation.

frankdejonge: Agreed, I unfortunately made this mistake for some other settings (like in the ftp stuff) and I don't like it.

dogiedog:
Like this ? :)

frankdejonge:
Indeed, like that. Now I think we need to get a better name for this. Possibly `writeFlags`. Suggestions @GrahamCampbell ?

GrahamCampbell:
Either is fine IMO @frankdejonge. :)

frankdejonge:
@dogiedog can you update the param name to writeFlags and squash the commits?

dogiedog:
Done

frankdejonge:
@dogiedog can you squash the commits too?

dogiedog:
Done

PR #512, opened 2015-06-18
[DX] Clearify usage of Dropbox Client params

davidfuhr:
It wasn't clear to me which params to use. So I thought it also might be useful to others to know..


frankdejonge:
Thanks! This is really helpful :)

PR #520, opened 2015-07-01
Allow retrieving the Filesystem from the Handler.

RyanThompson:
This is an absolute must for any project approaching complex!

Here's one particular issue I ran into: https://github.com/anomalylabs/files-module/blob/master/src/Adapter/AdapterFilesystem.php#L193

That class wraps the Filesystem class ONLY to keep a database in sync. The disk MUST be set on the Filesystem because there are different disks (Laravel disks) using different adapters and I need to know if file1 is on disk A or B for instance when I sync. 

The problem here is that now that I have a hard dependency on the disk (in construct) I have a hard dependency on the Filesystem. And when copying or moving from the mount manager - there very well may be a different source Filesystem. Sure, I could resolve it from the path, but it makes a ton more sense having it on the Handle. OOP FTW!

Thanks for considering!

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -3,6 +3,7 @@
 ## 1.0.5 - UNRELEASED
 
 * [Filesystem::listContent] Emulated directories didn't respect the natural sorting, this is now corrected in the listContents method.
+* [Handler] Added getter for the Filesystem.
 
 ## 1.0.4 - 2015-06-07
 
--- a/src/Handler.php
+++ b/src/Handler.php
@@ -71,6 +71,16 @@ abstract class Handler
 
         return $this;
     }
+    
+    /**
+     * Retrieve the Filesystem object.
+     *
+     * @return FilesystemInterface
+     */
+    public function getFilesystem()
+    {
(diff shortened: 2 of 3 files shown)


RyanThompson:
Ok ok it looks like the mount manager uses the filesystems directly using the common methods but STILL.. I think this would be convenient in general lol.

Cheers!

frankdejonge:
@RyanThompson if you can add a test + a changelog entry for this I'll merge this in.

RyanThompson:
Done, thanks! When do you think you will tag 1.0.5?

frankdejonge:
@RyanThompson I'm looking to release a new version around Wednesday.

RyanThompson:
Fantastic! I'll have a few code changes to reflect this then ^_^

Sent from my iPhone

On Jul 6, 2015, at 3:15 AM, Frank de Jonge <notifications@github.com<mailto:notifications@github.com>> wrote:


@RyanThompson<https://github.com/RyanThompson> I'm looking to release a new version around Wednesday.

-

PR #524, opened 2015-07-07
Fix Filesystem::listContents().

twistor:
Browsing through the code, and found some broken stuff.

1. Really need to stop using empty() on strings. It's always wrong.
2. Missing directory separator for sub-path comparison. Meaning, sub/dir will match sub/dir2. This also makes $recursive irrelevant when used with a non-empty $directory argument.

I tried to make the code somewhat legible by breaking up the filter and mapping functions. Any feedback is appreciated.

 I'll get around to adding test cases later in the week, just thought I'd throw this up here.

### DIFF
--- a/src/Filesystem.php
+++ b/src/Filesystem.php
@@ -281,19 +281,24 @@ class Filesystem implements FilesystemInterface
     {
         $directory = Util::normalizePath($directory);
         $contents = $this->getAdapter()->listContents($directory, $recursive);
-        $mapper = function ($entry) use ($directory, $recursive) {
-            if (
-                empty($entry['path'])
-                || (! empty($directory) && strpos($entry['path'], $directory) === false)
-                || ($recursive === false && Util::dirname($entry['path']) !== $directory)
-            ) {
+
+        $filter = function (array $entry) use ($directory, $recursive) {
+            if ($entry['path'] === '') {
                 return false;
             }
 
-            return  $entry + Util::pathinfo($entry['path']);
+            if ($recursive) {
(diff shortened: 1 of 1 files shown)


twistor:
None of this is so great. We shouldn't be doing any filtering at all in Filesystem::listContents(). That should be up to the adapter. That's probably a separate issue though.

twistor:
Case in point the above code is more-or-less copied from MemoryAdapter::doListContents().

src/Filesystem.php
```diff
@@ -281,19 +281,24 @@ public function listContents($directory = '', $recursive = false)
...
         $contents = $this->getAdapter()->listContents($directory, $recursive);
-        $mapper = function ($entry) use ($directory, $recursive) {
-            if (
-                empty($entry['path'])
-                || (! empty($directory) && strpos($entry['path'], $directory) === false)
-                || ($recursive === false && Util::dirname($entry['path']) !== $directory)
-            ) {
+
+        $filter = function (array $entry) use ($directory, $recursive) {
+            if ($entry['path'] === '') {
```
frankdejonge: This should also check for `false` since this happens when adapters list things that are outside of the adapter root, which in some filesystems is expected behaviour. This is also why empty was used, anything false is wrong, now you're just testing for empty string, but not other variations of "incorrect". So if you really want to use this, add a check for `is_string`.

frankdejonge:
@twistor I thought I'd merge this in, but the code-coverage dropped, there's a use of an undeclared variable because the new situation wasn't properly covered by tests. I've altered my implementation to include the directory separator in the check.

PR #525, opened 2015-07-08
Distinguish between file and directory permissions in the local adapter

### LINKED ISSUE #522: Default directory permissions of the local adapter
The local adapter defaults to `0777` when setting the permissions of [the root directory](https://github.com/thephpleague/flysystem/blob/c732c136963c5d4cdf4a678becc571220a3bf6bf/src/Adapter/Local.php#L62) or any [new directory](https://github.com/thephpleague/flysystem/blob/c732c136963c5d4cdf4a678becc571220a3bf6bf/src/Adapter/Local.php#L310). Shouldn't it default to the safer `0755`? A publicly writable directory on a shared host is a huge security risk.

In addition to that, the permissions could be configured by the `visibility` here, too, with `0755` being public and `0700` being private. What do you think?

mzur:
Closes #522.

This implements the permissions `0755` (public) and `0700` (private) for directories in contrast to `0744` and `0700` for files. Previously, files and directories were treated the same but `0744` is useless for directories.

Also, this sets `0755` instead of the insecure `0777` as default permissions for all newly created directories (including the root directory).

### DIFF
--- a/src/Adapter/Local.php
+++ b/src/Adapter/Local.php
@@ -16,8 +16,14 @@ use SplFileInfo;
 class Local extends AbstractAdapter
 {
     protected static $permissions = [
-        'public' => 0744,
-        'private' => 0700,
+        'file' => [
+            'public' => 0744,
+            'private' => 0700,
+        ],
+        'dir' => [
+            'public' => 0755,
+            'private' => 0700,
+        ]
     ];
 
     /**
@@ -59,7 +65,7 @@ class Local extends AbstractAdapter
     {
         if (! is_dir($root)) {
             $umask = umask(0);
-            mkdir($root, 0777, true);
+            mkdir($root, static::$permissions['dir']['public'], true);
             umask($umask);
         }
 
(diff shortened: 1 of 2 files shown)


twistor:
Awesome! I've been meaning to do something similar for a while.
- Why do we make the files executable?
- What about making this configurable in the adapter?

mzur:
> Why do we make the files executable?

Its not the files but the directories that get executable. That's needed for entering and accessing files in them.

> What about making this configurable in the adapter?

You can [set the visibility with the config](https://github.com/mzur/flysystem/blob/9ac4e6290fb606f3ba2a62090e9224219b56d277/src/Adapter/Local.php#L316) if that is what you mean.

twistor:
You're setting public files to 0744, which is executable. Standard webserver permissions are 0644 for files and 0755 for directories. However, another common configuration is 0664 and 0775.

Yes, I know I can set the visibility. I mean, why not make the permissions 0755, etc. configurable in the constructor.

The reason for needing configuration, for example, is to support the other common use case of group-writable files. It would also help when integrating with frameworks. I'm working on a Drupal module atm, and Drupal supports configuring the default directory permissions of new directories. It would be great if I could pass that along to the adapter.

frankdejonge:
@mzur thanks! Will tag this so it'll ship right away.

mzur:
@twistor I see. The `0744` file permissions were already there before and I didn't want to mess with this since I assumed there was a good reason for it. My main motivation for this PR was to get rid of the default `0777` for directories but more flexible permissions would certainly be a nice to have.

@frankdejonge Thanks for the quick response :-)

frankdejonge:
@mzur 1.0.6 has been tagged.

PR #529, opened 2015-07-13
Fix listContents().

twistor:
I imagine this issue is getting pretty old, my apologies.

First off, checking for the Local adapter in Filesystem is not nice. The part of this patch that changes that needs to be discussed. I'm replacing back slashes with forward slashes. This is not technically correct, since Linux file names can have backslashes in them, but at some point we're going to have to decide what a valid file path can be. I had actually assumed that `Util::normalizePath()` already did this.

Other bugs fixed:
- Treating the int 0 as a string. (This is debatable, but I think a zero int is much closer to false than a string) This also avoids calling strlen() on every path.
- When a sub-directory matches. I.e. listContents('dir') returns 'a/dir/file.
- Fixed test coverage. The big || list was hiding missing test coverage.

Overall, the logic was very hard to follow.

### DIFF
--- a/src/Adapter/Local.php
+++ b/src/Adapter/Local.php
@@ -403,7 +403,7 @@ class Local extends AbstractAdapter
         $path = $file->getPathname();
         $path = $this->removePathPrefix($path);
 
-        return trim($path, '\\/');
+        return trim(str_replace('\\', '/', $path), '/');
     }
 
     /**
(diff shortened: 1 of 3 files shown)


frankdejonge:
@twistor nope, this is :+1: not getting old. This was on my todo list, the last PR was close but had other problems, this was on my todo list too. I've also pushed another commit which converts the path at entry point so the package consumer only has to think in unix paths.

PR #535, opened 2015-07-28
Unlink links to directories

ilijastuden:
When implementing a cleanup method I noticed that directories that linked to external resources where not cleared by `Local::deleteDir()` method. This commit fixes that particular problem (target directories are not deleted, only links are removed).


frankdejonge:
@ilijastuden thanks! I merged this, and added the missing tests. In the future, if you can, please include tests so this behaviour is represented by that as well. This helps further development a lot.

ilijastuden:
Will do, thanks :)

PR #536, opened 2015-08-01
fix negative filesizes on 32-bit systems

Christian-Krupa:
on 32-bit systems the `(int)` cast will return negative file sizes for a file with more that 2gb.

    pi@odroid-server:~$ file /usr/bin/file
    /usr/bin/file: ELF 32-bit LSB  executable, ARM, EABI5 version 1 (SYSV), dynamically linked (uses shared libs), for GNU/Linux 2.6.32, BuildID[sha1]=f8aff96d2e3830bde0d9296568c4d160f50d8378, stripped

the attached test fails with

    1) League\Flysystem\FileTests::testLargeFilesize
    Failed asserting that -1178494033 matches expected 3116473263.

### DIFF
--- a/src/Filesystem.php
+++ b/src/Filesystem.php
@@ -326,7 +326,7 @@ class Filesystem implements FilesystemInterface
             return false;
         }
 
-        return (int) $object['size'];
+        return (float) $object['size'];
     }
 
     /**
--- a/tests/FileTests.php
+++ b/tests/FileTests.php
@@ -222,6 +222,25 @@ class FileTests extends \PHPUnit_Framework_TestCase
         $this->assertEquals($size, $file->getSize());
     }
 
+    public function testLargeFilesize()
+    {
+        $adapter = $this->getMock('League\Flysystem\AdapterInterface');
+        $adapter
+            ->expects($this->once())
+            ->method('getSize')
+            ->with('large.txt')
+            ->willReturn(array (
+                'type' => 'file',
+                'path' => 'file.txt',
+                'timestamp' => 12345678,
+                'size' => (float)3116473263,
+            ));
+        $filesystem = new Filesystem($adapter);
+        /** @var File $file */
+        $size = $filesystem->getSize('large.txt');
+        $this->assertEquals((float)3116473263, $size);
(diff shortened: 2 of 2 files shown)


GrahamCampbell:
Hmmm, this isn't ideal. We're loosing the actual filesize now.

Christian-Krupa:
@GrahamCampbell can you provide an example why you lose the filesize?

GrahamCampbell:
> can you provide an example why you lose the filesize?

You are no longer retaining the actual file size. A float is just an approximation.

Christian-Krupa:
ok so how to fix my 2gb problem on a 32-bit system, any recommendation?

GrahamCampbell:
I'd say it's just a limitation. This PR breaks more than it fixes.

GrahamCampbell:
Maybe someone else has a better suggestion?

frankdejonge:
This behaviour is described in the php.net docs (http://php.net/manual/en/function.filesize.php), I don't feel Flysystem should take ownership of this problem.

Christian-Krupa:
i used "league/flysystem": "0.5.*" for more than a year with the sftp adapter. the combination runs smoothly. so my current solution is downgrading.

frankdejonge:
@Christian-Krupa what's the return type in that version?

Christian-Krupa:
here is the remote file listing via bash:

	-rw------- 1 pi pi 3121381340 Jul 25 09:11 bigfile.txt

thats the call i used in version 0.5 :

	var_dump($filesystem->listContents());
        ...
	var_dump($filesystem->getSize('bigfile.txt'));

output

	directory list dump 
        ....
	float(3121381340)

so here we got an float with the valid filesize.

but when i call the `getSize` without previouse`listContents` it also returns the negative value 

	#var_dump($filesystem->listContents());
	var_dump($filesystem->getSize('bigfile.txt'));

output

	int(-1173585956)

as you can see the internal cache seems to convert the size to an float. 
this behaviour give me the valid filesize in the old 0.5 version. but i dont think that this behaviour was  intended.

PR #542, opened 2015-08-18
Allow defining custom permissions for the Local adapter.

twistor:
I decided to make a new member variable, rather than change the existing static $permissions in case subclasses are using that.

### DIFF
--- a/src/Adapter/Local.php
+++ b/src/Adapter/Local.php
@@ -45,6 +45,11 @@ class Local extends AbstractAdapter
     protected $pathSeparator = DIRECTORY_SEPARATOR;
 
     /**
+     * @var array
+     */
+    protected $permissionMap;
+
+    /**
      * @var int
      */
     protected $writeFlags;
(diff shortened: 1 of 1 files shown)


frankdejonge:
@twistor looks good! Could you add this to the changelog too?

bcat-eu:
@twistor any way to change permissions globally after the class was instantiated? Same as in constructor but not on file or directory basis

frankdejonge:
@bcat-eu I think altering this after construction is a bad idea, but what's you use-case?

bcat-eu:
@frankdejonge laravel instantiates the class behind the scenes, if they will not change the code so that permissions are defined explicitly then defaults will be set and I'd have to update them at run time or extend the driver (that works already)

frankdejonge:
@bcat-eu I think, when this is released a more viable option would be to ask if Taylor allows this to be configured for the local driver. Or send a PR his way.

bcat-eu:
@frankdejonge all set, thanks for the support

PatrickRose:
Ping @twistor / @frankdejonge: Is there an update for this PR. A project I'm working on might need this.

frankdejonge:
@PatrickRose I'll get this ready for a release tonight (amsterdam time).

PatrickRose:
@frankdejonge Thanks! :+1:

PR #544, opened 2015-08-19
Easy mimetype lookup by extension

dshafik:


frankdejonge:
@dshafik wouldn't it be better to extract the extension related stuff so it can be called just for that purpose, in which case the original function would just be a mediator between those two?

dshafik:
Well, we can either switch to this, or just use `Util\MimeType` directly. I'm going to do the latter, feel free to close unless you want this (IMO unnecessary) change:

```php
    /**
     * Guess MIME Type based on the path of the file and it's content.
     *
     * @param string $path
     * @param string $content
     *
     * @return string|null MIME Type or NULL if no extension detected
     */
    public static function guessMimeType($path, $content)
    {
        $mimeType = MimeType::detectByContent($content);

        if (empty($mimeType) || $mimeType === 'text/plain' || $mimeType === 'application/x-empty') {
            return self::guessMimeTypeByExtension($path);
        }

        return $mimeType;
    }

    public static function guessMimeTypeByExtension($path)
    {
        $extension = pathinfo($path, PATHINFO_EXTENSION);

        if ($extension) {
            return MimeType::detectByFileExtension($extension) ?: 'text/plain';
        }

        return null;
    }

    public static function guessMimeTypeByContent($content)
    {
        return MimeType::detectByContent($content);
    }
```

PR #561, opened 2015-10-12
Ftp Adapter bug fix

BakoBako:
In function createActualDirectory ( class FTP ) in FTP.php file is bug. There is a listing of directories with php function ftp_nlist and after that we iterate them. But when ftp_nlist goes to directory that it need to be created recursively it returning false and foreach statement is falling.

function:
    protected function createActualDirectory($directory, $connection)
    {
        // List the current directory
        $listing = ftp_nlist($connection, '.');

            foreach ( $listing as $key => $item) {
                if (preg_match('~^\./.*~', $item)) {
                    $listing[$key] = substr($item, 2);
                }
            }

            if (in_array($directory, $listing)) {
                return true;
            }

        return (boolean) ftp_mkdir($connection, $directory);
    }

fix:
    protected function createActualDirectory($directory, $connection)
    {
        // List the current directory
        $listing = ftp_nlist($connection, '.');

        if( is_array( $listing ) )
        {    
            foreach ( $listing as $key => $item) {
                if (preg_match('~^\./.*~', $item)) {
                    $listing[$key] = substr($item, 2);
                }
            }

            if (in_array($directory, $listing)) {
                return true;
            }
        }

        return (boolean) ftp_mkdir($connection, $directory);
    }


frankdejonge:
@BakoBako sorry for the late reply on this PR, could you look into why there are so many extra commits on this PR?

frankdejonge:
This PR isn't targeted correctly. Please correct it and re-open.

Issue #567, opened 2015-11-04
Atomic Writes with the Local Adapter


PatrickRose:
Hi,

A quick search of the issues and PRs didn't find this: Is there a plan for the core to support atomic writes with the local file system. AFAIK this is probably only important for the Local adapter instead of most of the others.

Should I instead extend the Local Adapter for my purposes and do atomic writes that way?

frankdejonge:
@PatrickRose I'd say extend it for now, there are many considerations when implementing atomic file operations.

PatrickRose:
I'll look into sending a PR at a later date with the skeleton of what I've done so far.

I think for my purposes I'm able to get away with writing to /tmp then `rename()` to the correct. I'm not so sure about the general case (by which I mean Windows).

frankdejonge:
@PatrickRose you could use `sys_get_temp_dir` to determine the path, which should cover this case.

PatrickRose:
The issue is that `rename()` will fail on Windows if the new file exists, whereas Unix will overwrite the file according to a [comment on the man page for rename](http://php.net/manual/en/function.rename.php#60341). Dealing with that sounds hard. 

You also can't do the same thing that I've done with updateStream (namely writing to a stream opened by tmpfile then renaming it), because you can't rename an opened stream on Windows and closing it will delete the file. That sounds less hard since you can use tempnam instead.

PR #574, opened 2015-12-01
Implement Atomic write adapter

PatrickRose:
See #567 

This is a "first go" - I'm not entirely sure it's the best way of doing this, and I'm not entirely confident I've added tests in the right place. Feedback greatly appreciated :-)

### DIFF
--- a/tests/AtomicLocalAdapterTests.php
+++ b/tests/AtomicLocalAdapterTests.php
@@ -2,56 +2,23 @@
 
 namespace League\Flysystem\Adapter;
 
+require_once('LocalAdapterTests.php');
+
+function rename ($oldname, $newname, $context = null) {
+    if (strpos($newname, 'rename.fail') !== false)
+    {
+        return false;
+    }
+
+    return call_user_func_array('rename', func_get_args());
+}
+
 use League\Flysystem\Config;
 
-function fopen($result, $mode)
-{
-    if (strpos($result, 'false') !== false) {
-        return false;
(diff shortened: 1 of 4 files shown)


frankdejonge:
@PatrickRose This looks like a good first go, I'll have a more thorough look over this in the upcoming week, but it looks like I'll merge this and improve on this if needed.

tests/AtomicLocalAdapterTests.php
```diff
@@ -0,0 +1,356 @@
+<?php
+
+namespace League\Flysystem\Adapter;
+
+require_once('LocalAdapterTests.php');
+
+function rename ($oldname, $newname, $context = null) {
```
GrahamCampbell: cs

tests/AtomicLocalAdapterTests.php
```diff
@@ -0,0 +1,356 @@
...
+    if (strpos($newname, 'rename.fail') !== false)
+    {
+        return false;
+    }
+
+    return call_user_func_array('rename', func_get_args());
+}
+
+use League\Flysystem\Config;
```
GrahamCampbell: Normally, we put all imports first.

PatrickRose:
@frankdejonge I'd appreciate a ping if you do update this - I'm currently using a version of this for a current project so any improvements would be great (though I'll likely just delete ours and update our deps) plus it's useful to learn :-)

PR #583, opened 2015-12-15
Check For Failed FTP Connection to Avoid PHP Warning

sparkweb:
If `$this->connection` has failed and is a boolean, calling `ftp_systype($this->connection)` throws a PHP warning (Warning: ftp_systype() expects parameter 1 to be resource, boolean given). I propose first checking to see if `$this->connection` is a resource or a boolean/null and if it's going to fail anyway, to start the failure sooner.

### DIFF
--- a/src/Adapter/Ftp.php
+++ b/src/Adapter/Ftp.php
@@ -423,9 +423,12 @@ class Ftp extends AbstractFtpAdapter
      */
     public function isConnected()
     {
-        if ( is_bool($this->connection) || is_null($this->connection) ) {
+        if ( !is_resource($this->connection) ) {
             return false;
         }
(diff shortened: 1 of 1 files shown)


barryvdh:
You're checking is_null twice. Why not replace the original !is_null with is_resource?

frankdejonge:
@sparkweb the suggestion from @barryvdh is indeed better, check for what we need, reject anything else.

sparkweb:
Great point. I updated so that it first checks to see if it's a resource, then checks for an active connection, then returns true.

barryvdh:
Why don't you just change the one-line return?

`return is_resource($this->connection) && ftp_systype($this->connection) !== false;`

PR #584, opened 2015-12-21
Ftp Adapter bug fix

BakoBako:
In function createActualDirectory ( class FTP ) in FTP.php file is bug. There is a listing of directories with php function ftp_nlist and after that we iterate them. But when ftp_nlist goes to directory that it need to be created recursively it returning false and for-each statement is falling.

### DIFF
--- a/src/Adapter/Ftp.php
+++ b/src/Adapter/Ftp.php
@@ -307,15 +307,18 @@ class Ftp extends AbstractFtpAdapter
         // List the current directory
         $listing = ftp_nlist($connection, '.');
 
-        foreach ($listing as $key => $item) {
-            if (preg_match('~^\./.*~', $item)) {
-                $listing[$key] = substr($item, 2);
+        if( is_array( $listing ) )
+        {         
+            foreach ($listing as $key => $item) {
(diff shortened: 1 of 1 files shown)


frankdejonge:
@BakoBako Could this be fixed by converting `$listing = ftp_nlist($connection, '.');` into `$listing = ftp_nlist($connection, '.') ?: [];`?

BakoBako:
@frankdejonge Yes the for-each loop will be skipped and  in_array check will be always false. In this case it will create directories.

BakoBako:
@frankdejonge Yes it will work like that. The for-each loop will be skipped and in_array check will be always false so it will create directories.

src/Adapter/Ftp.php
```diff
@@ -305,8 +305,8 @@ public function createDir($dirname, Config $config)
     protected function createActualDirectory($directory, $connection)
     {
         // List the current directory
-        $listing = ftp_nlist($connection, '.');
-
+        $listing = ftp_nlist($connection, '.') ?: [];
+
```
GrahamCampbell: cs

PR #592, opened 2016-01-14
drop dep on deprecated prophecy-phpunit

remicollet:
Prophecy test cases are supported by PHPUnit since version 4.5.
So phpspec/prophecy-phpunit is deprecated and unneeded.

Also allow PHPUnit 5 (tested with PHPUnit 5.1.4, PHP 5.6.17 and 7.0.2)

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -18,10 +18,9 @@
     },
     "require-dev": {
         "ext-fileinfo": "*",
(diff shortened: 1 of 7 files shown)


frankdejonge:
@remicollet thanks!!

remicollet:
Thanks for quick merge.

PR #594, opened 2016-01-16
Update license year to 2016

pra85:


PatrickRose:
Is this needed? Surely the licence needs updating when there's an update to the actual code?

GrahamCampbell:
> Is this needed? Surely the licence needs updating when there's an update to the actual code?

There already is an update to the actual code.

PatrickRose:
I just get the feeling that a PR that just updates the licence file is a pointless PR.

PR #600, opened 2016-02-03
remove size from write method in null adapter

lalop:
There is no variable size so compact can't return it


frankdejonge:
@lalop nice spot!

PR #608, opened 2016-02-16
Quick and dirty fix for Util::guessMimeType(). Fixed #607

### LINKED ISSUE #607: Util::guessMimeType is unreliable
When trying to guess the mimetype of a css file with the following content:

```
.event{}
```

I get the ```text/x-asm``` mimetype instead of the expected ```text/css```.

Apparently the bug is out there because Finfo has a [bug](https://bugs.php.net/bug.php?id=53035).

mjacobus:
This fixes the mime type guessing [issue](https://github.com/thephpleague/flysystem/issues/607). Apparently Finfo cannot be trusted.

### DIFF
--- a/src/Util.php
+++ b/src/Util.php
@@ -155,7 +155,7 @@ class Util
     {
         $mimeType = MimeType::detectByContent($content);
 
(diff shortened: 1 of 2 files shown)


frankdejonge:
@mjacobus thanks! I altered the implementation a little and released 1.0.17

PR #612, opened 2016-03-09
Add Gaufrette Adapter to list of Adapters

jenkoian:
Gaufrette recently added a flysystem adapter (https://github.com/KnpLabs/Gaufrette/pull/390) so I've created an adapter going in the other direction to simply reciprocate the love (plus I think it can be useful). 

https://github.com/jenkoian/flysystem-gaufrette


frankdejonge:
@jenkoian thank you, this is brilliant. Also mad props for the Hadouken.

Issue #229, opened 2014-08-20
Undefined offset: 8 when calling getContent()


divdax:
Hey,

the connection to the ftp and uploading files is working.
But when i call the getContent() method i get an error:

```
/var/www/pstool/vendor/league/flysystem/src/Adapter/AbstractFtpAdapter.php
     * @param   string  $item
     * @param   string  $base
     * @return  array   normalized file array
     */
    protected function normalizeObject($item, $base)
    {
        $item = preg_replace('#\s+#', ' ', trim($item));
        list ($permissions, /* $number */, /* $owner */, /* $group */, $size, $month, $day, $time, $name) = explode(' ', $item, 9);
        $type = $this->detectType($permissions);
        $timestamp = strtotime($month . ' ' . $day . ' ' . $time);
```
The error is on line 259:
```
list ($permissions, /* $number */, /* $owner */, /* $group */, $size, $month, $day, $time, $name) = explode(' ', $item, 9);
```

I'm using v0.5.3. Any idea what's going wrong?

frankdejonge:
This is fixed in dev-master (wil be released in 1.0.2). STAT is replaced with rawlist.

GrahamCampbell:
@frankdejonge Better to recommend `~1.0@dev` rather than `dev-master`. `dev-master` is **always** a terrible idea.

frankdejonge:
@GrahamCampbell I didn't recommend dev master, I recommended using the upcoming 1.0.2 if you need this.

GrahamCampbell:
You said "This is fixed in dev-master". That's a version constraint. It's not the same as saying it's been fixed on the master branch, if you see what I mean. I don't want people to get confused.

tianissimo:
@frankdejonge I tried dev-master and still get this error (Undefined offset: 8) when calling listContents(). 
Here is my $item dump:
```
string(46) "01-03-15 05:10PM 100641 testfile.xml" 
```
Will it be fixed on 1.0.2? or I need to write my own adapter.
Thank you.

bogdanghervan:
I'm also getting this error (I'm using 1.0.15 here).

Here's my $item dump:
```
string(51) "drwxr-xr-x+   6 510             6 Aug  4  2011 2003"
```
The last "2013" is a directory name, but clearly I'm missing some of the parts that the normalization method expects.

Also, `ftp_systype()` returns `UNIX`.

PR #613, opened 2016-03-09
Bugfix "Undefined offset: 8" in some unix FTPs

Edgpaez:
Some FTP return "total 0" line as the first item of the `ftp_rawlist`. If this is the case the array is shifted so the actual line is used

As one can see in this picture from a client's FTP:
screen shot 2016-03-09 at 12 02 00 pm

Otherwise the following error is thrown:
`PHP error:  Undefined offset: 8 in myrepo/vendor/league/flysystem/src/Adapter/AbstractFtpAdapter.php on line 398`

This is related to [#229](https://github.com/thephpleague/flysystem/issues/229)

### DIFF
--- a/src/Adapter/Ftp.php
+++ b/src/Adapter/Ftp.php
@@ -347,6 +347,10 @@ class Ftp extends AbstractFtpAdapter
             return false;
         }
 
+        if (preg_match('/^total [0-9]*$/', $listing[0])) {
+            array_shift($listing);
+        }
+
         return $this->normalizeObject($listing[0], '');
     }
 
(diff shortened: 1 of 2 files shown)


frankdejonge:
@Edgpaez could you add a test case for this?

Edgpaez:
Tests were added @frankdejonge . Let me know if this is what you wanted.

barryvdh:
:+1: for this, but perhaps the `detectSystemType` should also verify the format before trying to parse it, or throw a more descriptive error (eg. when not enough parts are found, throw exception for unexpected format). (But this PR should still happen, as in this case the first result shouldn't be parsed anyways)

Edgpaez:
I do hate obscure errors. I tried what @barryvdh suggested here: 
https://github.com/thephpleague/flysystem/pull/628

PR #614, opened 2016-03-10
Make changelog legible

remicollet:


frankdejonge:
Hmm, I thought the hash would automatically be converted to a link.

PR #617, opened 2016-03-14
Allow finding MIME types by extension only

deviantintegral:
When using the S3 adapter, images are uploaded without a MIME type, causing S3 to serve them with the default application/octet MIME type. This breaks browsers as they force downloads of direct image URLs.

The S3 adapter is doing an is_string() check on the request body (which it fails, because it's a stream) to comply with the method signature of guessMimeType(). By making $content optional, we can detect image types based on the extension only.

I didn't see a phpcs config or CONTRIBUTING.md file, so let me know if there's style or other changes to make on this.

### DIFF
--- a/src/Util.php
+++ b/src/Util.php
@@ -148,14 +148,18 @@ class Util
     /**
      * Guess MIME Type based on the path of the file and it's content.
      *
-     * @param string $path
-     * @param string $content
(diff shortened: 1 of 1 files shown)


frankdejonge:
@deviantintegral a little more refactoring was required, which was just pushed an released. Thank you for bringing up the issue.

PR #619, opened 2016-03-16
Do not call getters internally

BenMorel:
`FileSystem` internally calls `getAdapter()` in many places, instead of just using the local `$adapter` property. PHP function calls being quite expensive, this is a waste of resources.

### DIFF
--- a/src/Filesystem.php
+++ b/src/Filesystem.php
@@ -51,7 +51,7 @@ class Filesystem implements FilesystemInterface
     {
         $path = Util::normalizePath($path);
 
-        return (bool) $this->getAdapter()->has($path);
+        return (bool) $this->adapter->has($path);
     }
 
     /**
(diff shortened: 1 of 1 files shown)


frankdejonge:
@BenMorel This "feature" was included (through a contributed PR) so allow lazy instantiation and overloading.

GrahamCampbell:
I don't think this should be merged.

> PHP function calls being quite expensive

Not really. To be fair, resolving the property from memory is just as expensive, so the saving are minimal.

BenMorel:
I didn't know that this had been introduced on purpose, I saw that by blaming the file indeed.
This is against my own best practices, and I'm not sure to understand the real value of it, but I understand if you've chosen another way for a reason.

> To be fair, resolving the property from memory is just as expensive, so the saving are minimal.

The fact is, you're doing both by calling a getter: resolving the property, and calling a function. Calling a function *is* expensive in PHP 5 (much less in PHP 7), and thousands of calls in a typical application greatly influence the performance.

Here's a good paper about this, anyway:
http://jpauli.github.io/2015/01/22/on-php-function-calls.html

frankdejonge:
@BenMorel while I appreciate the import, those changes are relative. The absolute speed gain is negligible. Closing this issue in favour of the current behaviour.

BenMorel:
No worries, again it's just a matter of personal best practices I guess. On this project at least, the time spent on I/O will exceed the time spent on CPU by an order of magnitude, so it probably does not really matter.

PR #632, opened 2016-04-07
Fix root directory creation problem

jjanvier:
To fix a weird use case. Imagine the root directory is `/foo/bar/baz`

At the creation of the Local adapter, everything's OK. The root directory `/foo/bar/baz` is created.
Then the directory `/foo/bar` is deleted and the permissions on `/foo` are changed. No possibility to read/write anymore in this directory.

Then every call to the read/write methods of the adapter will "fail" silently.


frankdejonge:
@jjanvier looks good, could you add a testcase for this as well?

jjanvier:
@frankdejonge done

Issue #648, opened 2016-05-05
isConnected returns true on timeout'd connections - FTP Adapter


dominic-horbas:
ftp_systype does not return false on a FTP Buffer timeout. Using ftp_rawlist to check if the connection is still active is better.

PR #649, opened 2016-05-05
isConnected no longer returns true for timeout'd connections

dominic-horbas:
ftp_systype does not return false on FTP Buffer timeout

This will fix issue #648


frankdejonge:
@dominic-horbas Hi, can you update the tests for this too? The fix itself looks good.

PR #651, opened 2016-05-06
Fixed #650 - MountManager::filterPrefix only fixes the first path

### LINKED ISSUE #650: MountManager::filterPrefix only fixes the first path
Hi,
Great work with Flysystem! We're using it in production. 

I have a filesystem mounted with the `assets://` prefix. When I want to rename a file with `$manager->rename("assets://dir/foo.jpg", "assets://dir/bar.jpg");`, what FlySystem does is the following:

1. It deletes `dir/foo.jpg` (that's correct)
2. It creates `bar.jpg` in the path `assets:/dir/bar.jpg`, so it **creates a folder named `assets:`**! 

I found out that the reason for this behavior is the `filterPrefix` method in `MountManager`. This method takes the method call's arguments and extracts the prefix out of the first argument (which is a path) and fixes this path (i.e. removes the prefix itself from the path). 

However, `filterPrefix` should fix **all** the paths in the argument list. Otherwise, commands such as `rename` fail (as shown above). 

I already fixed that issue and will open a PR shortly. I hope I understood this problem correctly - if I'm mistaken, just correct me or close this issue... Thank you!

afroewis:
See https://github.com/thephpleague/flysystem/issues/650

### DIFF
--- a/src/MountManager.php
+++ b/src/MountManager.php
@@ -128,6 +128,7 @@ class MountManager
             throw new LogicException('At least one argument needed');
         }
 
+        // The prefix depends on the first argument.
         $path = array_shift($arguments);
 
         if ( ! is_string($path)) {
@@ -139,6 +140,15 @@ class MountManager
         }
 
         list($prefix, $path) = explode('://', $path, 2);
+
+        // If there are more paths in the arguments, we'll fix them as well.
+        foreach ($arguments as $key => $argument) {
+            $parts = explode('://', $argument, 2);
+            if (count($parts) == 2) {
+                $arguments[$key] = $parts[1];
+            }
+        }
+
         array_unshift($arguments, $path);
 
         return [$prefix, $arguments];
--- a/tests/MountManagerTests.php
+++ b/tests/MountManagerTests.php
@@ -1,5 +1,6 @@
 <?php
 
+use League\Flysystem\AdapterInterface;
 use League\Flysystem\Filesystem;
 use League\Flysystem\MountManager;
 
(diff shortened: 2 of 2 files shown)


sdsomma:
@frankdejonge the resolution that was merged may have been too aggressive. In my scenario I (or rather [Laravel](https://github.com/laravel/framework/blob/5.2/src/Illuminate/Foundation/Console/VendorPublishCommand.php#L131)) uses the put method to write the contents of a file from one location to another,

```php
$manager->put('to://'.$file['path'], $manager->read('from://'.$file['path']));
```

The problem I am facing is the contents of the file that i am writting has `://` somewhere in the file as shown below.

```js
 /*jQuery UI - v1.10.3 - 2013-07-31
 * http://jqueryui.com
 * Copyright 2013 jQuery Foundation and other contributors Licensed MIT */
```

Now the MountManager filters the contents of the file and the filterPrefix method matches http://. The resulting contents that gets written is

```js
jqueryui.com
 * Copyright 2013 jQuery Foundation and other contributors Licensed MIT */
```

Thus, this commit fixes the problem where the passthrough method has expects two filepaths but in turn creates a problem for other methods in the filesystem class such as [`put`](https://github.com/thephpleague/flysystem/blob/847841974f42aa85917f2c4467f9aa5bba39d34d/src/Filesystem.php#L90), [`write`](https://github.com/thephpleague/flysystem/blob/847841974f42aa85917f2c4467f9aa5bba39d34d/src/Filesystem.php#L60) and [`update`](https://github.com/thephpleague/flysystem/blob/847841974f42aa85917f2c4467f9aa5bba39d34d/src/Filesystem.php#L143)

afroewis:
Oh. Even though I added a unit-test to cover your case, I think that we should improve the validation that checks if a parameter is a path, so that ONLY PATHS (and not strings containing or starting with paths) get modified. 

@frankdejonge , maybe we could use something like `filter_var` to validate the parameter? I would be happy to commit another fix (as soon as I find time, maybe today in the night)

janhartigan:
@afroewis it seems to me that the tricky thing about doing the prefix filtering at the `MountManager` level is that there aren't a lot of great ways for it to know which method parameters are paths and which aren't in the `Filesystem`, but as it's built the `MountManager` necessarily needs to filter prefixes  to even know which `Filesystem` to pass the call through to. Even if we use `filter_var` to validate that it's URL-like, or if we use that in addition to validating that it only does filtering with prefixes defined as adapters, we run the risk of having someone try to put the contents of a file as the exact value of one of those paths.

```php
$manager->put('/some/path/to/file.txt', 'assets://dir/foo.jpg');
```

...fairly unlikely, but still a problem.

I can't think of a lot of great solutions to this problem that don't involve some level of refactoring of the public interface. Some initial (semi-bad) ideas:

1. Keep a list of methods and parameters at the `MountManager` that define which arguments should be filtered. Downside of this is that it's incredibly ugly and is prone to failure in the future. We could write tests to mitigate this problem, but that might not help protect against future, unwritten methods.

2. Use [`ReflectionClass`](http://php.net/manual/en/class.reflectionclass.php) to read all public methods on the `Filesystem` class and only filter for parameters that contain the string `path`. This has the obvious problem of introducing a logical dependency on parameter naming in `Filesystem`, even though right now it would work. This might fail in unexpected ways, but a good bunch of unit tests might help mitigate the problem

I'm open to more ideas from @frankdejonge or other collaborators. I hope there's a relatively simple fix that doesn't require a major version upgrade due to modification of the public interface.

PR #655, opened 2016-05-12
isConnected no longer returns true for timeout'd connections

hannesvdvreken:
Credits to #649 by @dominic-horbas 

ftp_systype does not return false on FTP Buffer timeout

This will fix issue #648


dsurra:
I have a problem with this change against filezilla ftp server. It hangs, it may be possible?

sorry for my English

Issue #292, opened 2014-10-21
Feature request: copy & rename methods allowing overwrite the destination file


aanton:
Hi.

Currently the copy & rename methods throw an exception if the destination file exists. It would be great if this behavior could be defined when calling the methods, perhaps with an optional parameter to allow overwriting (by default it would be false to keep backward compatibility).

@FrenkyNet what do you think?

Regards,
Armando

patkar:
:+1:

bogdanghervan:
+1

kallealberg:
+1

mediamonks-robert:
+1

justinkwanlee:
+1

briedis:
+1

anydasa:
+1

Ronnie-J:
+1

frankdejonge:
ForceCopy and ForceRename plugins are now provided, tagged and released.

Anahkiasen:
@frankdejonge What version are those tagged under, I don't see any release since June 3?

schmunk42:
@frankdejonge The `MountManager` should also get these methods.

PR #665, opened 2016-06-03
Implement copy and rename plugins that overwrite destination.

twistor:
RE: #292 

Trying to extend the interface, or default implementation, leads to problems.

Here's an implementation via plugins.

### DIFF
--- a/src/Filesystem.php
+++ b/src/Filesystem.php
@@ -8,6 +8,8 @@ use League\Flysystem\Util\ContentListingFormatter;
 
 /**
  * @method array getWithMetadata(string $path, array $metadata)
+ * @method bool  forceCopy(string $path, string $newpath)
+ * @method bool  forceRename(string $path, string $newpath)
  * @method array listFiles(string $path = '', boolean $recursive = false)
  * @method array listPaths(string $path = '', boolean $recursive = false)
  * @method array listWith(array $keys = [], $directory = '', $recursive = false)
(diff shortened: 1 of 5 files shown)


twistor:
I haven't actually tested this.

Will write tests if we decide this is the direction to go.

GrahamCampbell:
Pretty nice. :+1:

frankdejonge:
@twistor I like it! 👍

twistor:
Added return value checking for the delete() call. Makes things awkward.

franksl:
Hi,
Is it possible to forceCopy between two filesystem in a MountManager? I tried adding the plugin to the MountManager instance but plugins are called on a specific filesystem (the one indicated by the first prefix)
Thanks,
Frank

frankdejonge:
Hi @franksl, you'd need to create specific ones for the mount manager, because while the concept is the same, the actual implementation can't be made portable within the same plugin.

franksl:
Ok, thanks a lot for the answer, I'll try to work around it.

basz:
I don't believe a forcedCopy for a MountManager was ever implemented right?

frankdejonge:
@basz no they weren't. However I'm thinking of making deletes idempotent, which makes it possible to do:

```php
$fs->delete($destination);
$fs->copy($source, $destination);
```

Which IMO eliminates the need for forcesCopy entirely.

PR #669, opened 2016-06-07
Allow users of StreamedReadingTrait to determine type of stream and fix docs.

twistor:


frankdejonge:
@twistor wouldn't you then just not use this trait and implement the `readStream` differently? There's not a lot of logic in there.

twistor:
That's true. What I really wanted to do was change it to php://temp and fix the docs.

twistor:
It seems odd to have the string in memory, and then write it to a file.

frankdejonge:
@twistor we could also use `fopen('php://memory', 'r')`?

frankdejonge:
@twistor since we already have the string in memory when we use `php://memory` it'll always be in mem, the temp will write to file over a certain threshold.

twistor:
@frankdejonge I thought about that. I was worried about making a copy of `contents` if it is large.

frankdejonge:
Makes sense, let's go with this 👍

Issue #618, opened 2016-03-16
Call to a member function getTimestamp() on boolean (Incorrect format detection)


DuckThom:
I'm using laravel and the allFiles function seems to be throwing an exception in `src/Adapter/AbstractFtpAdapter.php` on line 429.

I've done some debugging and it looks like the 'correct date/time check' doesn't work correctly.
The $date variable is '03-15-2016' which is a 10 character string while the check is looking for an 8 character string.

```
var_dump($date); // 03-15-2016
var_dump($time); // 05:00PM

$format = strlen($date) === 8 ? 'm-d-yH:iA' : 'Y-m-dH:i';
```

The expected input for the $date variable is mm-dd-yy while the actual input is mm-dd-yyyy (The full year notation).

Changing the format check to `$format = strlen($date) === 10 ? 'm-d-YH:iA' : 'Y-m-dH:i';` fixes the issue for me but I suspect it would create new problems for other users if their input is mm-dd-yy.

frankdejonge:
@DuckThom what kind of ftp server are you talking to?

DuckThom:
It's not my server. Is there a way to get info about the FTP server from a client?
The only message I get back from the server is `Remote system type is Windows_NT.`

DuckThom:
I've changed the check locally to this:
```
if ($dt = DateTime::createFromFormat('m-d-YH:iA', $date . $time)) { // dd-mm-yyyy
    $timestamp = $dt->getTimestamp();
} else if ($dt = DateTime::createFromFormat('m-d-yH:iA', $date . $time)) { // dd-mm-yy
    $timestamp = $dt->getTimestamp();
} else if ($dt = DateTime::createFromFormat('Y-m-dH:i', $date . $time)) { // yyyy-mm-dd
    $timestamp = $dt->getTimestamp();
} else {
    throw new Exception("Invalid datetime format: {$date}{$time}");
}
```

This works for me but I cant test the other statements. I'd write a test for it but I'm at work right now so, unfortunately, I don't have time for that.

DuckThom:
I've made a new pull request as I've replace my previous fix with, what I think, a better solution.

PR #672, opened 2016-06-09
Make the windows date/time normalizer not cause an exception if the date could not be parsed

DuckThom:
Currently, if the date/time doesn't match either of the 2 formats. DateTime will return false and ->getTimestamp() causes an exception `Call ->getTimestamp() on boolean`.
I order to prevent this exception. I've made a check that will first check if DateTime is not false and if it is, it will try to create a timestamp using `strtotime`.
The value from `strtotime` is cast to an int so if that also returns false, it will return 0 (which is a valid timestamp).


DuckThom:
Fix for https://github.com/thephpleague/flysystem/issues/618

Issue #72, opened 2014-02-10
Use file access mode 'r' instead of 'r+' for readStream


pschirch:
Hi,

the [`Local Adapter`](https://github.com/thephpleague/flysystem/blob/master/src/Adapter/Local.php) uses file access mode `r+` in his [`readStream()`](https://github.com/thephpleague/flysystem/blob/master/src/Adapter/Local.php#L157) method.

The naming suggesting that a source is used read only. But `r+` allows [reading and writing](http://www.php.net/manual/en/function.fopen.php). In my current project I have only read access on the files, so following exception throws:

```
ContextErrorException: Warning: fopen(...): failed to open stream: Permission denied in .../vendor/league/flysystem/src/Adapter/Local.php line 157
```

The more suitable file access mode is `r`.

What's your opinion?

barryvdh:
Is `rb` not more suitable? Shouldn't matter on Linux, but does in Windows, according to http://php.net/manual/en/function.fread.php#example-2648
> Warning: On systems which differentiate between binary and text files (i.e. Windows) the file must be opened with 'b' included in fopen() mode parameter.

(Via https://github.com/laravel/framework/pull/14319 by @BePsvPT )

Frondor:
What's the actual status of this?

I'm getting this same error when trying to copy a folder 
`$manager->copy('local://'.$path, 'zip://'.$path);`

And I'm developing on windows 7, latest flysystem version.

mirfilip:
The mode is `rb` for some time now - available since `1.0.25`, see #677. @frankdejonge maybe close this issue :)

PR #677, opened 2016-07-18
Read stream in binary mode

barryvdh:
See http://php.net/manual/en/function.fopen.php and https://github.com/thephpleague/flysystem/issues/72#issuecomment-233303480
> For portability, it is strongly recommended that you always use the 'b' flag when opening files with fopen().

### DIFF
--- a/src/Adapter/Ftp.php
+++ b/src/Adapter/Ftp.php
@@ -415,7 +415,7 @@ class Ftp extends AbstractFtpAdapter
      */
     public function readStream($path)
     {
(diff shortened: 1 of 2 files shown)


barryvdh:
It seems that this was present on half of the `fopen()` calls, so not sure if this was intentionally not done, or just an oversight..

frankdejonge:
@barryvdh NAIS! Waiting for travis but will merge and tag when it's done 👍

frankdejonge:
@barryvdh tagged as 1.0.25

PR #680, opened 2016-07-22
Fixed ensureDirectory method in Local adapter - Dir seems to be not readable although it is

miloshavlicek:
FIX: no need to use realpath($root) in Local adapter for ensureDirectory method - directory already ensured by is_dir($root) (realpath function causes issues when running script without executable permissions on all directories in the hierarchy)

### DIFF
--- a/src/Adapter/Local.php
+++ b/src/Adapter/Local.php
@@ -73,13 +73,13 @@ class Local extends AbstractAdapter
     {
         $root = is_link($root) ? realpath($root) : $root;
         $this->permissionMap = array_replace_recursive(static::$permissions, $permissions);
-        $realRoot = $this->ensureDirectory($root);
+        $this->ensureDirectory($root);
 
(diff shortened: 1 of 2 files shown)


frankdejonge:
@miloshavlicek I don't think this fix is correct, there must be something else going on. On what kind of installation are you running this?

miloshavlicek:
@frankdejonge Hi Frank, my project runs on IIS 8.5 (Windows Server 2012 R2), but the same issue could occur on linux server. Project is located in dir with executable permissions for its app pool (C:\WWW\LocalUser\xyz\web\), but local storage for files is in directory like C:\WWW\LocalUser\xyz2\web\ with full permissions for the app pool.

But, my app pool does not have any permissions for C:\WWW\LocalUser\, so realpath($root) returns false, because:

> The running script must have executable permissions on all directories in the hierarchy, otherwise realpath() will return FALSE. 

source: http://php.net/manual/en/function.realpath.php

PR #681, opened 2016-07-27
Add missing return type declarations

technodelight:
`MountManager` and `Filesystem` classes had missing return type declarations for `assert*` methods. This prevents using these classes when you're doing `phpspec`. An exception is thrown stating these methods are not existing.
In short, `phpspec/prophecy` relies on `phpdocumentor`, where an empty type declaration causes an exception to be thrown (`'Attempted to resolve "' . $type . '" but it appears to be empty'`) during parsing the class docblock. This will prevent phpspec to get a list of any of the supported magic methods.

### DIFF
--- a/src/Filesystem.php
+++ b/src/Filesystem.php
@@ -376,6 +376,8 @@ class Filesystem implements FilesystemInterface
      * @param string $path path to file
      *
      * @throws FileNotFoundException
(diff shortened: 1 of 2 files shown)


frankdejonge:
@technodelight 👍

Issue #235, opened 2014-08-28
Problem with UTF-8 file/directory names


zaak:
I have following directory structure:

```
.
└── files
    ├── 繁體中文字
    │   └── test.txt
    └── 繁體中文字.txt
```

When I call `$filesystem->listContents('files')` I'm getting following output:

```
array (size=2)
  0 => 
    array (size=6)
      'dirname' => string 'files' (length=5)
      'basename' => string 'files' (length=5)
      'filename' => string 'files' (length=5)
      'path' => string 'files/繁體中文字' (length=21)
      'type' => string 'dir' (length=3)
      'timestamp' => int 1409223899
  1 => 
    array (size=8)
      'dirname' => string 'files' (length=5)
      'basename' => string '.txt' (length=4)
      'extension' => string 'txt' (length=3)
      'filename' => string '' (length=0)
      'path' => string 'files/繁體中文字.txt' (length=25)
      'type' => string 'file' (length=4)
      'timestamp' => int 1409223047
      'size' => int 0
```

As you can see `filename` and `basename` are invalid. This is caused by lack of multibyte support in `pathinfo` used in [Util::pathinfo()](https://github.com/thephpleague/flysystem/blob/master/src/Util.php#L18). Expected output is:

```
array (size=2)
  0 => 
    array (size=7)
      'dirname' => string 'files' (length=5)
      'basename' => string '繁體中文字' (length=15)
      'extension' => string '' (length=0)
      'filename' => string '繁體中文字' (length=15)
      'path' => string 'files/繁體中文字' (length=21)
      'type' => string 'dir' (length=3)
      'timestamp' => int 1409223899
  1 => 
    array (size=8)
      'dirname' => string 'files' (length=5)
      'basename' => string '繁體中文字.txt' (length=19)
      'extension' => string 'txt' (length=3)
      'filename' => string '繁體中文字' (length=15)
      'path' => string 'files/繁體中文字.txt' (length=25)
      'type' => string 'file' (length=4)
      'timestamp' => int 1409223047
      'size' => int 0
```

ThaDafinser:
@zaak did you had this issue under windows or unix?

zaak:
@ThaDafinser Linux. On Windows it fails completely if it's about UTF-8 names, but it's PHP itself at fault, not the Flysystem.

ThaDafinser:
@zaak thank you.

Just tried on PHP7 and still not solved.

I found this for windows (in case someone comes around here):
[...]

PR #682, opened 2016-08-03
Implement custom pathinfo() functions.

twistor:
As mentioned in #235, `pathinfo()` has inconsistent handling of UTF-8 file paths.

This has been a known problem for quite some time, and I'm surprised it's taken this long to get reported.

I know that Drupal implements their own version of `basename()` for this reason.

See http://bugs.php.net/bug.php?id=37738

This is a clean-room implementation of the required methods.

### DIFF
--- a/src/Util.php
+++ b/src/Util.php
@@ -16,9 +16,15 @@ class Util
      */
     public static function pathinfo($path)
     {
-        $pathinfo = pathinfo($path) + compact('path');
-        $pathinfo['dirname'] = array_key_exists('dirname', $pathinfo)
-            ? static::normalizeDirname($pathinfo['dirname']) : '';
+        $pathinfo = compact('path');
+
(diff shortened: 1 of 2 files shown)


frankdejonge:
@twistor Hi, it looks good, but we need to take the windows behaviour into account. Because on windows it separates on `/` as wel as on `\`.

twistor:
So, it looks like 'extension' and 'filename' work correctly with utf-8 characters as long as we get basename() correct.

twistor:
@frankdejonge I'm not sure about the codeCoverageIgnore* bits. Other than that, it should be good to go.

Some of the test cases might be redundant, not sure.

Also, sorry if anyone got their inbox filled. I was using appveyor as a testing environment for Windows.

twistor:
@frankdejonge ping

Issue #503, opened 2015-06-10
filter criteria


kanariezwart:
First, thanks and props for this awesome library. :bow: 
When calling `listContents` it should always return all contents by default. Makes sense. 
Is there a way of finding files by filter? Is this functionality already present and did I fail to find it? I created a simple plugin to filter based on regex, and that got me thinking to create a filter plugin that can filter the content list by multiple criteria:

Say I want to find/filter all entries that match a certain type, base name, filesize.

``` php
$criteria[] = ['basename' => '*something*']; // simple filter on base name
$criteria[] = ['basename' => ['regex' => '/bla_.*/']]; // regex like filter
$criteria[] = ['extension' => '.txt']; // filter on file type or mime type
$criteria[] = ['size' => '< 1024' ];  // filter on size

// fetch only entries that match filter
$contents = $filesystem->listContentsBy($criteria) ?
```

Would there be any appreciation for such a plug-in? Or is this beyond the scope of Flysystem?
Any thoughts?

barryvdh:
Something like the [Symfony Finder](https://github.com/symfony/Finder) perhaps, a fluent interface?

`$filesystem->finder()->name('photos*')->size('< 100K')->date('since 1 hour ago')->in('/media');`

kanariezwart:
@barryvdh a finder plugin like symfony finder, sounds like a great start. I'll have a look.  :+1:

frankdejonge:
Hi all, I'd say a package which relies on flysystem would be the way to do. I don't see flysystem ship with this itself since the majority of the adapters can't support, leading to segmented feature coverage (which is what I want to avoid). The way to go would be to decorate the response, manipulate the output from the listContents call. Because it's a decoration it would be relatively easy to supply it as a companion package instead of in the main package. If you guys have any objections or feel like I'm way off in this conclusion, let me know and we can discuss.

barryvdh:
Why wouldn't most adapters not support it? Only downside would that the filtering would ofcourse be _after_ the contents are fetched, but most adapters support timestamps/size right? So it would basically provide filtering after the listContents() result.

frankdejonge:
@barryvdh as you stated, it would probably be a "post listing" operation. Mapping all those criteria's to actual listing calls sounds pretty magical to me. Though if you have idea's how to implement this, I'm very much open to suggestions.

barryvdh:
[...]

PR #689, opened 2016-08-30
Listing filtered contents mechanism, set of example filters, example …

daryush:
…usage with Local adapter

| Q | A |
| --: | --- |
| Bug fix? | no |
| New feature? | yes |
| BC breaks? | no |
| Deprecations | no |
| Tests pass? | yes |
| Issue | #503 |

It's proposition for #503. 

There is new interface to implement to avoid BC breaks and allow filtering. Mechanism prefer filtering inside adapter but when adapter doesn't support filtering it is done by the filesystem. There is implemented filtering inside Local adapter as an example.

Filtering is made by Specification patter and that allows to easily adding new type of filters without need to build or modify huge tree of if's an else's.

There are implemented set of example filter criteria which operates on file path, name, size, timestamp, visibility, type, extension. Also logical operation on filters are implemented to make possible easy combining of criteria.

If some of operation are unsupported because of lack of file information given by adapter then exception is thrown.

Examples of usage

Filtering by filename

```
$filesystem->listFilteredContents(
    new FilenameEquals('sample_file')
);
```

JPG files modified before now

```
$filesystem->listFilteredContents(
    new AndX(
        new ExtensionEquals('jpg'),
        new ModifiedBefore(new \DateTime('now'))
    )
);
```

PNG or JPG files bigger than 1024 bytes

```
$filesystem->listFilteredContents(
    new AndX(
        new SizeBiggerThan(1024),
        new OrX(
            new ExtensionEquals('jpg')
            new ExtensionEquals('png')
        )
    )
);
```

Creation of new filters
It's simple as creation of new class that implements `FilterCriteriaInterface`

```
<?php

namespace League\Flysystem;

interface FilterCriteriaInterface
{
    /**
     * Determines that file info satisfying criteria or not.
     *
     * @param FilterFileInfo $filterFileInfo
     *
     * @return bool
     */
    public function isSatisfiedBy(FilterFileInfo $filterFileInfo);
}
```

Example of creating new criteria as combination of existing ones

```
<?php

namespace League\Flysystem\Filter;

use League\Flysystem\FilterCriteriaInterface;
use League\Flysystem\FilterFileInfo;

class IsImage implements FilterCriteriaInterface
{
    /**
     * @inheritdoc
     */
    public function isSatisfiedBy(FilterFileInfo $filterFileInfo)
    {
        return (new AndX(
            new IsFile(),
            new OrX(
                new ExtensionEquals('jpg'),
                new ExtensionEquals('png'),
                new ExtensionEquals('gif'),
            )
        ))->isSatisfiedBy($filterFileInfo);
    }
}
```

### DIFF
--- a/spec/FilesystemSpec.php
+++ b/spec/FilesystemSpec.php
@@ -2,6 +2,8 @@
 
 namespace spec\League\Flysystem;
 
+use League\Flysystem\FilterFileInfo;
+use League\Flysystem\UnsupportedFilterException;
 use PhpSpec\ObjectBehavior;
 use Prophecy\Argument;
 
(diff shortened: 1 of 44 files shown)


frankdejonge:
This is too complex in order to generalise, I've decided not to want to support this.

PR #696, opened 2016-09-16
Ability to append file contents

chris1312:
This commit adds support for Filesystem class to append file contents.

I haven't modified the AdapterInterface in order to not break BC.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,10 @@
+._*
+.~lock.*
+.buildpath
+.DS_Store
+.idea
+.project
+.settings
 vendor
 bin
 coverage
--- a/src/Adapter/Local.php
+++ b/src/Adapter/Local.php
@@ -185,6 +185,7 @@ class Local extends AbstractAdapter implements AppendableAdapterInterface
     public function append($path, $contents, Config $config)
     {
         $location = $this->applyPathPrefix($path);
+        $this->ensureDirectory(dirname($location));
         $mimetype = Util::guessMimeType($path, $contents);
         $size = file_put_contents($location, $contents, FILE_APPEND);
 
--- a/src/Adapter/Local.php
+++ b/src/Adapter/Local.php
@@ -182,6 +182,55 @@ class Local extends AbstractAdapter implements AppendableAdapterInterface
     /**
      * @inheritdoc
      */
+    public function append($path, $contents, Config $config)
+    {
+        $location = $this->applyPathPrefix($path);
+        $mimetype = Util::guessMimeType($path, $contents);
+        $size = file_put_contents($location, $contents, FILE_APPEND);
+
+        if ($size === false) {
+            return false;
+        }
+
+        return compact('path', 'contents', 'mimetype');
+    }
+
+    /**
+     * @inheritdoc
+     */
+    public function appendStream($path, $resource, Config $config)
+    {
+        $location = $this->applyPathPrefix($path);
+        $this->ensureDirectory(dirname($location));
+        $stream = fopen($location, 'a');
+
+        if (!$stream) {
+            return false;
+        }
+
+        if (!flock($stream, LOCK_EX)) {
+            throw new Exception('Could not get file lock');
+        }
+
+        stream_copy_to_stream($resource, $stream);
+
+        fflush($stream);
+        flock($stream, LOCK_UN);
+
+        if (!fclose($stream)) {
+            return false;
+        }
+
+        if ($visibility = $config->get('visibility')) {
+            $this->setVisibility($path, $visibility);
+        }
+
+        return compact('path', 'visibility');
+    }
+
+    /**
+     * @inheritdoc
+     */
     public function updateStream($path, $resource, Config $config)
     {
         return $this->writeStream($path, $resource, $config);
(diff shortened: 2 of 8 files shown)


src/Filesystem.php
```diff
@@ -202,6 +202,27 @@ public function readStream($path)
     /**
      * @inheritdoc
      */
+    public function appendStream($path, $resource, array $config = [])
+    {
+        $adapter = $this->getAdapter();
+
+        if (!$adapter instanceof AppendableAdapterInterface) {
+            throw new Exception('Filesystem adapter does not support file append');
+        }
```
twistor: Would it make sense to do a read + append + write here if append is not supported?

At least this should be a more specific exception. InvalidMethodCallException?

src/AppendableAdapterInterface.php
```diff
@@ -0,0 +1,15 @@
...
+{
+    /**
+     * Append file stream
+     *
+     * @param string $path
+     * @param resource $resource
+     * @param \League\Flysystem\Config $config
+     * @return array|false
+     */
+    public function appendStream($path, $resource, Config $config);
```
twistor: Why isn't there a corresponding `append()` that accepts a string?

src/Filesystem.php
```diff
@@ -202,6 +202,27 @@ public function readStream($path)
...
+
+        if (!is_resource($resource)) {
+            throw new Exception(__METHOD__ . ' expects argument #2 to be a valid resource.');
+        }
```
twistor: This should be InvalidArgumentException.

src/FilesystemInterface.php
```diff
@@ -126,6 +126,20 @@ public function write($path, $contents, array $config = []);
...
+     * @param array    $config   An optional configuration array.
+     *
+     * @throws InvalidArgumentException If $resource is not a file handle.
+     * @throws FileExistsException
+     *
+     * @return bool True on success, false on failure.
+     */
+    public function appendStream($path, $resource, array $config = []);
+    
+    /**
```
twistor: We can't change the FilesystemInterface. That's a BC break.

chris1312: So as a solution could be additional interface with those methods, like for adapters?

src/AppendableAdapterInterface.php
chris1312: OK, I'll add it!

chris1312:
I have done some updates in order to avoid BC breaks.

src/InvalidMethodCallException.php
```diff
@@ -0,0 +1,6 @@
+<?php
+namespace League\Flysystem;
+
+class InvalidMethodCallException extends Exception
```
twistor: BadMethodCallException is a PHP builtin.

src/Adapter/Local.php
```diff
@@ -181,6 +182,55 @@ public function readStream($path)
     /**
      * @inheritdoc
      */
+    public function append($path, $contents, Config $config)
+    {
+        $location = $this->applyPathPrefix($path);
+        $mimetype = Util::guessMimeType($path, $contents);
+        $size = file_put_contents($location, $contents, FILE_APPEND);
```
twistor: Does this need an `ensureDirectory()` check?

src/Filesystem.php
```diff
@@ -202,6 +202,48 @@ public function readStream($path)
     /**
      * @inheritdoc
      */
+    public function append($path, $contents, array $config = [])
+    {
+        $adapter = $this->getAdapter();
+
+        if (!$adapter instanceof AppendableAdapterInterface) {
+            throw new InvalidMethodCallException('Filesystem adapter does not support file append');
+        }
```
twistor: I think a fallback to read + write makes more sense here, since that's what the user will do in this case anyway.

chris1312:
Yes, S3 does not support "append" feature, same as Google.

But other adapters could support it. like Local or FTP.

As I added AppendableAdapterInterface, so not every Adapter must implement it.

chris1312:
As I understand, tests are failing not because of my commit?

src/Adapter/Local.php
```diff
@@ -181,6 +182,56 @@ public function readStream($path)
...
+    }
+
+    /**
+     * @inheritdoc
+     */
+    public function appendStream($path, $resource, Config $config)
+    {
+        $location = $this->applyPathPrefix($path);
+        $this->ensureDirectory(dirname($location));
+        $stream = fopen($location, 'a');
```
jhdxr: I think the mode should be `ab` since the php strongly recommend always using `b`, and `w+b` is used for `writeStream`

src/Adapter/Local.php
```diff
@@ -181,6 +182,56 @@ public function readStream($path)
     /**
      * @inheritdoc
      */
+    public function append($path, $contents, Config $config)
+    {
+        $location = $this->applyPathPrefix($path);
+        $this->ensureDirectory(dirname($location));
+        $mimetype = Util::guessMimeType($path, $contents);
+        $size = file_put_contents($location, $contents, FILE_APPEND);
```
jhdxr: I think this should be `$size = file_put_contents($location, $contents, $this->writeFlags | FILE_APPEND);`
see what `write` does.

src/Adapter/Local.php
```diff
@@ -181,6 +182,56 @@ public function readStream($path)
...
+    {
+        $location = $this->applyPathPrefix($path);
+        $this->ensureDirectory(dirname($location));
+        $stream = fopen($location, 'a');
+
+        if (!$stream) {
+            return false;
+        }
+
+        if (!flock($stream, LOCK_EX)) {
```
jhdxr: I doubt if this lock is needed, since I think this should keep the same like what `writeStream` does.

src/FilesystemInterface.php
```diff
@@ -2,6 +2,8 @@
 
 namespace League\Flysystem;
 
+use InvalidArgumentException;
```
jhdxr: all the changes in this file are not necessary

Stadly:
What about creating a simulated `append` for the adapters that don't support append:

    public function append($path, $contents) {
       $oldContents = $this->has($path) ? $this->read($path) : '';
       $this->put($path, $oldContents.$contents);
    }

And then of course true appending for the adapters that do support it.

.gitignore
```diff
@@ -1,3 +1,10 @@
+._*
```
twistor: These changes should be added to your global gitignore. Not relevant to this patch.

peinhu:
> creating a simulated `append` for the adapters

@Stadly I think it's a good idea, someone should take a good look at this.

PR #701, opened 2016-09-27
CaseSensitivity fix for Dropbox

furkanayata:
When renaming directories in Dropbox (from and to upper to lower case) it causes path comparision bug to discard existing directories in Listing.
This commit is an easy fix on that bug

### DIFF
--- a/src/Util/ContentListingFormatter.php
+++ b/src/Util/ContentListingFormatter.php
@@ -97,7 +97,7 @@ class ContentListingFormatter
      */
     private function isDirectChild(array $entry)
     {
-        return Util::dirname($entry['path']) === $this->directory;
+        return strtolower(Util::dirname($entry['path'])) === strtolower($this->directory);
     }
 
     /**


src/Util/ContentListingFormatter.php
```diff
@@ -97,7 +97,7 @@ private function residesInDirectory(array $entry)
      */
     private function isDirectChild(array $entry)
     {
-        return Util::dirname($entry['path']) === $this->directory;
+        return strtolower(Util::dirname($entry['path'])) === strtolower($this->directory);
```
twistor: This should probably be a UTF-8 aware check, since the Dropbox API expects UTF-8.

Since mbstring isn't required, we'll probably need a check for that too.

twistor:
I didn't notice before, but this needs to go in the Dropbox repository. Most file systems are case sensitive, so this would break most file systems.

twistor:
This should be dealt with directly in the Dropbox adapter.

PR #702, opened 2016-10-03
add return types to API documentation

zburke:
Update API documentation by showing return types for all methods.


frankdejonge:
@zburke hi, these return types aren't correct. They are the ones from the adapters, not the Filesystem class.

Issue #563, opened 2015-10-13
FTP adapter bug: $filesystem->has() returns false


flylord:
Similar to https://github.com/thephpleague/flysystem/issues/539
Next command return false if I am testing regular file. It is working for directories  

``` php
Context::getContext()->fs->has('ftp://' . $testFile);
```

My system:
vsFTPd 3.0.2
"league/flysystem": "dev-master"
PHP Version 5.6.4-4ubuntu6.3

If I do some small change in getMetadata() in ftp adapter, function has() is working

``` php
    public function getMetadata($path)
    {
        $connection = $this->getConnection();

        if ($path === '') {
            return ['type' => 'dir', 'path' => ''];
        }

        if (@ftp_chdir($connection, $path) === true) {
            $this->setConnectionRoot();

            return ['type' => 'dir', 'path' => $path];
        }

        $rootDir = ftp_pwd($connection);
        ftp_chdir($connection, dirname($path));

        $listing = ftp_rawlist($connection, basename($path));

        ftp_chdir($connection, $rootDir);

        if (empty($listing)) {
            return false;
        }

        return $this->normalizeObject($listing[0], '');
    }
```

frankdejonge:
@flylord could you look if the Ftpd adapter works for you? I see you're using FTPd instead of FTP.

flylord:
@frankdejonge Ftpd adapter is not working

flylord:
I forgot to put my ftp adapter setting:
```php
$ftp = new Filesystem(new FtpAdapter([
    'host' => FTP_HOST,
    'username' => FTP_USER,
    'password' => FTP_PASSWD,

    /** optional config settings */
    'port' => 21,
    'root' => FTP_UPLOAD_PATH,
    'passive' => true,
    'ssl' => false,
    'timeout' => 30,
]));
```

flylord:
By the way, even with pure PHP, using only http://php.net/manual/en/ref.ftp.php , I have to do chdir() before I can check with ftp_rawlist if some regular file exists

flylord:
I did some more testing. Function listDirectoryContents($directory, true) is working only from Ftpd but not from Ftp adapter. When I find out some info more, I will post it here. I don't have any deep knowledge about ftp servers, so I am slow with solving or finding cause of this problem  :(

a-harvie:
I've got the same problem, using `vsftpd: version 3.0.2`.

[...]

PR #708, opened 2016-10-10
Fix: escape space characters for ftp adapter

RubtsovAV:
Bug example:

```
$filesystem->has('that path/have/the space character.txt');
// Return false every time if in path have space character.
```

Because before call ftp_rawlist we need add a slash in front of each space character in path.

```
$options = '-A';
$path = str_replace(' ', '\ ', $path);
$result = ftp_rawlist($connection, $options . ' ' . $path);
```

### DIFF
--- a/src/Adapter/Ftp.php
+++ b/src/Adapter/Ftp.php
@@ -512,7 +512,11 @@ class Ftp extends AbstractFtpAdapter
 
     protected function detectFtpServerName()
     {
-        $response = ftp_raw($this->getConnection(), 'HELP');
+        if (!$this->connection) {
+            return null;
+        }
+        
+        $response = ftp_raw($this->connection, 'HELP');
         if (!$response) {
             return null;
         }
--- a/src/Adapter/Ftp.php
+++ b/src/Adapter/Ftp.php
@@ -353,7 +353,7 @@ class Ftp extends AbstractFtpAdapter
             return ['type' => 'dir', 'path' => $path];
         }
 
-        $listing = ftp_rawlist($connection, '-A ' . str_replace('*', '\\*', $path));
+        $listing = $this->ftp_rawlist('-A ', str_replace('*', '\\*', $path));
 
         if (empty($listing)) {
             return false;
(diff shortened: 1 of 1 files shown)


RubtsovAV:
Removed change of Utils.php from the commit.

RubtsovAV:
I not sure how add that feature in the test suite. If you tell me, then I can try do that.

RubtsovAV:
ping

sifex:
Tested. Fixed issue that I was having regarding https://github.com/thephpleague/flysystem/issues/563 and spaces.

frankdejonge:
I'm at a conference now, but: We need tests for this, because the suite currently breaks because of this. So either it should be added in this PR or I can add it myself, but that'd be done when I have time after the conf.

barryvdh:
This breaks on my Linux ftp with dirs with spaces (old versions works there).

Do you have a test Windows FTP account?

barryvdh:
So perhaps use the systemType property to check if it's windows and needs to be escaped?

sifex:
I'm not 100% on this and I may be wrong, but this fix was for my FTPd
Server, maybe this belongs in the FTPd Adapter?

-- 
Alex Sinnott

RubtsovAV:
I think we need create test suite with a docker containers with different ftp servers. Then we will be sure in the adapter on 100%. What do you think?

RubtsovAV:
Write your ftp servers, please.

RubtsovAV:
By now it should work 100% with:
- Pure-FTPd
- vsftpd

RubtsovAV:
@sifex , @barryvdh can you test it on your envorments, pls?

P.s. and if do not complicate, tell me of your ftp server name.

sifex:
@RubtsovAV Sorry, company prohibits it. I'll see what I can do about setting up a ftpd server on my casual box today.

RubtsovAV:
Why PR is not merged?

sifex:
@RubtsovAV Needs to have tests written for it and have the tests to pass.

src/Adapter/Ftp.php
```diff
@@ -502,4 +509,31 @@ public function isConnected()
...
+
+        $response = implode("\n", $response);
+        if (stripos($response, 'Pure-FTPd')) {
+            return 'Pure-FTPd';
+        }
+
+        return null;
+    }
+
+    protected function ftp_rawlist($options, $path)
```
frankdejonge: This should be named `ftpRawlist`.

RubtsovAV:
Hello?

RubtsovAV:
@frankdejonge Done.

PR #709, opened 2016-10-14
Introduce putRequiresUpdate and putStreamRequiresUpdate

lukewaite:
# What does this PR do?

Add `putRequiresUpdate` and `putStreamRequiresUpdate` so that `Adapters` which do not differentiate between a `write` and a `update` can skip the `has` call, which has detrimental effect on some filesystem performance for initial write.
# Why?

We have observed poor performance calling `put` when using the [`flysystem-aws-s3-v3`](https://github.com/thephpleague/flysystem-aws-s3-v3) adapter on initial upload of an object.
# What is the cause

[`Filesystem->put()` calls a `has`](https://github.com/thephpleague/flysystem/blob/master/src/Filesystem.php#L97-L99) to determine if the object should be created or updated. This triggers up to two requests to s3.
1. First, `has` calls [`S3Client->doesObjectExist`](https://github.com/aws/aws-sdk-php/blob/master/src/S3/S3ClientTrait.php#L197) which causes a `HEAD` to be triggered
2. If the HEAD returns with no object found, it runs [`doesDirectoryExist`](https://github.com/thephpleague/flysystem-aws-s3-v3/blob/master/src/AwsS3Adapter.php#L655) which checks if the "directory" exists by calling `listObjects` with a prefix up to the final `/` of the path in question.

As an s3 bucket gets busier, performance sometimes suffers. The trace below shows a request where the entire `has` function has taken 635ms to run, while the actual `write` to s3, once started, takes only 69ms. Ideally we should avoid making this request, since there is no actual need. 
#### Column Headers: Duration (ms), Duration (%), Segment, Drilldown, Timestamp

image

Update: I've added a second trace where the second part of `has`, the call to `doesDirectoryExist`, has taken 1960ms to run.
image

The s3 adapter in particular does not make any distinction between a [`write()` and `update()`](https://github.com/thephpleague/flysystem-aws-s3-v3/blob/master/src/AwsS3Adapter.php#L115-L132) or [`writeStream()` and `updateStream()`](https://github.com/thephpleague/flysystem-aws-s3-v3/blob/master/src/AwsS3Adapter.php#L362-L379) so for this adapter there is no reason to call `has`. Just call `write()` or `writeStream()`.
# Other impacts

Removing the un-needed `HEAD` call on new object uploads will prevent the following issue from being needlessly introduced.

> Amazon S3 provides read-after-write consistency for PUTS of new objects in your S3 bucket in all regions with one caveat. The caveat is that if you make a HEAD or GET request to the key name (to find if the object exists) before creating the object, Amazon S3 provides eventual consistency for read-after-write.
> (from http://docs.aws.amazon.com/AmazonS3/latest/dev/Introduction.html#ConsistencyModel)
# Finally

I'm happy to work on the implementation here if changes are requested.

### DIFF
--- a/spec/FilesystemSpec.php
+++ b/spec/FilesystemSpec.php
@@ -115,6 +115,7 @@ class FilesystemSpec extends ObjectBehavior
 
     public function it_should_write_when_putting_a_new_file()
     {
+        $this->adapter->putRequiresUpdate()->willReturn(true);
         $this->adapter->has('file')->willReturn(false);
         $this->adapter->write('file', 'contents', Argument::type('League\Flysystem\Config'))->willReturn($cache = [
             'path' => 'file',
@@ -126,6 +127,7 @@ class FilesystemSpec extends ObjectBehavior
     public function it_should_write_when_putting_a_new_file_using_stream()
     {
         $stream = tmpfile();
+        $this->adapter->putStreamRequiresUpdate()->willReturn(true);
         $this->adapter->has('file')->willReturn(false);
         $this->adapter->writeStream('file', $stream, Argument::type('League\Flysystem\Config'))->willReturn($cache = [
             'path' => 'file',
@@ -136,6 +138,7 @@ class FilesystemSpec extends ObjectBehavior
 
     public function it_should_update_when_putting_a_new_file()
     {
+        $this->adapter->putRequiresUpdate()->willReturn(true);
         $this->adapter->has('file')->willReturn(true);
         $this->adapter->update('file', 'contents', Argument::type('League\Flysystem\Config'))->willReturn($cache = [
             'path' => 'file',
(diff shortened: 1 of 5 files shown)


twistor:
You can disable the `has()` calls by adding `disable_asserts` when constructing the Filesystem.

lukewaite:
@twistor Unfortunately that does not work in this case. Neither of the `assert...` methods which respect the `disable_assert` settings are called during `Filesystem::put`:
https://github.com/thephpleague/flysystem/blob/master/src/Filesystem.php#L117

Would an acceptable alternative solution to the one implemented in this PR be to modify `put` and `putStream` to respect `disable_asserts`?

lukewaite:
Updated the initial MR comments with why `disable_asserts` currently doesn't resolve the issue.

Would appreciate it if we could get eyes on this one again.

frankdejonge:
@lukewaite flysystem 1.0.40 and flysystem-aws-s3-v3 1.0.15 are now tagged which targets this issue without any configuration needed. A simple composer update of the two should resolve your performance issues.

lukewaite:
Thanks @frankdejonge!

frankdejonge:
@lukewaite welcome 👍

PR #734, opened 2016-11-25
Make ContentListingFormatter case insensitive

ramondelafuente:
Based on the discussions about the DropBox adapter not being able to traverse directories due to inconsistent behaviour regarding case, this is the actual change for a case-insensitive ContentListingFormatter.

This probably breaks all kinds of shit (except tests, those all pass just fine).

### DIFF
--- a/src/Util/ContentListingFormatter.php
+++ b/src/Util/ContentListingFormatter.php
@@ -24,7 +24,7 @@ class ContentListingFormatter
      */
     public function __construct($directory, $recursive)
     {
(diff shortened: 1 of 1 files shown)


ramondelafuente:
I see I have created a (partial) duplicate of #701

frankdejonge:
@ramondelafuente did you try out the direction of removing this filtering entirely?

twistor:
This should be dealt with directly in the Dropbox adapter.

PR #738, opened 2016-12-20
Update API docs

viraj-khatavkar:
1. Indicate that `deleteDir` deletes directory recursively
2. Add plugin first to use `listPaths()`


frankdejonge:
Thank you!

Issue #739, opened 2017-01-02
Local Adapter can't create directories when absolute path is given


ArminVieweg:
Because `$this->applyPathPrefix($dirname);` appends double backslash. 

Example:
```
C:\path\to\file => \\C:\path\to\file
```
Which is not a valid path and results with warning:

> PHP Warning:  mkdir(): Invalid argument in C:\path\to\vendor\league\flysystem\src\Adapter\Local.php on line 361

**Suggestion:** I would just don't apply path prefix on windows. For me this works.

Thanks for merging my PR, I'll attach.
~Armin

twistor:
Absolute paths shouldn't be used in Flysystem. All paths are relative to the configured adapter root. The Local adapter prefix can be absolute, but any calls into it should be relative to the prefix.

garygreen:
@twistor why? Absolute paths make perfect sense for local adapter I don't see why Flysystem needs to force you to do everything relative to a particular root path.

frankdejonge:
@garygreen this is because it breaks the encapsulation guarantee it provides. It's easy to work around it by specifying `/` as the root. But breaking out of the root is not permitted by design.

garygreen:
@frankdejonge Sometimes you don't always want to operate within root and as you say, you can get around it by specifying `/` as root but that doesn't translate to Windows very well. If you are use to interacting with absolute paths, which most people will be prior to using Flysystem, it's much easier to interact with the library by giving it a full path (particularly for the Local adapter) rather than having to strip out the path to where Flysystem wants the root to be.

To me, using everything relative to the configuration root file leads to uncertainty, I would prefer to be explicit in specifying paths but Flysystem by design is forcing you to be bound to be within a certain box.

ArminVieweg:
Also on Windows you have no dedicated root, which embraces all drive letters. So how to copy a file from C:\ to D:\ using Flysystem (without creating a workarround symlink)

twistor:
@garygreen 

> For example, if you find later on in your app you need access to one level above your root

That's an application level design. It's a security risk to allow arbitrary path traversal.

@ArminVieweg This is unfortunate I suppose. You can create a `MountManager` and move files between filesystems that way.

garygreen:
> It's a security risk to allow arbitrary path traversal.

[...]

PR #740, opened 2017-01-02
[BUGFIX] Do not apply pathPrefix in Local adapter on Windows

ArminVieweg:
Resolves: #739


frankdejonge:
I'll investigate this. Just removing the feature for windows, thus altering behaviour for a platform, is simple not a solution.

PR #743, opened 2017-01-09
Fix README.md

CthulhuDen:
PHP logical negation operator has higher precedence then comparison so old expression was equivalent to `(! substr($class, 0, 17)) === $prefix`.


frankdejonge:
🥇

PR #746, opened 2017-01-13
[Plugin] PassiveWrite: Write a file or pass if it exists without overwriting or failing

aalaap:
This adds a new plugin, PassiveWrite, which attempts to write a file, but still returns a success message, even if the target path already exists.

It is a drop-in replacement for the standard `write` method.

### DIFF
--- a/src/Plugin/PassiveWrite.php
+++ b/src/Plugin/PassiveWrite.php
@@ -1,0 +1,37 @@
+<?php
+
+namespace League\Flysystem\Plugin;
+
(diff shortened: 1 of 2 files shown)


CarsonF:
Have you seen the `update` and `put` methods? Seems like exactly what you are looking for.

twistor:
Thanks for contributing!

While this is not the same functionality as `update` or `put`, I think it's too simplistic to add as a plugin. It's simply swapping a try/catch for a return value..

PR #752, opened 2017-01-25
Simplify prefix handling.

twistor:
Relating to #739 and #740.

I was doing some investigating, and I couldn't find a problem.

I wonder @ArminVieweg, are you using the  Local adapter directly?

Anyway, while I was digging in, I added some tests, and found some ways to simplify the logic surrounding path prefixes.

### DIFF
--- a/src/Adapter/AbstractAdapter.php
+++ b/src/Adapter/AbstractAdapter.php
@@ -25,13 +25,14 @@ abstract class AbstractAdapter implements AdapterInterface
      */
     public function setPathPrefix($prefix)
     {
-        $is_empty = empty($prefix);
+        $prefix = (string) $prefix;
 
-        if ( ! $is_empty) {
-            $prefix = rtrim($prefix, '\\/') . $this->pathSeparator;
+        if ($prefix === '') {
+            $this->pathPrefix = null;
+            return;
         }
 
-        $this->pathPrefix = $is_empty ? null : $prefix;
+        $this->pathPrefix = rtrim($prefix, '\\/') . $this->pathSeparator;
     }
 
     /**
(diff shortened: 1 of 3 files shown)


src/Adapter/AbstractAdapter.php
```diff
@@ -53,17 +54,7 @@ public function getPathPrefix()
      */
     public function applyPathPrefix($path)
     {
-        $path = ltrim($path, '\\/');
-
-        if (strlen($path) === 0) {
-            return $this->getPathPrefix() ?: '';
-        }
```
twistor: This fixes a bug where the prefix is '0'.

src/Adapter/AbstractAdapter.php
```diff
@@ -53,17 +54,7 @@ public function getPathPrefix()
...
-
-        if ($prefix = $this->getPathPrefix()) {
-            $path = $prefix . $path;
-        }
```
twistor: This fixes a bug where the prefix is '0'.

src/Adapter/AbstractAdapter.php
```diff
@@ -25,13 +25,14 @@
      */
     public function setPathPrefix($prefix)
     {
-        $is_empty = empty($prefix);
+        $prefix = (string) $prefix;
 
-        if ( ! $is_empty) {
-            $prefix = rtrim($prefix, '\\/') . $this->pathSeparator;
+        if ($prefix === '') {
+            $this->pathPrefix = null;
```
twistor: This fixes a bug where the prefix is '0'.

ArminVieweg:
Hey @twistor. I am using the LocalAdapter like this (not directly):
https://bitbucket.org/InstituteWeb/deployer-scripts/src/fa3911ab8f5b324c7a855285c671979e8477ec54/src/functions.php?at=master&fileviewer=file-view-default#functions.php-100

While pasting the link, I've realized that I should store the created Filesystem like in method above... anyway. I'll try your fix. Thanks!

PR #753, opened 2017-01-29
Array To String Conversion throw when ErrorException

lukesnowden:
Multiple instances of failed composer installs when Finfo::buffer
throws ErrorException resulting in logs of Array to String Conversion.
Simple catch to return null

### DIFF
--- a/src/Util/MimeType.php
+++ b/src/Util/MimeType.php
@@ -3,6 +3,7 @@
 namespace League\Flysystem\Util;
 
 use Finfo;
+use ErrorException;
 
 /**
  * @internal
@@ -21,11 +22,13 @@ class MimeType
         if ( ! class_exists('Finfo') || ! is_string($content)) {
             return;
         }
-
-        $finfo = new Finfo(FILEINFO_MIME_TYPE);
-        $mimeType = $finfo->buffer($content);
-
-        return $mimeType ?: null;
+        try {
+            $finfo = new Finfo(FILEINFO_MIME_TYPE);
+            $mimeType = $finfo->buffer($content);
+            return $mimeType ?: null;
+        } catch( ErrorException $e ) {
+            return null;
+        }
     }
 
     /**


frankdejonge:
@lukesnowden apart from the issue itself, how do composer installs relate here?

lukesnowden:
The League has Laravel packages that that use flysystem, composer is used by Laravel for dependencies. This error screws up the install/update when publishing files. Regards.

lukesnowden:
If it helps to test this out for yourself;

`https://raw.githubusercontent.com/RubaXa/Sortable/master/README.md`

ErrorLog: 

```
#0 [internal function]: Illuminate\Foundation\Bootstrap\HandleExceptions->handleError(8, 'Array to string...', '/Users/lukesnow...', 26, Array)
#1 /Users/lukesnowden/git/luke-snowden/luke-snowden-back/vendor/league/flysystem/src/Util/MimeType.php(26): finfo->buffer('# Sortable\nSort...')
#2 /Users/lukesnowden/git/luke-snowden/luke-snowden-back/vendor/league/flysystem/src/Util.php(163): League\Flysystem\Util\MimeType::detectByContent('# Sortable\nSort...')
#3 /Users/lukesnowden/git/luke-snowden/luke-snowden-back/vendor/league/flysystem/src/Adapter/Local.php(195): League\Flysystem\Util::guessMimeType('README.md', '# Sortable\nSort...')
#4 /Users/lukesnowden/git/luke-snowden/luke-snowden-back/vendor/league/flysystem/src/Filesystem.php(98): League\Flysystem\Adapter\Local->update('README.md', '# Sortable\nSort...', Object(League\Flysystem\Config))
#5 [internal function]: League\Flysystem\Filesystem->put('README.md', '# Sortable\nSort...')
```

frankdejonge:
@lukesnowden Looking at that stack trace, it seems somebody is trying to use put with an array as the second argument. Which is not correct.

lukesnowden:
The stack trace seems to show the exception being thrown after `finfo->buffer('# Sortable\nSort...')` is being called. For that to be true, line 22 would not be having any effect and `$content` was not a string (which it is).

ref: https://github.com/thephpleague/flysystem/pull/753/files#diff-be6b6e0a22689b3417ebd77fd1f7e967L21

frankdejonge:
Ahhh I see now, my stack interpretation foo is getting mixed up (doing other typed langs recently).

lukesnowden:
👍

lukesnowden:
can you advise on when a new release is scheduled?

frankdejonge:
Somewhere tonight Amsterdam time.

lukesnowden:
thank you

twistor:
For posterity:

https://bugs.php.net/bug.php?id=71434

I wonder if this solution is better than silencing with `@`.

PR #757, opened 2017-02-09
Use strict comparison because folder "01" is not equal "1" for example.

brainless81:
Right now, you cannot create a folder named "01" if already a folder named "1" exists because the check for existing directories is not strict enough.

### DIFF
--- a/src/Adapter/Ftp.php
+++ b/src/Adapter/Ftp.php
@@ -329,7 +329,7 @@ class Ftp extends AbstractFtpAdapter
             }
         }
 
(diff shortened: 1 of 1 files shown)


frankdejonge:
@brainless81 released as 1.0.35

brainless81:
@frankdejonge not totally sure about what happens if you do pass an number as folder name. Maybe it's even better to cast to string before running "in_array()". What do you think?

frankdejonge:
@brainless81 that could be another improvement, indeed 👍

PR #758, opened 2017-02-09
Ensure that a directory name passed as integer ie. 123 is also being …

brainless81:
…detected as existing directory.


twistor:
Since `createActualDirectory()` is called from `createDir` when know that `$directory` is always a string.

PR #777, opened 2017-04-19
Suggest a Dropbox adapter that uses v2 of the Dropbox API

freekmurze:
The current Dropbox adapter uses v1 of their API. That version has been deprecated and will be switched off on Juni 2, 2017.

More info on that here: https://blogs.dropbox.com/developers/2016/06/api-v1-deprecated/

This PR replaces the old adapter with a newer one that is built on v2 of the Dropbox API.

### DIFF
--- a/adapter/dropbox.md
+++ b/adapter/dropbox.md
@@ -9,22 +9,21 @@ title: Dropbox Adapter
 ## Installation
 
 ~~~ bash
-composer require league/flysystem-dropbox
+composer require spatie/flysystem-dropbox
 ~~~
 
 ## Usage
 
-Visit [https://www.dropbox.com/developers/apps](https://www.dropbox.com/developers/apps) and get your "*App secret*".
-
-You can also generate OAuth access token for testing using the Dropbox App Console without going through the authorization flow.
+A token can be generated in the [App Console](https://www.dropbox.com/developers/apps) for any Dropbox API app. You'll find more info at [the Dropbox Developer Blog](https://blogs.dropbox.com/developers/2014/05/generate-an-access-token-for-your-own-account/).
 
 ~~~ php
-use League\Flysystem\Dropbox\DropboxAdapter;
 use League\Flysystem\Filesystem;
-use Dropbox\Client;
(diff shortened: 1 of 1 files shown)


frankdejonge:
@freekmurze lekker bezig! Thanks for the contribution.

GrahamCampbell:
Surely it's better to update the package on the league?

frankdejonge:
@GrahamCampbell I don't mind the community stepping in here, it distributes workload across multiple teams of maintainers, which is good.

frankdejonge:
That said, if @freekmurze is open to publishing/maintaining it under the league's umbrella that's also be possible.

barryvdh:
If the updates needed for v2 are relatively small, it might be easier to update the v1 package, as it might be confusing for others.
Otherwise probably best to mark the old one as deprecated.

barryvdh:
And I agree with @frankdejonge that sometimes it's just as easy to use community packages, especially when there is no time to maintain it 'officially'

GrahamCampbell:
> @GrahamCampbell I don't mind the community stepping in here, it distributes workload across multiple teams of maintainers, which is good.

I agree, but I don't agree with the way in which this is happening. We should encourage the community to contribute together, rather than to separate into their own packages. I think the league package should be updated, and other community members should be added to that particular package, to help it along.

freekmurze:
I created the adapter under the Spatie organisation because I thought that was the preferred way to go after reading [the relevant issue](https://github.com/thephpleague/flysystem-dropbox/issues/14).

Concerning the maintenance I'm probably more motivated to maintain it under the umbrella of my own organisation.  Currently I'm not going to make any commitments to maintaining stuff for other organisations. 

Of course you may just copy over the code from our package to the league's package and then I'll abandon the spatie adapter.

frankdejonge:
I prefer it this way actually, it removed some of the burden of maintaining the adapters. The league itself is more an example of how open-source PHP work _can_ be, rather than be _the_ place for packages imho. We'll also happily help getting people setup and guide them to make the most of their package, which in your case is not needed because you're got your stuff covered.

PR #782, opened 2017-04-22
Fix the ftpRawList path escape

RubtsovAV:
The first call of the ftpRawList was not escaped, because getConnection() must be called before checking on the Pure-Ftpd Server.


frankdejonge:
Nice catch @RubtsovAV!

RubtsovAV:
Thanks :)

PR #793, opened 2017-05-21
provided a set method to change the adapter at runtime

domgraziano:
I was attempting to use this amazing library with OneupFlysystemBundle for a symfony app [see this](https://github.com/1up-lab/OneupFlysystemBundle/issues/143).
A setter injection can be very helpful in a context when you want the adapter to be chosen by a final user rather than set by the application.

### DIFF
--- a/src/Filesystem.php
+++ b/src/Filesystem.php
@@ -47,6 +47,18 @@ class Filesystem implements FilesystemInterface
         return $this->adapter;
     }
 
+
+    /**
+     * Set the Adapter
+     *
+     * @param AdapterInterface $adapter
+     * @return AdapterInterface
+     */
+    public function setAdapter(AdapterInterface $adapter)
+    {
+        return $this->adapter = $adapter;
+    }
+
     /**
      * @inheritdoc
      */


twistor:
I'm unconvinced. 

It seems like you would want to register all of the available filesystems to Symfony. Then, in the application choose the appropriate one.

domgraziano:
@twistor, thanks for reviewing this. Generally you are right, if I knew all the storage details in advance and is what I would do but what if, like in my particular case, I want to provide a Dropbox storage system based on user dropbox credential (after they've logged in not through dropbox) and to do that I need an access token aka every user get the access token and create the storage system at runtime whit their own access token? In this particular case I do need to create it at runtime, and a setter method would be the only solution as far as I can see. The setter method would be helpful to give this level of flexibility every time you don't know full access credential in advance, not only in my particular case.

frankdejonge:
Seems to me like you need a filesystem factor rather than replacing the inner adapter. If you cab creatie the client, you could just as wel wrap it in Flysystem?

Op vr jul 14 2017, om 15:35 schreef domgraziano:
> @twistor[1], thanks for reviewing this. Generally you are right, if I knew all the storage details in advance and is what I would do but what if, like in my particular case, I want to provide a Dropbox storage system based on user dropbox credential (after they've logged in not through dropbox) and to do that I need an access token aka every user get the access token and create the storage system at runtime whit their own access token? In this particular case I do need to create it at runtime, and a setter method would be the only solution as far as I can see. The setter method would be helpful to give this level of flexibility every time you don't know full access credential in advance, not only in my particular case.> —
> You are receiving this because you are subscribed to this thread.
> Reply to this email directly, view it on GitHub[2], or mute the thread[3]. > 


Links:

  1. https://github.com/twistor
  2. https://github.com/thephpleague/flysystem/pull/793#issuecomment-315360878

domgraziano:
@frankdejonge thanks frank. Do you mean a factory ? Well, yes this is another option but it would be the equivalent of unplug this bundles https://github.com/1up-lab/OneupFlysystemBundle / https://github.com/dustin10/VichUploaderBundle from my project since they provide this sort of factory but just when symfony is initialised ... that's why I thought "well if the library would have a setter method the constructed object already relying on those extra bundles could be changed at runtime" if you know what I mean (agree it's maybe less elegant than implementing a custom factory, but I think a setter method would make the library more flexible then leave the choice of client side design to the client) ... feel free to reject this if you believe is not the case to include a setter :) not gonna take it personal

twistor:
I'm going to close this. The answer is that you need a runtime factory for your filesystems. It's not very much code, and assumptions are already made that the `Filesystem` wraps a specific adapter. If you can create the adapter during runtime, it's trivial to wrap it in a `Filesystem` object.

PR #803, opened 2017-06-21
Implement serialize prevention.

twistor:
Coming from https://github.com/thephpleague/flysystem-sftp/issues/36.

I've moved the actual data storage into a function static, that should help to ensure that it doesn't get serialized. I'm also not sure if class statics leak.

Edit: Added some tests to ensure behavior.

### DIFF
--- a/src/SafeStorage.php
+++ b/src/SafeStorage.php
@@ -2,38 +2,75 @@
 
 namespace League\Flysystem;
 
+/**
+ * Prevents data from being leaked via exceptions and serialization.
+ */
 final class SafeStorage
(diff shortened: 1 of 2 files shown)


twistor:
I would argue that `League\Flysystem\Filesystem` should implement `__sleep()`, but that would be unfortunate for people successfully serializing current filesystems.

frankdejonge:
@twistor I'm thinking of closing this one. While it protects some, I think this should be the user's responsibility. Agree?

twistor:
That's fine with me.

PR #804, opened 2017-06-26
Allow ContentListingFormatter inheritance

chosroes:
It should be possible to override this class.

### DIFF
--- a/src/Util/ContentListingFormatter.php
+++ b/src/Util/ContentListingFormatter.php
@@ -4,19 +4,17 @@ namespace League\Flysystem\Util;
 
 use League\Flysystem\Util;
 
-/**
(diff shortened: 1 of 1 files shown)


twistor:
Can you explain why?

frankdejonge:
This is marked as internal and will remain to be that way because it's just an extracted part that isn't part of the public API. If there's a problem somewhere that can be fixed by allowing this to be extendable I'd rather hear what the problem is so we can fix that. When you want to use this class outside the context of Flysystem I suggest copying the code, because when you couple with this and something internally changes your code will fail. This is something we want to prevent. Also, the code inside this class is not super complicated, nor is it a large class. So copying it is not a big issue.

PR #810, opened 2017-07-05
Fixed #809: FTP Server throws error while using has() method

### LINKED ISSUE #809: FTP Server throws error while using has() method
Since I'm using the "has()" method very often in my application I'm realizing that the has() method is falling back to an getMetadata() in the FTP implementation. This one is using a

`if (@ftp_chdir($connection, $path) === true) {`

to early detect that the given $path is an directory before further checking file related things. By doing so, the FTP Server may/will throw an error while the client ignores the possible "error"-feedback once it's a folder.

bildschirmfoto 2017-07-05 um 11 03 03

Nothing critical with our FTP Server is further happening, besides an endless error logfile which is quiet annoying. Since I read in #421 that this may cause other servers to strike, I would like to open this issue to improve the code.

It hopefully is possible to optimize the code so that no server side error is triggered while providing a similar performance and stability with various FTP server implementations.

RubtsovAV:
I tried a lot of variants and the only this consistently works on Pure-ftpd and vsftpd.

P.S. I don't know what about throws errors on server side. I just tried implement the `has()` method without `chdir`

### DIFF
--- a/src/Adapter/Ftp.php
+++ b/src/Adapter/Ftp.php
@@ -386,27 +386,20 @@ class Ftp extends AbstractFtpAdapter
             return ['type' => 'dir', 'path' => ''];
         }
 
-        if (@ftp_chdir($connection, $path) === true) {
-            $this->setConnectionRoot();
+        $dirname = strpos($path, '/') ? dirname($path) : '';
 
-            return ['type' => 'dir', 'path' => $path];
-        }
-
-        $listing = $this->ftpRawlist('-A', str_replace('*', '\\*', $path));
-
-        if (empty($listing) || in_array('total 0', $listing, true)) {
+        $listing = $this->listDirectoryContents($dirname, false);
+        if (empty($listing)) {
             return false;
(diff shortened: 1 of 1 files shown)


Gummibeer:
This solves my problem - thanks. My ftp Server (synology diskstation) answered with
> ftpd: 1.txt: No such file or directory.
I first added a simple second `preg_match` but this solution looks more solid and reusable.

Is there any reason why it's closed/rejected and not merged?

RubtsovAV:
I rejected PR, because the issue was closed.

If you want, I can open it again. But you will need to fix the unit tests and I do not have free time for this now.

Gummibeer:
@RubtsovAV right, this changes affects the following methods a lot.
How would it be to just add a `try/catch`!?

```php
try {
    return $this->normalizeObject($listing[0], '');
} catch(\RuntimeException $ex) {
    return false;
}
```

with this I don't get problems with any method.

PR #813, opened 2017-07-10
Rename when moving file within the same file system

Stadly:
When moving a file using the MountManager, if from and to are in the same file system, perform a rename instead of copying and deleting the file.

### DIFF
--- a/src/MountManager.php
+++ b/src/MountManager.php
@@ -255,7 +255,7 @@ class MountManager
     {
         list($prefixFrom, $pathFrom) = $this->getPrefixAndPath($from);
         list($prefixTo, $pathTo) = $this->getPrefixAndPath($to);
-        
+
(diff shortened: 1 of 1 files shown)


src/MountManager.php
```diff
@@ -253,10 +253,24 @@ public function listWith(array $keys = [], $directory = '', $recursive = false)
...
+        if ($prefixFrom === $prefixTo) {
+            $filesystem = $this->getFilesystem($prefixFrom);
+            $renamed = $filesystem->rename($pathFrom, $pathTo);
+
+            if ($renamed && isset($config['visibility'])) {
+                return $filesystem->setVisibility($pathTo, $config['visibility']);
+            }
+            
+            return $renamed;
+        } else {
```
twistor: We can get rid of this `else`.

PR #818, opened 2017-07-22
Delete all files first so directories are empty in FTP Adapter.

### LINKED ISSUE #817: Recursive Directory Delete Support for FTP Adapter
Currently, the Ftp adapter doesn't support recursive removal of a directory.

```
    public function deleteDir($dirname)
    {
        $connection = $this->getConnection();
        $contents = array_reverse($this->listDirectoryContents($dirname));

        foreach ($contents as $object) {
            if ($object['type'] === 'file') {
                if ( ! ftp_delete($connection, $object['path'])) {
                    return false;
                }
            } elseif ( ! ftp_rmdir($connection, $object['path'])) {
                return false;
            }
        }

        return ftp_rmdir($connection, $dirname);
    }
```

This doesn't make sense if you really intend to delete the directory and all its content. I understand the limitation of `ftp_rmdir` function but it would be great to have such feature.

twistor:
This should fix #817. We should add a test but it won't be fun.

### DIFF
--- a/src/Adapter/Ftp.php
+++ b/src/Adapter/Ftp.php
@@ -313,12 +313,20 @@ class Ftp extends AbstractFtpAdapter
         $connection = $this->getConnection();
         $contents = array_reverse($this->listDirectoryContents($dirname));
 
+        foreach ($contents as $key => $object) {
+            if ($object['type'] !== 'file') {
+                continue;
+            }
+
+            unset($contents[$key]);
+
+            if ( ! ftp_delete($connection, $object['path'])) {
+                return false;
+            }
+        }
+
         foreach ($contents as $object) {
-            if ($object['type'] === 'file') {
-                if ( ! ftp_delete($connection, $object['path'])) {
-                    return false;
-                }
-            } elseif ( ! ftp_rmdir($connection, $object['path'])) {
+            if ( ! ftp_rmdir($connection, $object['path'])) {
                 return false;
             }
         }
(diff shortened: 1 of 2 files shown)


frankdejonge:
Couldn't it just be: 

```php
    public function deleteDir($dirname)
    {
        $connection = $this->getConnection();
        $contents = array_reverse($this->listDirectoryContents($dirname));

        foreach ($contents as $object) {
            if ($object['type'] === 'file') {
                if ( ! ftp_delete($connection, $object['path'])) {
                    return false;
                }
            } elseif ( ! $this->deleteDir($object['path'])) {
                return false;
            }
        }

        return ftp_rmdir($connection, $dirname);
    }
```

twistor:
@frankdejonge, I was worried about memory usage and multiple list calls with that approach. Also, recursing arbitrarily deep can be problematic.

twistor:
It could work, if we don't list recursively. That should keep memory reasonable. It would still be slower, since we can get the listing in one go without recursion.

twistor:
In summary, it's worse big-O wise, but better performance wise. It also changes the behavior less than a recursive approach.

I would think that a recursive approach would be considered if listing the whole directory had been a problem, but since it hasn't, I'm not sure what the problem is.

frankdejonge:
@twistor could we test this using docker and some ftp server? It looks good and I think your reasoning is solid 👍

vsdm:
When will you accept this PR?

PR #820, opened 2017-07-26
Fix more empty() string calls.

twistor:


frankdejonge:
@twistor should we issue a release for this?

PR #821, opened 2017-07-29
Adding PSR-11 factories to list of Integrations

wshafer:


wshafer:
Would love to get these PSR-11 factories in the list.

Issue #822, opened 2017-07-31
Local Adapter ignore permission/file/public property


DarkSide666:
Adapter_Local ignores custom permission/file/public if it is set. When working with files it doesn't use permissionMap property.

For example,
```
new Filesystem(new LocalAdapter(
    'folder',
    LOCK_EX,
    LocalAdapter::DISALLOW_LINKS,
    ['file'=>['public'=>0664],'dir'=>['public'=>0775]]
))
```
This filesystem will create directories with 0775 permissions, but files still with default 06**4**4.

PR #824, opened 2017-08-02
Set permissions on write() in Local adapter.

twistor:
As mentioned in #822, we need to always set permissions on write in Local adapter.


frankdejonge:
I'd be ok with this if we can make it so that a new default_visibility option of some sort was introduces because this modifies behaviour for existing applications. For example, the default might have a different default visibility  (which is why we need to set it) but the application might have relied on that not happening.

PR #825, opened 2017-08-06
getTimestamp() in Filesystem

isaackwan:
Hello,

I am proposing to check for existence of the `timestamp` key in the returned file info array instead of the key `mimetype`, which apparently was a result of copy-pasting the implementation for `getMimetype()`.

If this was intentional -- sorry!

Thanks


frankdejonge:
@isaackwan good catch!

frankdejonge:
@isaackwan it's tagged as 1.0.41

PR #842, opened 2017-09-28
Recursive Ftp folder remove script

jcgaudette:
You cannot use ftp_rmdir() if the directory is not empty. So if you want to delete a folder, that contains one or more sub-folder containing some files, the script return false and it cause an error. That way, you can completely remove a folder and all the child at once, which is, I think, a better way.


jcgaudette:
Just saw that there's a pull request discussing about that. I'll close mine and vote +1 for the feature.

PR #843, opened 2017-09-29
remove check for empty path value to preserve proceeding slash

rferons:
the function `normalizeRelativePath` removes the proceeding slash from absolute paths. 
By removing the check for empty `$part` you can preserve the beginning slash.

### DIFF
--- a/src/Util.php
+++ b/src/Util.php
@@ -108,7 +108,6 @@ class Util
 
         foreach (explode('/', $path) as $part) {
             switch ($part) {
-                case '':
                 case '.':
                 break;
 


frankdejonge:
Flysystem works with paths relative to a root. This is the single most important thing Flysystem does to ensure portability and cross-fs behaviour. If you want to access files relative to the root, simply specify `/` as the root. For a given filesystem/adapter combination one should always remain in that root, breaking out of it is a security violation.

rferons:
Thanks for the information. After reviewing how we were initializing the local adapter the issue was resolved by setting the root to the application root and not the fs root.

frankdejonge:
👍

frankdejonge:
Also, @rferons thanks for trying to contribute, I saw this was first contribution on github. This one may not have worked out in a merge, but rock on 👊

PR #846, opened 2017-10-05
Ability to append file contents

Lctrs:
Supersed #696.

Since we have no news from @chris1312 in the parent PR, I took the liberty to fork his work, fix the remarks made by the reviewers and creating this new PR.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,10 +1,3 @@
-._*
-.~lock.*
-.buildpath
-.DS_Store
-.idea
-.project
-.settings
 vendor
 bin
 coverage
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,10 @@
+._*
+.~lock.*
+.buildpath
+.DS_Store
+.idea
+.project
+.settings
 vendor
 bin
 coverage
--- a/src/Adapter/Local.php
+++ b/src/Adapter/Local.php
@@ -210,14 +210,7 @@ class Local extends AbstractAdapter implements AppendableAdapterInterface
             return false;
         }
 
-        if (!flock($stream, LOCK_EX)) {
-            throw new Exception('Could not get file lock');
-        }
-
         stream_copy_to_stream($resource, $stream);
-
-        fflush($stream);
-        flock($stream, LOCK_UN);
 
         if (!fclose($stream)) {
             return false;
--- a/src/Adapter/Local.php
+++ b/src/Adapter/Local.php
@@ -205,7 +205,7 @@ class Local extends AbstractAdapter implements AppendableAdapterInterface
     {
         $location = $this->applyPathPrefix($path);
         $this->ensureDirectory(dirname($location));
-        $stream = fopen($location, 'a');
+        $stream = fopen($location, 'ab');
 
         if (!$stream) {
             return false;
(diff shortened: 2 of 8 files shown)


cocochepeau:
Any news on this one?

Stadly:
Why was this and https://github.com/thephpleague/flysystem/pull/696 closed with no comment? I think append would be useful. Especially combined with this idea: https://github.com/thephpleague/flysystem/pull/696#issuecomment-314140800

frankdejonge:
@Stadly append is not useful for file storage. The cases to append are mostly for things like logs and other things that only should ever be local. To facilitate this means I have to support it for numerous adapters for which it makes no sense to have it in the first place and most of the adapter SDK don't support it either, making it a burden to support.

jhdxr:
@frankdejonge no. `AppendableAdapterInterface` is a independent interface which means author of adaptors can choose whether to support append or not. What's more, if some SDK can really support append, but this widely used filesystem library doesn't support append **interface**, then as an application developer, I have to bypass this abstraction, and call SDK directly, which obviously violate the purpose of using this project.

Potherca:
It isn't true that you need to call an SDK directly or that this violates the purpose of this project. You can alway create a [plugin](https://flysystem.thephpleague.com/docs/advanced/plugins/) that provides `append` functionality.

I would be more interested in knowing which SDK's are actually _used_ that provide append.

The lack of attention to this ticket combined with the amount of work it would take to support this make me in favour of @frankdejonge's point of view.

@jhdxr You may not be aware of this but the tone you are using isn't particularly friendly.

peinhu:
@Stadly 
Thanks for your efforts on this issue, I believe a lot of people have the same trouble with us.

@frankdejonge
> append is not useful for file storage.

I'm afraid I consider that's not the case, currently I'm working on a laravel package that can help people to upload large files to their hosts without change the memory limit, it will slice the big file into small chunks with HTML5 api and upload them to the server, and finally put them together in one piece.   

At the last stage when I need to combine the files, an error occured, which turned out to be a memory exhausted issue, then I found there is no real append function, the append function I use was to read the whole file into the memory, add contents to the data, then overwrite the original file. I was shocked and could not believe my eyes.  

I'm not making you guys to break BC, and  I'm not going to add a burden to you. I'm just wondering if there is a way to add some data to an existing file while using the local adapter, I suppose this is a reasonable requirement.

If supporting append is unforgivable and sinful to you, at least please give us an opportunity to add some data to a file in any other form, like allowing `$writeFlag` to be modified:
```php
 /**
     * @param $writeFlags
     * @return array
     */
    public function setWriteFlags($writeFlags)
    {
        $this->writeFlags = $writeFlags;

        return compact('writeFlags');
    }

```
Then in Laravel I can append data to a file like this:
```php
Storage::disk('local')->getAdapter()->setWriteFlags(FILE_APPEND);
Storage::disk('local')->put($filePath, $contents, "");# I know it's ugly but it works
```
  
I totally agree with what @jhdxr has said, maybe `AppendableAdapterInterface` is one way to solve this issue.

absolux:
Hi @peinhu ,

> at least please give us an opportunity to add some data to a file in any other form

You don't have to wait the maintainers to add the missing method. Do it yourself.
Extend the local adapter, add the missing method(s), include the adapter in your project.
It's an open source project, if you would like to share your code, create a PR, merged or not it doesn't matter.

tadasauciunas:
@Potherca 
> 
> I would be more interested in knowing which SDK's are actually _used_ that provide append.
> 

In a project I personally work on, we have implemented plugins for the S3 Adapter and the WebDAV Adapter to support appendable interface. They work very well and I would be willing to make pull requests to both of those mentioned adapters, if this PR wouldn't have been closed.

Issue #620, opened 2016-03-16
Introduce a common exception for failed commands


BenMorel:
This is somewhat related to #611.

If I try to write a file on S3 using Flysystem, and a connection problem occurs, I get an S3 exception:

> Aws\S3\Exception\S3Exception: Error executing "PutObject" on ...

For Dropbox, I get a (subclass of) `Dropbox\Exception`.

First of all, these adapters do not respect the `AdapterInterface` contract by allowing an exception to be thrown, instead of returning [false on failure](https://github.com/thephpleague/flysystem/blob/master/src/AdapterInterface.php#L24).

Secondly, I would suggest throwing an exception in all cases when something goes wrong on the underlying storage. And to keep code portable (i.e. allowing to change the storage adapter at the flip of a switch), this exception should belong to the Flysystem package.

At the moment, I have to do:

``` php
try {
    $filesystem->write($path, $contents);
} catch (S3Exception $e) {
    // ...
}
```

If I decide to switch to Dropbox, I have to search through my entire codebase for Flysystem calls, and update them to something like:

``` php
try {
    $filesystem->write($path, $contents);
} catch (Dropbox\Exception $e) {
    // ...
}
```

And if I'm writing an app that can accept any adapter, using information from a config file, I have to take all possible errors into account:

``` php
try {
    $filesystem->write($path, $contents);
} catch (S3Exception $e) {
    // ...
} catch (Dropbox\Exception $e) {
    // ...
}
```

And so on, without forgetting any adapter. As you can see, this is not a very good practice.

Introducing a common interface for these errors would allow us to call:

``` php
try {
    $filesystem->write($path, $contents);
} catch (AdapterException $e) {
    // ...
}
```

This way, there is no need to rewrite any code when switching to another adapter.

As I understand the project in its current status, you decided to return `false` in some methods, when something goes wrong. The fact is, the few adapters I had a chance to look at do not respect this contract properly (S3 and Dropbox thrown their own exceptions, Ftp and Local raise PHP errors).

There are two solutions to this problem:
- Fix all the adapters to catch any exception or error (including PHP warnings for adapters relying on native functions) and return `false` where appropriate
- Change the contract to never return `false` on methods such as `write()`, and throw an `AdapterException` on failure instead

I would strongly suggest the second option, because it breaks the application flow when something goes wrong; otherwise an error could go unnoticed if the developer forgets to check the return value  using something like:

``` php
if (! $fs->write($path, $contents)) {
    ...
}
```

[...]

Issue #611, opened 2016-03-04
Can't catch exceptions during testing with AWS credentials


bandtank:
One of my tests breaks AWS credentials on purpose and it has been working with regular filesystem access, but flysystem throws FileNotFound (which makes no sense because the file is present in the S3 bucket - the credentials are the problem) and I can't seem to catch it.

```
try {
    $s3->delete($path);
} catch (Exception $e) {
    return $this->respondError($e->getMessage());
}
```
I saw the following comment in a closed issue: "... It [flysystem] is meant to be used when you have valid credentials."

Why doesn't flysystem fail earlier than here in src\Filesystem.php?
```
public function assertPresent($path)
{
    if ( ! $this->has($path)) {
        throw new FileNotFoundException($path);
    }
}
```

Here's the relevant code snippet from my controller, which includes the first snippet as well:
```
$s3 = \Storage::disk('s3');
if (!$s3) {
    return $this->respondInternalError("Couldn't establish a connection to file storage");
}

try {
    $s3->delete($path);
} catch (Exception $e) {
    return $this->respondError($e->getMessage());
}
```

The Laravel app handler is catching the error instead of my controller. That seems like a separate, but related, issue because I suppose it's possible that could be the first exception handler found by PHP. I'm still trying to find out how to have the exception thrown sooner than when delete is called or if that's even the right expectation. Lastly, is there a better way to test connectivity with S3/cloud filesystems? I'm not suggesting this method is the best in any scenario.

bandtank:
Also, I should mention that the put doesn't fail the same way:

```
try {
    $s3->put($path, file_get_contents($file), 'public');
} catch (S3Exception $e) {
    return $this->respondInternalError([
        'message'           => "Trying to store the uploaded image failed",
        'AWS_Error_Code'    => $e->getAwsErrorCode()
    ]);
}
```

Fails with this error when the credentials are incorrect:
```
{
  "error": {
    "message": "Trying to store the uploaded image failed",
    "AWS_Error_Code": "InvalidAccessKeyId"
  },
  "status_code": 500
}
```

That's what I was expecting to see when I ran the delete test.

bandtank:
[...]

PR #847, opened 2017-10-06
Use League\Flysystem\Exception in Adapters, users can catch single exception

mhlavac:
First of all, excellent work on the library, thanks everyone who is contributing.

The goal of this PR is to suggest using `League\Flysystem\Exception` for all errors coming from adapters. The change is related to issue #620 and #611 

**Why?**
I need to differentiate between false and exception. Also when I call `Filesystem->has()` or different method I want to put the logic inside try/catch block and act on it.

Right now different exceptions might be returned and there is no rule that custom adapters should extend or wrap exceptions inside `League\Flysystem\Exception`. This means that if I use `Filesystem` and I want to catch Google Cloud Storage exception I have to check Google SDK exceptions.

The fix for adapter can be implemented without BC break, I can extend or wrap exceptions in already existing `League\Flysystem\Exception`. So here I suggest changing the phpdocs to include `@throws Exception`.

**Possible BC break**
I would love to keep BC breaks minimal or none if possible, otherwise this will end up in 2.0 branch, but I changed Local and FTP adapters to throw `League\Flysystem\Exception`. I am happy to remove these changes from this PR if it's too much.

**Discussion**
Let's discuss what's the best approach to deal with this. Right now I think it's should be the Adapter's job to deal with specific exceptions and throwing something Filesystem user expects, but maybe there is better solution to this.

### DIFF
--- a/src/Adapter/Ftp.php
+++ b/src/Adapter/Ftp.php
@@ -6,9 +6,9 @@ use ErrorException;
 use League\Flysystem\Adapter\Polyfill\StreamedCopyTrait;
 use League\Flysystem\AdapterInterface;
 use League\Flysystem\Config;
+use League\Flysystem\Exception;
 use League\Flysystem\Util;
 use League\Flysystem\Util\MimeType;
-use RuntimeException;
 
 class Ftp extends AbstractFtpAdapter
 {
@@ -124,6 +124,8 @@ class Ftp extends AbstractFtpAdapter
 
     /**
      * Connect to the FTP server.
+     *
+     * @throws Exception
      */
     public function connect()
     {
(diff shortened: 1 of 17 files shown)


prisis:
Pls dont change the Runtime or Logic exceptions!
Create for all used base exception a flysystem exception, like League\Flysystem\RuntimeException or League\Flysystem\LogicException.

mhlavac:
@prisis thanks for the feedback. I introduced new Flysystem exceptions RuntimeException, LogicException and ErrorException.

There is no Exception interface at the moment, but they are all extending the `League\Flysystem\Exception` same as the other exceptions do.

I also updated the tests to reflect the new behavior.

prisis:
Much better :) 
But you need to change the throws on the interface with the right exception, Exception is only the base class :)

mhlavac:
@prisis sorry, this took me a while. Thanks for feedback again.

I updated the exception descriptions. For me it's important to tell apart if file doesn't exist or if it there was a connection error. Adapters can throw more specific exceptions, by extending the generic Exception.

Should I introduce `AdapterException` for that matter, or is using `Exception` good enough?

PR #852, opened 2017-10-16
Update PHPUnit dependencies

srmklive:


GrahamCampbell:
What does this fix please?

srmklive:
Initially when I added this PR, tests were failing on PHP 5.5 & 7.1. They are now fixed, so there is no need for the PR.

PR #854, opened 2017-10-16
853 added clearstatcache to Local::getSize

jkromski:


twistor:
I think this is backwards. `clearstatcache()` should be called on writes, otherwise we're doing lots of pointless work.

frankdejonge:
@twistor very good point.

jkromski:
@twistor if we assume that only php script/process access the system then this is correct and clearing on save is better option. But if file is accessed by multiple processes e.g. one is reading other is writing then clearing on write won't help. If I'm wrong please correct me


taken from php.net: 
However, in certain cases, you may want to clear the cached information. For instance, if the same file is being checked multiple times within a single script, and that file is in danger of being removed or changed during that script's operation, you may elect to clear the status cache.

jkromski:
maybe flysystem should have method e.g clearMetadataCache which can be called on demand?

jkromski:
any thoughts?

twistor:
@jkromski What is your use-case specifically?

The stat cache is per-request, and would only be cached if you're calling `stat()` multiple times in a single request/process. This could be an issue when writing a daemon, for instance, but I think in that case, it would be better to just make a custom adapter.

frankdejonge:
I think extending is the best option here. The other option would be to merge this, which would cause problems for the general usecase (so a no-go) or adding an option for this and checking a boolean of some sort (which is not ideal because then we might get an overload of those).

jkromski:
@twistor
I did small test:
`$file = 'test.txt';
touch($file);
for ($i = 0; $i < 10; $i++)
{
    clearstatcache(false, $file);
    $before = new SplFileInfo($file);
    exec('echo 1 >> test.txt'); // this represent some third party process in this case bash
    $after = new SplFileInfo($file);
    $sizes = [$before->getSize(), $after->getSize()];
    echo $sizes[0] .' = '. $sizes[1]
        .' :: '
        . ($sizes[0] == $sizes[1] ? 'same still': 'correct')
        . PHP_EOL;
    sleep(1);
}`

@frankdejonge 
from my point of vew 'option' for local adapter is reasonable solution, but this is your call. Let me know what would u like to implement I'll be happy to follow

PR #864, opened 2017-11-22
Fix setRoot on AbstractFtpAdapter

The-Hasanov:
i use laravel filesystem but have problem if  i  change root folder first ftp use folder not changed
before fix:
        $storage->createDir('root');
        $storage->getDriver()->getAdapter()->setRoot($storage->getDriver()->getAdapter()->getRoot() . '/root');
        $storage->createDir('products');
folders:
  /root
  /products

after:
folders:
  /root
  /root/products

### DIFF
--- a/src/Adapter/AbstractFtpAdapter.php
+++ b/src/Adapter/AbstractFtpAdapter.php
@@ -212,7 +212,11 @@ abstract class AbstractFtpAdapter extends AbstractAdapter
     public function setRoot($root)
     {
         $this->root = rtrim($root, '\\/') . $this->separator;
-
+        
+        if (method_exists($this, 'setConnectionRoot')) {
+            $this->setConnectionRoot();
+        }
+        
         return $this;
     }
 


frankdejonge:
The root is not supposed to be changed at runtime, you should create a new Filesystem for the next call.

The-Hasanov:
but if i use 
$storage->exists('something');
before create products dir 
root folder automatically changed

The-Hasanov:
after changed root  try create 8 folder only first folder not true path

frankdejonge:
Correct, because if needs to change into directories to check if they exist and then reset the connection root which you're altered in the meanwhile. But as I said before, you need to create a new filesystem or just use the same filesystem without corrupting the root path. Just because something happens after you use it incorrectly that doesn't mean it's intended usage nor that I will support the use-case. I've now given you two alternatives, stated it's not intended usage and explained why your unintended usage had "worked" for you. Besides all that the ecosystem around flysystem doesn't support changing of root directories, so I'm closing this PR.

PR #867, opened 2017-11-24
FTP: Fix manual recursion mode

valga:
FTP listings must be normalized by prepending the base directory to all paths.

Automatic recursion mode already does that (note the `$directory` parameter): https://github.com/thephpleague/flysystem/blob/3b7394d78ad5936dab8a7041c63ac83722af27d9/src/Adapter/Ftp.php#L500

### DIFF
--- a/src/Adapter/Ftp.php
+++ b/src/Adapter/Ftp.php
@@ -507,7 +507,7 @@ class Ftp extends AbstractFtpAdapter
      */
     protected function listDirectoryContentsRecursive($directory)
     {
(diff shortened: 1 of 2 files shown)


valga:
The bug was introduced in  https://github.com/thephpleague/flysystem/commit/58a469abdeab67e7c112c288e34b425f9bbb5c1b#diff-444155fd1c8c44e1eecbb35e7a9cdd17L471

alex-sainer:
would be nice if this could be merged soon...

frankdejonge:
This is merged, I've added tests for it, and tagged as 1.0.42.

PR #871, opened 2017-11-29
Update Filesystem.php method createDir with the option to return the object

Braunson:
Added parameter to return the object from `createDir`, otherwise defaults to current behavior of returning bool.

Use Case: I wold be nice to get the response from the adapter, in my case https://github.com/nao-pon/flysystem-google-drive returns an object/array of the path in Google Drive which I need.

### DIFF
--- a/src/Filesystem.php
+++ b/src/Filesystem.php
@@ -254,12 +254,16 @@ class Filesystem implements FilesystemInterface
     /**
      * @inheritdoc
      */
-    public function createDir($dirname, array $config = [])
+    public function createDir($dirname, array $config = [], $return_object = false)
     {
         $dirname = Util::normalizePath($dirname);
         $config = $this->prepareConfig($config);
-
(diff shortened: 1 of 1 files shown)


frankdejonge:
This is not part of the public interface. If you want the response you should call it directly and not solve that in the Filesystem class.

Braunson:
Wow that was a quick

frankdejonge:
@Braunson I'm in the middle of issue triage right now. Saw this come in so figured I'd handle it right away. In general the stability and consistency of the interface is important for flysystem, these kind of workaround generally pollute the interface, so I'm not in favour of them, especially since the "workaround" is easy.

frankdejonge:
I hope this explains why I'm closing the PR.

Braunson:
@frankdejonge No worries, I just wasn't expecting that quick of a response 😆 I'll call the adapters method directory. Cheers.

PR #873, opened 2017-12-03
Remove Mockery as a dependency.

twistor:
This removes all usage of Mockery. Prophecy works fine for most things. The only complication is around testing magic `__call` methods.

It also changes the `setExpectedException` calls to the modern `expectException` and adds in a hack for `createMock`.


frankdejonge:
Awesome

PR #876, opened 2017-12-18
correct read function return

hossein142001:
different return by other adapter


frankdejonge:
Hi @hossein142001, you should wrap an adapter in a `Filesystem` instance and call that, this does the return value extraction. Never use the adapter directly unless you've got a non-standard use-case.

PR #883, opened 2018-01-12
Fixed phpdoc and in FTP check root content instead of isset

mhlavac:
* $root in Ftp::setConnectionRoot is always defined, but sometimes it doesn't have value. Check $root instead of isset($root)
* Fixed many phpdoc errors.
* Correct case for finfo instead of Finfo

I found these errors thanks to phpstan static analysis tool. phpstan can be executed with following script:
``` sh
ls phpstan.phar || wget https://raw.githubusercontent.com/phpstan/phpstan-shim/0.9.1/phpstan.phar && \
chmod 777 phpstan.phar && \
./phpstan.phar analyse -a vendor/autoload.php -l 7 src
```

### DIFF
--- a/src/Adapter/AbstractAdapter.php
+++ b/src/Adapter/AbstractAdapter.php
@@ -7,7 +7,7 @@ use League\Flysystem\AdapterInterface;
 abstract class AbstractAdapter implements AdapterInterface
 {
     /**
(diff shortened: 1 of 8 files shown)


frankdejonge:
Good stuff!

PR #884, opened 2018-01-13
Use aliases in documentation

willemstuursma:
I found the documentation a bit confusing due to the same aliases used in all examples. Hope this clears it a bit up.


frankdejonge:
👍

PR #889, opened 2018-01-29
Add mimetype detection for bpmn and dmn files

steffenbrand:


frankdejonge:
@steffenbrand Hi, I tries looking up these extensions to verify but both resulted in a mime-type of `application/octet-stream`. Can you supply links to something I can use to verify this with?

steffenbrand:
I can't, because what you say is right and what I did was wrong!
I'll correct this.

PR #895, opened 2018-02-19
Added @method phpdocs on Handler class

ikeyan:
I omitted listWith because its first argument is `array $keys = []`.


twistor:
@ikeyan Can you clarify why `array $keys = []` makes it undesirable to add?

frankdejonge:
The handlers have been deprecated and will be removed in the new version. I recommend moving away from using them.

PR #900, opened 2018-02-23
Made has() and getMetadata() behavior consistent across adapters

### LINKED ISSUE #899: Inconsistent handling of directories across adapters
Directories are second class citizens in Flysystem because not all filesystems support them in the same way. However, the methods has() and getMetadata() should still return consistent results.

Local Adapter:
has() on directory that exists: true
has() on directory that doesn't exist: false
getMetadata() on directory that exists: array
getMetadata() on directory that does not exist: exception

AWS S3 Adapter:
has() on directory that exists: true
has() on directory that doesn't exist: false
getMetadata() on directory that exists: false
getMetadata() on directory that does not exist: exception

Azure Blob Adapter:
has() on directory that exists: false
has() on directory that doesn't exist: false
getMetadata() on directory that exists: exception
getMetadata() on directory that does not exist: exception

These can be made to all return consistent results with some fairly simple wrapping code that makes use of listContents() which seems to have consistent behavior across adapters.

```php
/**
 * Overrode to deal with directories consistently across adapters
 *
 * @param string $path
 *
 * @return bool
 */
public function has($path)
{
    $path = Util::normalizePath($path);

    // Root path always exists. Skip the filesystem check
    if ('' == $path || '.' == $path) {
        return true;
    }

    if ($this->getAdapter()->has($path)) {
        return true;
    }

    // Deal with inconsistent adapter behavior.
    // Some adapters always return false for directories.
    // Check if this is an existing directory by checking if it exists in its parent's contents
    $parts = explode('/', $path);
    array_pop($parts);
    $parentPath = implode('/', $parts);

    foreach ($this->listContents($parentPath) as $parentContent) {
        if ($path == $parentContent['path']) {
            return true;
        }
    }

    return false;
}

/**
 * @param string $path
 *
 * @throws FileNotFoundException if file does not exist
 * @return array
 */
public function getMetadata($path): array
{
    $path = Util::normalizePath($path);
    $this->assertPresent($path);

    // Deal with inconsistent adapter behavior.
    // Some adapters will throw exceptions for directories; Some return false; Some return an array
    try {
        $metaData = $this->getAdapter()->getMetadata($path);
    } catch (\Exception $e) {
        // Do nothing
    }

    // If file exists, but returns no metadata, assume it's a directory
    if (empty($metaData)) {
        $metaData = [
            'type' => 'dir',
            'path' => $path,
        ];
    }
    return $metaData;
}
```

This results in all of the above three adapters returning these consistent results:
has() on directory that exists: true
has() on directory that doesn't exist: false
getMetadata() on directory that exists: array
getMetadata() on directory that does not exist: exception

If we really want to say directories are second class citizens and that has() should only work for files, then the LocalAdapter should be updated to return false on all directories. Adapters should be considered invalid if they don't conform to a defined behavior.

I'd strongly vote for making has() and getMetadata() actually work with directories consistently though. The alternative makes for really awkward code throughout the rest of your codebase.

kirkmadera:
Resolves #899

### DIFF
--- a/src/Filesystem.php
+++ b/src/Filesystem.php
@@ -54,7 +54,29 @@ class Filesystem implements FilesystemInterface
     {
         $path = Util::normalizePath($path);
 
-        return strlen($path) === 0 ? false : (bool) $this->getAdapter()->has($path);
+        // Root path always exists. Skip the filesystem check
+        if ('' == $path || '.' == $path) {
+            return true;
+        }
+
+        if ($this->getAdapter()->has($path)) {
+            return true;
+        }
+
+        // Deal with inconsistent adapter behavior.
+        // Some adapters always return false for directories.
+        // Check if this is an existing directory by checking if it exists in its parent's contents
+        $parts = explode('/', $path);
+        array_pop($parts);
+        $parentPath = implode('/', $parts);
+
+        foreach ($this->listContents($parentPath) as $parentContent) {
+            if ($path == $parentContent['path']) {
+                return true;
+            }
+        }
+
+        return false;
     }
 
     /**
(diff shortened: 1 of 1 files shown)


src/Filesystem.php
```diff
@@ -54,7 +54,29 @@ public function has($path)
...
+        array_pop($parts);
+        $parentPath = implode('/', $parts);
+
+        foreach ($this->listContents($parentPath) as $parentContent) {
+            if ($path == $parentContent['path']) {
+                return true;
+            }
+        }
+
+        return false;
```
twistor: Sorry, this will wildly slow down every miss.

src/Filesystem.php
```diff
@@ -350,7 +372,22 @@ public function getMetadata($path)
...
+        }
+
+        // If file exists, but returns no metadata, assume it's a directory
+        if (empty($metaData)) {
+            $metaData = [
+                'type' => 'dir',
+                'path' => $path,
+            ];
+        }
+        return $metaData;
```
twistor: This will swallow all sorts of exceptions that could be thrown by an adapter that the developer should be aware of.

twistor:
Thanks for the PR, but this is not a good direction to take.

PR #909, opened 2018-04-05
Catch exceptions thrown by SplFileInfo

Stadly:
Methods such as `SplFileInfo::getType()`, `SplFileInfo::getMTime()` and `SplFileInfo::getSize()` throw `RuntimeException`s when they fail. Such exceptions should be caught and dealt with. In this case, when generating the file info map, the information is simply not populated for calls that fail.

### DIFF
--- a/src/Adapter/Local.php
+++ b/src/Adapter/Local.php
@@ -14,6 +14,7 @@ use League\Flysystem\Util;
 use LogicException;
 use RecursiveDirectoryIterator;
 use RecursiveIteratorIterator;
+use RuntimeException;
 use SplFileInfo;
 
 class Local extends AbstractAdapter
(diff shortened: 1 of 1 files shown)


kubk:
> the information is simply not populated for calls that fail.

It is possible to check the existence of the file before trying to access file info?
For example using ``League\Flysystem\Adapter::has()`` method: https://github.com/thephpleague/flysystem/blob/master/src/Adapter/Local.php#L114
Catching exceptions leads to silent failures, read about Fail Fast design: https://en.wikipedia.org/wiki/Fail-fast

Stadly:
> Catching exceptions leads to silent failures.

My intention was not to create silent failures, but rather to make Flysystem behave as documented. For example, `League\Flysystem\FilesystemInterface::getSize()`'s docblock says:

> @return int|false The file size or false on failure.

Currently, if getting the file size fails, a `RuntimeException` is thrown. With this fix, the exception is caught and the method will return `false`, as documented.

But maybe this error should rather be fixed on another level? What do you think?

frankdejonge:
@Stadly Hi, there are multiple reasons when a call might fail. In Flysystem we've looked at the situations and reasons for failure, check to see if it's somehow expected during normal operations. Things that are caused by invalid configuration (permissions/ownership) are generally not caught because then the developer doesn't know something should change or why/what is going on. It mimics how PHP internally handles filesystem operations, which is generally done in a similar manner although with more warnings (which are a pain, but ok). When your installation is properly configured these calls should never result in runtime exceptions.

Just out of curiosity, how did you end up fixing this? I assume it was a permission issue?

Stadly:
@frankdejonge Hi, I'm creating a library for serving files stored in different file systems, and I'm working on an adapter for `Flysystem`. I discovered this error when creating tests for the library. The test in question checks that my library behaves correctly when a file's size cannot be determined by the adapter. To do this, I provoke a failure in `League\Flysystem\FilesystemInterface::getSize()` simply by asking for the size of a non-existing file. I expected to get `false` returned, but was surprised by the `RuntimeException`.

I realize that checking the size of a non-existing file should generally not happen in normal operation. But, as you mention, there are also other reasons why getting the file size might fail, and I think it would be good to handle these cases, making `Flysystem` behave as documented as often as possible.

Alternatively, we could just add an `@throws RuntimeException` to each relevant method's docblock. Then everyone would be aware of the possibility of the exception, and I would handle it in my own library. But I think it would be cleaner to handle it internally in `Flysystem`.

Stadly:
@frankdejonge I'm calling the method on the `Filesystem` class. Consider this example:

    $FileSystem = new League\Flysystem\Filesystem(new League\Flysystem\Adapter\Local(__DIR__));
    $FileSize = $FileSystem->getSize('foo');

The expected value of `$FileSize` is `false`, but instead a `RuntimeException` is thrown.

frankdejonge:
@Stadly ok, I actually see where the error is now, the main filesystem implementation missed 2 places for assets. This was an actual bug but somewhere else. I've pushed a fix and will release soon.

PR #916, opened 2018-04-22
Properly import PHPCR\SimpleCredentials

holtkamp:
And sort other imports alphabetically


frankdejonge:
@holtkamp hi! I'm reworking all the docs right now so I'll be replacing the current docs entirely, might be better to touch up the docs when that's finished. Agree?

Issue #920, opened 2018-04-25
Allow to explicitly indicate the modification timestamp of a local file


holtkamp:
In some cases it is useful to explicitly indicate the modification timestamp of a (local) file. For example when the content is derived from another (file) system and should have exactly the same timestamp. 

This can be achieved using [`touch()`](http://php.net/manual/en/function.touch.php) after the file has been written. Kind of the same approach of how the visibility is set.

https://github.com/thephpleague/flysystem/blob/f61ec6c5b6567e005b57fe5379f7b25b98458ad2/src/Adapter/Local.php#L136-L139

Would a PR be considered?

frankdejonge:
Could you give a more specific example?

frankdejonge:
I see your problem, but it's not really Flysystem's concern to deal with this. My suggestion would be to write a plugin for this if you want to support this in your application through Flysystem. But more generally speaking, it sounds like you're re-creating what's normally covered by a HTTP cache mechanism like Varnish. If it's almost to complicated to explain, it might be too complicated to begin with.

holtkamp:
Aah, already started a PR... Hopefully we can still discuss this. 

> but it's not really Flysystem's concern to deal with this

Well writing content to a filesystem and use specific timestamp... seems useful to me and other users...

**Context**
A reverse proxy like Varnish has the same effect indeed. We use CloudFlare, but the amount of (possible) files is quite large, so the applications till gets hammered by search engines after a restart, which Heroku does on a daily basis.

holtkamp:
As suggested in https://github.com/thephpleague/flysystem/pull/921#issuecomment-384340005, I had a look on a way to realize setting a specific timestamp using a plugin. 

This seems possible, however, since it is quite a low level operation, we would need access to information that typically only the adapters have. For example:

- when used with LocalAdapter: no way to determine the (root) directory of the file, or the plugin must be aware of this: loss of abstraction
- when used with PHPCRAdapter: no way to access the node, or the plugin should have access to the PHPCR Session: loss of abstraction
  - also see the PR I submitted for this: https://github.com/thephpleague/flysystem-phpcr/pull/8

Leaving my exotic use case out of scope, I still think adopting this functionality in the adapters can be justified. But, for now => closing this issue.

PR #921, opened 2018-04-25
Allow a specific timestamp using the configuration object

holtkamp:
As suggested in https://github.com/thephpleague/flysystem/issues/920


frankdejonge:
@holtkamp as I replied in #920 this is not in Flysystem's concern and should be handled in a plugin.

holtkamp:
Ok, thanks, will try to achieve this using a plugin.

PR #927, opened 2018-05-14
Fix strftime in api doc

TomK32:
According to the php documentation %G is not the year you might expect but can in some years be the year number from the previous year.


frankdejonge:
👍

PR #931, opened 2018-05-17
Suppressed Warnings on unlink and added tests to support change.

guyradford:
A PHP Warning is thrown if the file to delete is missing, this can be due to a race condition.

This fix suppresses the warning allowing unlink to return true/false as expected.

### DIFF
--- a/src/Adapter/Local.php
+++ b/src/Adapter/Local.php
@@ -261,7 +261,7 @@ class Local extends AbstractAdapter
     {
         $location = $this->applyPathPrefix($path);
 
-        return unlink($location);
+        return @unlink($location);
     }
 
     /**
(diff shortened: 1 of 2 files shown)


guyradford:
Wow that was quick! How often does the repo get tagged and released?
Thank you
Guy

frankdejonge:
@guyradford most of the times pretty quickly, I'll see if I can get this out of the door soon. In general I like to release quick and often.

guyradford:
Hi @frankdejonge Any thoughts when this will be tagged?
Thank you

PatrickRose:
> A PHP Warning is thrown if the file to delete is missing, this can be due to a race condition.

That seems like something that developers should know about? `unlink` still returns false so it shouldn't matter if PHP is triggering the error:

```
$ php -a
Interactive mode enabled

php > var_dump(unlink('nonexisting'));
PHP Warning:  unlink(nonexisting): No such file or directory in php shell code on line 1
PHP Stack trace:
PHP   1. {main}() php shell code:0
PHP   2. unlink() php shell code:1

Warning: unlink(nonexisting): No such file or directory in php shell code on line 1

Call Stack:
    8.0272     353688   1. {main}() php shell code:0
    8.0272     353688   2. unlink() php shell code:1

php shell code:1:
bool(false)
```

PR #939, opened 2018-05-25
Introduced new CanOverwriteFiles method in AdapterInterface

mhlavac:
I rebased on master and I am reopening this pull request.

* Problem: CanOverwriteFiles interface can't be decorated. Decorator either implements interface or not
* Proposed solution: Introduce method instead
* Reward: Decorators can call Adapter's CanOverwriteFiles method

The existing interface CanOverwriteFiles can't be easily decorated, because adapter can either implement it or not, but it can't keep original behavior.

The changes are backward compatible, anything that implements the existing CanOverwriteFiles will still work as before. Additional change is that Local, Null and Ftp adapters can overwrite files now to spare one more has call and improve performance.

I noticed this issue when we decorated one of network file systems and we ended up with 2x slowdown, the reason was that the decorated adapter didn't implement the CanOverwriteFiles interface.

### DIFF
--- a/src/Adapter/AbstractAdapter.php
+++ b/src/Adapter/AbstractAdapter.php
@@ -68,4 +68,14 @@ abstract class AbstractAdapter implements AdapterInterface
     {
         return substr($path, strlen($this->getPathPrefix()));
     }
+
+    /**
+     * Can adapter overwrite files
+     *
+     * @return bool
+     */
+    public function canOverwriteFiles()
+    {
+        return false;
+    }
 }
--- a/src/Adapter/AbstractFtpAdapter.php
+++ b/src/Adapter/AbstractFtpAdapter.php
@@ -625,4 +625,14 @@ abstract class AbstractFtpAdapter extends AbstractAdapter
      * @return bool
      */
     abstract public function isConnected();
+
+    /**
+     * Can adapter overwrite files
+     *
+     * @return bool
+     */
+    public function canOverwriteFiles()
+    {
+        return true;
+    }
 }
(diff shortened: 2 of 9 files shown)


frankdejonge:
@mhlavac could you base this PR on the 2.0 branch? This PR is break BC because adapters that purely implement the AdapterInterface will still break without change. I agree with the implementation btw, so it's just a case of which version, which should be 2.0

mhlavac:
@frankdejonge the idea behind this was to still support both. If method is not implemented the interface will be still considered.

If we decide to merge this, then I can also open PR for 2.0 to remove the interface all together, does it make sense?

I'll update this MR with fixed tests.

frankdejonge:
@mhlavac that's not possible without breaking BC. A change in the interface is cause for a major version increase, which this is.

mhlavac:
oh you are right, the only other solution would be to have new interface that has the method. Adapters that would not implement it would behave same as they do now. Adapters that implement it would call the new method to decide if the overwrite is supported or not.

Do you think it's worth to introduce second interface, to allow implementation in current major version?

mhlavac:
@frankdejonge what are the current plans for 2.0? Especially release date? I have few things that might break backwards compatibility, just to know how much should I hurry ;-)

frankdejonge:
@mhlavac there's no release date yet, I'm not yet able to dedicate a fixed amount of time to work on this. At the current sponsoring amount I've got about half an hour of dedicated time. If you have any 2.0 idea's I'll gladly hear them, but bear in mind that the scope of flysystem is not likely to expand.

mhlavac:
@frankdejonge the biggest I have in my mind is to take a look at flysystem error reporting. I would prefer exceptions to returning false. For example in Local adapter, there is no way how to react to false accordingly, because the error is swallowed. Instead with exception, the client code could react to different exceptions accordingly.

So in general, any changes I had in mind are just changes, not adding any new functionality.

Huge thank for working on Flysystem, it's great already! ;-)

frankdejonge:
@mhlavac I've already started converting it to exceptions on the 2.0 branch, so that can be expected.

PR #940, opened 2018-05-26
Handle warnings

Stadly:
Handle warnings generated by `fopen` and `file_get_contents`.

### DIFF
--- a/src/Adapter/Local.php
+++ b/src/Adapter/Local.php
@@ -178,7 +178,12 @@ class Local extends AbstractAdapter
     public function readStream($path)
     {
         $location = $this->applyPathPrefix($path);
-        $stream = fopen($location, 'rb');
+        // Suppress warnings and handle them.
+        $stream = @fopen($location, 'rb');
+
+        if ($stream === false) {
+            return false;
+        }
 
         return ['type' => 'file', 'path' => $path, 'stream' => $stream];
     }
@@ -220,7 +225,8 @@ class Local extends AbstractAdapter
     public function read($path)
     {
         $location = $this->applyPathPrefix($path);
-        $contents = file_get_contents($location);
+        // Suppress warnings and handle them.
+        $contents = @file_get_contents($location);
 
         if ($contents === false) {
             return false;


Stadly:
It is easy to provoke these errors, though I agree that they would normally not happen in a properly configured environment. But that does not mean that it isn't a good idea to handle them ;)

Try for example:
```php
$FileSystem = new \League\Flysystem\Filesystem(new \League\Flysystem\Adapter\Local(dirname(__DIR__)));
$FileSystem->read(basename(__DIR__));
```

Reading a directory is not possible, hence a return value of `false` is expected (based on `read`'s docblock). Instead a warning is raised. Flysystem actually has the code for handling the failure and return `false`. But since the warning is never suppressed, the test checking whether `file_get_contents` succeeded will actually never fail, the way it is now. But a simple warning suppression fixes it all :) And then I did the same for `readStream` (which, contrary to `read`, currently does not check whether `fopen` failed before returning).

In case you wonder, I discovered this while writing a really thoroughly tested file server with an adapter for Flysystem. In one of the tests, I check that reading a directory fails by expecting `false` returned by Flysystem. I was surprised by the warning, which is not in accordance with the Flysystem documentation.

Stadly:
Sorry, I think I'm wrong that the test of whether file_get_contents succeeded never will fail. It is just a warning that is raised, so the script will not stop executing with the default error handler.

frankdejonge:
@Stadly I understand it from a "code" point of view, but for the users it provides no actual value. Instead they'll be less likely to get the information (the warning) that they need to fix it. Their integrations should be tested, and this should run in the tests without errors.

Stadly:
@frankdejonge I think I agree :) We might consider adding in the documentation of these methods that they may raise warnings. I'll just close this PR, as error suppression is unwanted, and the test in `readStream` is unnecessary since `Filesystem::readStream` will return false even though the adapter returns an array.

stephanvierkant:
> Sorry, I think I'm wrong that the test of whether file_get_contents succeeded never will fail. It is just a warning that is raised, so the script will not stop executing with the default error handler.

Well, in my case this causes a problem. Symfony throws a `Symfony\Component\Debug\Exception\ContextErrorException` when the file is not found. This stops execution.

Those crappy throwing/returning warning/exception/null/false thing in PHP is what I really hate and I'm happy that's getting better and better with each version, but it's still a pain.

PR #951, opened 2018-07-10
[CSS] Fix the overflow of code snippets

absolux:


frankdejonge:
Hi @absolux, could you update the source stylesheet instead or the template? The dist file should not be directly edited.

absolux:
@frankdejonge could you help with the appveyor failure

PR #954, opened 2018-07-23
Idealized Appveyor config that only took all day.

twistor:
Sunk some time getting this to work for https://github.com/twistor/flysystem-stream-wrapper.

- Use Chocolatey to install the latest release of the PHP and Composer.
- I don't see any reason to check if PHP is cached, file-system operations are trivial.
- Check lowest and highest deps.
- I think Appveyor has a cache size limit, so don't want to check every intermediate PHP version. Maybe could though.
- Set minimum PHPUnit version to `5.7.10`. Removes 7.2 problems.
-  Run phpspec
- I don't even want to start with docker and the integration tests.
- :tada:


frankdejonge:
Awesome! 👍👍

PR #978, opened 2018-10-22
SFTP usually uses port 22

simoheinonen:
21 is usually plain FTP


frankdejonge:
Thanks @simoheinonen!

PR #982, opened 2018-11-02
Local::ensureDirectory - improve error message; clearstatcache

hutnikau:
That would be great to be able to get the error message with exact reason why the directory was not created.

Also the result of `is_dir` function is cached. So if directory was created by another php process, then it still will return `false`. So the cache should be cleared before the second `is_dir` call

### DIFF
--- a/src/Adapter/Local.php
+++ b/src/Adapter/Local.php
@@ -105,7 +105,8 @@ class Local extends AbstractAdapter
             umask($umask);
             clearstatcache();
             if ( ! is_dir($root)) {
-                throw new Exception(sprintf('Impossible to create the root directory "%s". ' . $mkdirErrorArray['message'], $root));
+                $errorMessage = isset($mkdirErrorArray['message']) ? $mkdirErrorArray['message'] : '';
+                throw new Exception(sprintf('Impossible to create the root directory "%s". %s' , $root, $errorMessage));
             }
         }
     }
--- a/src/Adapter/Local.php
+++ b/src/Adapter/Local.php
@@ -103,7 +103,6 @@ class Local extends AbstractAdapter
                 $mkdirErrorArray = error_get_last();
             }
             umask($umask);
-            clearstatcache();
             if ( ! is_dir($root)) {
                 $errorMessage = isset($mkdirErrorArray['message']) ? $mkdirErrorArray['message'] : '';
                 throw new Exception(sprintf('Impossible to create the root directory "%s". %s' , $root, $errorMessage));
--- a/src/Adapter/Local.php
+++ b/src/Adapter/Local.php
@@ -99,11 +99,13 @@ class Local extends AbstractAdapter
     {
         if ( ! is_dir($root)) {
             $umask = umask(0);
-            @mkdir($root, $this->permissionMap['dir']['public'], true);
+            if( ! @mkdir($root, $this->permissionMap['dir']['public'], true)){
+                $mkdirErrorArray = error_get_last();
+            }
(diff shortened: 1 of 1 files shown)


hutnikau:
Hi @frankdejonge , could you please look at this PR. This change fixes an issue that may occur when two concurrent processes attempt to create a directory with the same name.

twistor:
PHP doesn't cache lookups that do not exist.

http://php.net/manual/en/function.clearstatcache.php

Try this in one terminal:
```php
<?php

while (!is_dir('foo')) {
    echo "not dir\n";
    sleep(1);
}
```

then in another terminal:
```sh
mkdir foo
```

src/Adapter/Local.php
```diff
@@ -99,11 +99,13 @@ protected function ensureDirectory($root)
...
-            @mkdir($root, $this->permissionMap['dir']['public'], true);
+            if( ! @mkdir($root, $this->permissionMap['dir']['public'], true)){
+                $mkdirErrorArray = error_get_last();
+            }
             umask($umask);
-
+            clearstatcache();
             if ( ! is_dir($root)) {
-                throw new Exception(sprintf('Impossible to create the root directory "%s".', $root));
+                throw new Exception(sprintf('Impossible to create the root directory "%s". ' . $mkdirErrorArray['message'], $root));
```
twistor: The test failure here is real. `$mkdirErrorArray` might not be set.

hutnikau:
Hello @twistor ,

PHP caches the result of `is_dir` call even if it returned `false`. Please try to run this code:
`create_dir.php`:
```php
$path = __DIR__ . DIRECTORY_SEPARATOR . 'foo';
mkdir($path);
echo 'directory created' . PHP_EOL;
```
`remove_file.php`:
```php
$path = __DIR__ . DIRECTORY_SEPARATOR . 'foo';
unlink($path);
echo 'file deleted' . PHP_EOL;
```
`index.php`:
```php
<?php
$path = __DIR__ . DIRECTORY_SEPARATOR . 'foo';
file_put_contents($path, '');
var_dump(is_dir($path));

if (!@mkdir($path)) {
    $mkdirErrorArray = error_get_last();
}
var_dump($mkdirErrorArray);
$output = shell_exec('php ./remove_file.php');
$output = shell_exec('php ./create_dir.php');

var_dump(is_dir($path)); //false
clearstatcache();
var_dump(is_dir($path)); //true
```

result:
```
index.php:4:boolean false
index.php:9:
array (size=4)
  'type' => int 2
  'message' => string 'mkdir(): File exists' (length=20)
  'file' => string 'D:\domains\playground\index.php' (length=31)
  'line' => int 6
index.php:13:boolean false
index.php:15:boolean true
```

boajer:
Jumping into this topic, I see exactly same behaviour with sample above on `PHP 7.2.12 (cli) (built: Nov  9 2018 11:00:45) ( NTS )` at MacOsX

twistor:
@hutnikau

> Also the result of is_dir function is cached. So if directory was created by another php process, then it still will return false. So the cache should be cleared before the second is_dir call

So, what you really mean is that the directory was a file, but it was deleted and made into a directory in another process.

I would say that's a very strange edge case.

 Your example code does a very different thing than what you described. What we really need is another check after the first `is_dir()` to check if it's a file, link, whatever. A simple `file_exists()` should cover it. Then throw an exception describing the situation.

twistor:
Calling `clearstatcache()` should be avoided. Especially without both arguments.

PR #984, opened 2018-11-11
Updated README.md with link to new Google Drive adapter

masbug:
Hi,

I've made an adaptation of flysystem-google-drive adapter which allows you to use regular paths instead of paths built from unique identifiers. All ID handling is performed in the adapter (also with heavy built-in caching).
For example:
"native google drive path" **/Xa3X9GlR6EmbnY1RLVTk5VUtOVkk/0B3X9GlR6EmbnY1RLVTk5VUtOVkk** 
becomes **/My Nice Dir/myFile.ext**

I also included a fix for Google_Http_MediaFileUpload which reduces memory usage by using Psr7 streams and sends everything in chunks instead of loading the whole file in memory before sending.

This PR request adds a link to the new Google Drive adapter.

### DIFF
--- a/README.md
+++ b/README.md
@@ -84,6 +84,7 @@ Want to get started quickly? Check out some of these integrations:
 * Gaufrette: https://github.com/jenkoian/flysystem-gaufrette
 * Google Cloud Storage: https://github.com/Superbalist/flysystem-google-storage
 * Google Drive: https://github.com/nao-pon/flysystem-google-drive
+* Google Drive V2 (using regular paths): https://github.com/masbug/flysystem-google-drive-ext
 * OneDrive: https://github.com/jacekbarecki/flysystem-onedrive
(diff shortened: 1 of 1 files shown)


GrahamCampbell:
> I also included a fix for Google_Http_MediaFileUpload which reduces memory usage by using Psr7 streams and sends everything in chunks instead of loading the whole file in memory before sending.

Why not contribute this back to the original package?

masbug:
> Why not contribute this back to the original package?
Fair point.

The reason is that most of the code changed and it also completely changes how the adapter is used. In the original adapter you had to access files/folders via unique identifiers which is quite different from other adapters here. All the client code had to be adapter specific...and that is far away from what flysystem stands for. My implementation aims to correct that.

I've been using my implementation in the production code for about a year now and appart from 1 bug no other bugs were found...so I figured it's ready for wider use.

PR #986, opened 2018-11-23
Added RDF extension to mimetype util

martijn-tao:
When retrieving files with the .rdf extension we get `text/plain` as a ContentType. We would however expect to see `application/rdf+xml` set as a ContentType.

This pull request adds the rdf extension to the mimetype util, and resolves this issue.


frankdejonge:
👍 looks good, I'll release this today

Issue #985, opened 2018-11-14
Case Sensitive with Dropbox


davemin:
A possible bug in adapter for dropbox. function isDirectChild, works only if return always true, can you fix!???

frankdejonge:
@davemin I've opened up a PR you can try: #987

frankdejonge:
Should be solved with the new case sensitivity option which will be shipped in 1.0.50.

PR #987, opened 2018-11-23
Allow for case-insensitive directory listings

frankdejonge:
This should resolve the issues with the dropbox adapter which can cause case-insensitive listings.


frankdejonge:
Related: https://github.com/thephpleague/flysystem/issues/985

freekmurze:
Thanks for your work on this @frankdejonge 

I'd appreciate people using this package extensively to test this out and let me know if it works or not.

I'll also test it out myself early next week and merge it then.

OskarStark:
Ok, no random, it doesn't work:
```php
    /**
     * @Route("/refresh")
     *
     * @return \Symfony\Component\HttpFoundation\Response
     */
    public function refreshAction(Filesystem $dropbox)
    {
        dump($dropbox->listContents('my/path'));

        $dropbox->getAdapter()->getCache()->flush();

        dump('---------');
        dd($dropbox->listContents('my/path'));
        return $this->redirectToRoute('app_default_index');
    }
```

result:
screenshot 2018-11-27 09 36 59

and I removed the 3 `lazy: true` options...

frankdejonge:
Ah yeah, that's an issue with the cache-adapter, not this adapter.

frankdejonge:
Or at least, it's not an issue, the cache-adapter works, but it's incompatible with the logic of case-insensitive paths.

OskarStark:
> Or at least, it's not an issue, the cache-adapter works, but it's incompatible with the logic of case-insensitive paths.

thats my thought too... but normally it shouldn't care..

WYT of `caseSensitive(): bool` method in Adapter?

frankdejonge:
I think the cache adapter needs to be aware of case-sensitivity too, just like the Filesystem is now. I don't like the capability getters for configuration.

frankdejonge:
BTW why are you caching dropbox responses in the first place? I'd expect you'd always want the not cached response because users will change it from somewhere else and then the cache is a lie.

OskarStark:
> BTW why are you caching dropbox responses in the first place? I'd expect you'd always want the not cached response because users will change it from somewhere else and then the cache is a lie.

In this case I am the only one who uses it and the problem is, the Dropbox is quite large (~43 TB) and my feeling is, that it gets slower if it gets bigger, but I could be wrong...

> I don't like the capability getters for configuration.

me neither, but it belongs to the Adapter

OskarStark:
Any news here?

frankdejonge:
This can be merged if no other case is made other than the case-insensitive cache. The caching adapter needs adjustments like this. I'm not a proponent of making the caching aware of any special requirements an adapter has, so I won't be adding that. I'm waiting on the spatie people to give me their feedback too, see if we've missed something and then I'll merge and tag.

OskarStark:
In this case you should add a note in the docs, that this can lead to
unexpected behavior.
And in this (Cache-) case the option is useless for me... 😢

Frank de Jonge <notifications@github.com> schrieb am Di. 4. Dez. 2018 um
08:55:

twistor:
I haven't seen it mentioned in this issue, but what about mbstring support? Seems like the current implementation is going to bite us.

With that in mind, I'm not sure why we're adding this here, and not just fixing Dropbox directly.

rubenvanassche:
@frankdejonge Sorry for the late response, we've tested this with the Spatie Dropbox package and everything works as it should. Thanks!

frankdejonge:
@twistor this is the result of a tradeoff. The filesystem removed any excess listings, so it's a safeguard, but Dropbox returns something is "should not". I put that in quotes because Dropbox is case insensitive while all the others are not. So the listing it returns doesn't match the prefix, which causes the result to be filtered out. In the time this was introduced this was because then only one place needed to do the filtering instead of in all the adapters. This can be changed but that would warrant a 2.0 I think.

PR #989, opened 2018-12-03
Fixed some inconsistencies in API docs

alexhinds:
Just some variable name inconsistencies and I believe an incorrect return type.


frankdejonge:
@alexhinds thank you and hooray for your first contribution to Flysystem!

PR #998, opened 2019-01-16
Cache clear added to avoid potential issues with concurrent processes

boajer:
Inspired by https://github.com/thephpleague/flysystem/pull/982

### DIFF
--- a/src/Adapter/Local.php
+++ b/src/Adapter/Local.php
@@ -106,6 +106,7 @@ class Local extends AbstractAdapter
             }
 
             umask($umask);
+            clearstatcache(false, $root);
 
             if ( ! is_dir($root)) {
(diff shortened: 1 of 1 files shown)


frankdejonge:
Can you describe the issue and error you're encountering?

Alroniks:
@frankdejonge As I see from the previous discussion in the https://github.com/thephpleague/flysystem/pull/982 should be called function `clearstatcache` but with specified parameters. Accidentally it was removed in the last commit instead of adding the parameters. 
https://github.com/thephpleague/flysystem/pull/982/commits/0360506f45e3694fb208d2e827eb80527fc6c6a3 - comment sounds "clearstatcache for a specific dirname only" but in the code this function just removed. So this PR returns proper call if this function.

PR #1001, opened 2019-01-28
fix some typo in docs about `Create Directories`

bestony:
the Create Directories has some typo of the description and type for $response.

change it same to Delete Directories


frankdejonge:
Thanks @bestony! 👍

PR #1004, opened 2019-02-01
Change adapter Local::listContents() to a generator

janhopman:
To prevent extensive memory usage on reading large directories.


frankdejonge:
This change will happen in 2.0, but not in 1.0 because of the BC break.

janhopman:
Thanks for the quick reply. Do you have any clue on when 2.0 will be released?

PR #1005, opened 2019-02-14
[DOC] Update S3Client invocation

jrjohnson:
S3Client::factory is deprecated and this is the right way to instantiate
the class.

I'm not sure about how this works in the v2 library so I left it alone.


frankdejonge:
@jrjohnson thanks!

PR #1007, opened 2019-02-19
get ThumbnailLink

akinayturan:


frankdejonge:
Hi @akinayturan I understand the need for such a function Flysystem is not the place for this. In your application you're advised to create a boundary around Flysystem where you can facilitate such functionalities. So for example, you can create a `ImageStorage` with a Flysystem based implementation and this would internally just fetch the metadata and/or make additional calls to specific adapters. This ensures you are in control of how this all happens. It also protects Flysystem from having to implement none-filesystem specific functionality. For example, what if somebody uses Flysystem for storing video's. Should flysystem expose the thumbnails for that too? What about MP3's and previews? I hope you see my point.

PR #1008, opened 2019-02-24
mkdir() side effects

shehi:
mkdir() may emit E_WARNING type error, which can transform into \ErrorException exception.


frankdejonge:
This PR removed the calling of `mkdir`, which is not correct.

PR #1013, opened 2019-03-20
Ftp::listContents support for 'timestamp' attribute

tristanveness:
I've been using the 'timestamp' property to sort/filter files on local and ftp filesystems.

Calls to listContents depending on the 'timestamp' property had been working on some FTP servers and crashing on others (due to missing 'timestamp'), discovered it was due to 'windows' listings returning timestamps, but 'unix' listings not.

This patch provides returning a best-effort (given the limitations of the 'LIST' command) listContents timestamp for FTP servers using 'unix' style listings.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -1,5 +1,10 @@
 # Changelog
 
+## 1.0.51 - 2019-03-20
+
+* [Ftp::listContents] Added support to return 'timestamp' attribute. Note that accuracy is limited
+  due to limitations in the 'LIST' command.
+
 ## 1.0.50 - 2019-02-01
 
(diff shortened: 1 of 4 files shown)


frankdejonge:
Hi @tristanveness this looks good. 👍 I would however like to see this be an opt in feature based on a configuration option in the adapter. This is due to to the inaccuracy in combination with caching. If this in accurate value is returned from a listing (and therefor cached) you can't get a more accurate one. So it's really a trade-off that users need to be aware of.

tristanveness:
Thanks Frank! I'll update with a configuration option.

tristanveness:
@frankdejonge should be good to go now, feel free to amend (naming, etc)

PR #1021, opened 2019-04-21
Added missing equal sign in documentation

supergrecko:
I noticed there is a missing equals sign in the docs.


frankdejonge:
Thanks @supergrecko! 👍

PR #1029, opened 2019-05-28
Don't guess mimeType when it's known

Mechazawa:
Right now in our API calling the guessMimeType method is causing out of memory exceptions because it loads the entire file into a buffer. 
See `return $finfo->buffer($content) ?: null;` in `MimeType.php`

### DIFF
--- a/src/Adapter/Ftp.php
+++ b/src/Adapter/Ftp.php
@@ -249,7 +249,7 @@ class Ftp extends AbstractFtpAdapter
         }
 
         $result['contents'] = $contents;
-        $result['mimetype'] = Util::guessMimeType($path, $contents);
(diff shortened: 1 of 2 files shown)


Mechazawa:
Could this please be reviewed? This is currently causing memory issues for us.

Mechazawa:
Can this PR please be considered? This is causing a lot of issues for us in production at the moment.

frankdejonge:
I've researched this, but when dealing with large files you shouldn't be getting this for sufficiently large files. Are you using the stream based writing? If you're using string based writing you're going to run into memory issues.

Mechazawa:
We are using string based writing because we're receiving JSON files through our API, parsing them and we then write them to an S3 bucket. We sadly don't have the option to use streams unless we first write the file to disk (which I'd want to avoid).

PR #1036, opened 2019-06-18
PHPStan level 5 + upgrade to 0.11.8

mhlavac:
* Updated phpstan to version 0.11.8
* Bumped phpstan level to 5
* Fixed clearstatcache location being used as 1st parameter instead
  as 2nd parameter
* Fixed config fallback incorrect phpdoc

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -60,5 +60,8 @@
         "branch-alias": {
             "dev-master": "1.1-dev"
         }
+    },
+    "scripts": {
+        "phpstan": "php phpstan.php"
     }
 }
--- a/composer.json
+++ b/composer.json
@@ -19,7 +19,8 @@
     },
     "require-dev": {
         "phpspec/phpspec": "^3.4",
-        "phpunit/phpunit": "^5.7.10"
+        "phpunit/phpunit": "^5.7.10",
+        "phpstan/phpstan": "^0.11.8"
     },
     "autoload": {
         "psr-4": {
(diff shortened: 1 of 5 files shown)


frankdejonge:
The clearstatscache call was already fixed in master.

mhlavac:
@frankdejonge great, I just saw you merged the other pull request. I'll rebase this on master in a minute. I was surprised that phpstan found the issue as well.

Would you mind if I PR local dev set up, so php-cs-fixer, phpstan and other tools can be run via single command in local? I think it will ease the contribution effort

frankdejonge:
Can you set that up as a composer script?

mhlavac:
Yes, I think that should be possible

mhlavac:
Rebased and updated the PR to reflect what it changes

frankdejonge:
I'm about to do a release, so this will at the earliest go into the next release.

mhlavac:
I'll open separate PR for adding the scripts as composer scripts

composer.json
```diff
@@ -19,7 +19,8 @@
     },
     "require-dev": {
         "phpspec/phpspec": "^3.4",
-        "phpunit/phpunit": "^5.7.10"
+        "phpunit/phpunit": "^5.7.10",
+        "phpstan/phpstan": "^0.11.8"
```
frankdejonge: adding this dev dependency causes unwanted failures in the test suite, please check the travis output.

mhlavac:
@frankdejonge fixed, I had hard times to get travis-ci script to work, so instead I implemented short php script to download phpstan shim that does the job. Is it reasonable?

PR #1047, opened 2019-07-18
Fixed listing of the existing directory in Ftpd

SilverFire:
Follows the parent class fix 5901bb55
Before the change, calling getMetadata() over an existing directory produced a fatal error.

```
PHP Fatal error:  Uncaught RuntimeException: Metadata can't be parsed from item 'total 334920' , not enough parts. in /app/vendor/league/flysystem/src/Adapter/AbstractFtpAdapter.php:435
```

Ftpd produces the following output on `STAT` call over directory:

```
array(46) {
  [0]=>
  string(32) "212- Status of content10/export:"
  [1]=>
  string(12) "total 334920"
  [2]=>
  string(72) "-rw-r--r--  1 1000  1000   3645884 Jul 18 06:10 movie.ts"
  
  [....]

  [45]=>
  string(18) "212 End of status."
}
```

Then, it can not be parsed by `normalizeObject`

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## 1.0.54 - 2019-08-11
+
+* Fixed metadata fetching of the existing directory in Ftpd.
+
 ## 1.0.53 - 2019-06-18
 
 * Clear stat cache before getting file metadata.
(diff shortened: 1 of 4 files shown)


SilverFire:
Ok, I see the tests fail. Will check details later

SilverFire:
Fixed tests, ready for review

PR #1063, opened 2019-09-02
Feature/docs default credential provider usage

wired00:
- Added section to docs - `adapter/aws-s3-v3` for `Default credential provider usage`. There is no need to specifically include `credentials.key` and `credentials.secret` configuration values when IAM Role is already assigned to the EC2. The documentation should point this out.


frankdejonge:
Thank you for this! 👍

PR #1067, opened 2019-09-18
Correct "disc" to "disk"

MrGabigoo:
"Disc" (with a C) is used for audio records, CDs and the like, whereas a computer hard-drive is a "disk" (with a K)


frankdejonge:
👍

PR #1068, opened 2019-09-19
Added flysystem-raid to Community Supported Adapters list

PHPGuus:
As per twitter messages between Frank and Guus


frankdejonge:
👍👍

PR #1070, opened 2019-09-26
Refactoring for capture exceptions 

eddycek:
The change will allow you to better capture particular exceptions and catch all the exceptions by iterface `FlysystemException ` (the main thing I needed).

This is not a BC break, as the new expetion inherits from the original. As evidenced by successful tests, see https://github.com/eddycek/flysystem/blob/refactoring-exception/tests/FtpTests.php#L446 and https://github.com/eddycek/flysystem/blob/refactoring-exception/tests/FtpTests.php#L794 etc.

The related PR has already been approved in `flysystem-sftp`, see https://github.com/thephpleague/flysystem-sftp/pull/86

### DIFF
--- a/src/Adapter/Ftp.php
+++ b/src/Adapter/Ftp.php
@@ -6,9 +6,11 @@ use ErrorException;
 use League\Flysystem\Adapter\Polyfill\StreamedCopyTrait;
 use League\Flysystem\AdapterInterface;
 use League\Flysystem\Config;
+use League\Flysystem\ConnectionErrorException;
+use League\Flysystem\ConnectionRuntimeException;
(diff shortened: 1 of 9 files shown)


eddycek:
@frankdejonge Any chance of this getting merged or at least some estimation when/if it might be? Thanks :)

GrahamCampbell:
This looks like quite a modest change, in principle.

eddycek:
@frankdejonge hello, how is it with merging?

frankdejonge:
I've altered some naming and made the marker interface specific for exceptions. Will release soon.

eddycek:
@frankdejonge thank you 👍

PR #1072, opened 2019-10-03
Fix config parameter name for S3.

alexsegura:
It is actually `endpoint` that should be used

See https://github.com/aws/aws-sdk-php/blob/0326ba62f6ee0bb075f2b875631ec4abee93bee1/src/AwsClient.php#L107-L109


frankdejonge:
Thanks @alexsegura! 👍

Issue #1051, opened 2019-08-03
Specify permissions using strings


tontonsb:
### Feature Request

|    Q        |   A
|------------ | ------
| New Feature | yes
| BC Break    | yes

#### Scenario

There's a possibility to specify "visibility => permission"  mapping for the local adapter. While the syntax is fine for specifying `0775` instead of `0755`, a catastrophe occurs if you naively set the first octet. `1777` actually sets the permissions to `3361`. To set the `1777` permissions one actually has to specify either the octal `01777` or the decimal `1023`. 

#### Summary

I propose to allow optional string syntax for permissions. The strings should be translated to permissions as naively expected - `'1777'` should actually set `1777` permissions.

This syntax should also be used in the codebase so the example in the code would be the one that behaves as would be expected.

#### How is this breaking?

`'sticky' => 01777` works and would keep working. `'sticky' => 1023` works and would keep working. `'sticky' => 1777` is crazy, but would keep working as it does.

The case that breaks is where one would already be using strings. Undocumented but it works. `'sticky' => '1777'` currently sets `3361`, but according to my proposal this stringed directive should set permissions to `1777`.

frankdejonge:
I get what you're saying, but I don't want to bump a major for this. We could ship with an added utility function that safely translates the string permissions to their octal counterparts, this way there's no BC break but people can use the tooling. It would even accept either type and convert based on the need.

tontonsb:
Sure, it coud be as simple as

```php
function mode($mode)
{
	if (!is_string($mode))
		return $mode;

	return octdec($modeString);
}
```

Or, tu support those fancy `drwSrwxr-T` something like this would work

```php
function mode($modeString)
{
	if (!is_string($modeString))
		return $modeString;

	if (is_numeric($modeString))
		return octdec($modeString);
	
	$mode = 0;
	$special = 0;

	for ($i = 0; $i < strlen($modeString); $i++)
	{
		$mode = $mode << 1;
		
		if (0 == $i % 3)
		    $special = $special << 1;

		$char = $modeString[$i];
		
		if (in_array($char, ['S', 's', 'T', 't']))
			$special |= 1;

		if (in_array($char, ['-', 'S', 'T', 'd']))
			continue;

		$mode |= 1;
	}

	return $mode |= ($special << 9);
}
```

PR #1076, opened 2019-10-06
Add a mode conversion utility

tontonsb:
A PR of what I proposed in #1051 . The current problem is that if one naively changes `'public' => 0755` to `'public' => 1755` the result is not `rwxr-xr-t` but `-wx-ws-wt` (`03333`).

This adds a utility function that converts mode strings to correct represntations.

- `Mode::mode(0755)` and `Mode::mode(1755)` just passes it through.
- `Mode::mode('0755')` and `Mode::mode('1755')` assumes the chars to be octets and sets mode to `rwxr-xr-x` and `rwxr-xr-t` respectively - that works according to naive scenario. `Mode::mode('01755')` works as well.
- `Mode::mode('rwxr-xr-x')` sets mode to `rwxr-xr-x`. Adding `d` or placing guid, suid and sticky bits also work as expected, both with executable bit on or off.

Each of the features described above is tested as well.

I went ahead and used the function in `Local` adapter. That's the purpose of this PR - to provide the local adapter with a default that's safe to copy and edit without running into unexpected results.

### DIFF
--- a/src/Adapter/Local.php
+++ b/src/Adapter/Local.php
@@ -69,7 +69,7 @@ class Local extends AbstractAdapter
                 'private' => Util\Mode::mode('0700'),
             ],
         ];
-        
+
         $root = is_link($root) ? realpath($root) : $root;
         $this->permissionMap = array_replace_recursive($defaultPermissions, $permissions);
         $this->ensureDirectory($root);
--- a/src/Adapter/Local.php
+++ b/src/Adapter/Local.php
@@ -28,20 +28,6 @@ class Local extends AbstractAdapter
     const DISALLOW_LINKS = 0002;
 
     /**
-     * @var array
-     */
-    protected static $permissions = [
-        'file' => [
-            'public' => Util::mode('0644'),
-            'private' => Util::mode('0600'),
-        ],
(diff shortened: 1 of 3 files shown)


src/Adapter/Local.php
```diff
@@ -73,8 +59,19 @@ class Local extends AbstractAdapter
...
+            'file' => [
+                'public' => Util\Mode::mode('0644'),
+                'private' => Util\Mode::mode('0600'),
+            ],
+            'dir' => [
+                'public' => Util\Mode::mode('0755'),
+                'private' => Util\Mode::mode('0700'),
+            ],
+        ];
+
```
ms-slasher13: codes tyle issue: remove empty spaces here!

ms-slasher13:
Created pull request: https://github.com/tontonsb/flysystem/pull/1

frankdejonge:
I've designed a new way to facilitate this for V2, which will allow for more flexibility. I suggest using this as a standalone package for now. I won't be merging this in because it would only be supported for a little while. Thanks for your contribution, it helped spark the design process of the new method.

PR #1078, opened 2019-10-08
Correct to correct use of capitals in GitLab

DannyvdSluijs:
In order to complete the chin addition to #1075 this PR will correct Gitlab to GitLab in the left menu under Adapters


frankdejonge:
Thanks @DannyvdSluijs 👍

PR #1079, opened 2019-10-08
Describe optional config param and possible false return in case of failure

DannyvdSluijs:
I've added the optional `$config` parameter where allowed using the square brackets (like PHP.net does) to indicate it isn't mandatory. Also I've added the `false` as a union type when ever a `string` content is returned or false in case of failure.


frankdejonge:
Thanks you, @DannyvdSluijs! 👍

PR #1089, opened 2019-10-30
consistent spelling of AWS S3

wernerkrauss:
afaik it's called "AWS S3" or "Amazon S3", not "AwsS3". Also in the docs it's written like this.


frankdejonge:
Thanks 👍

PR #1090, opened 2019-10-30
add silverstripe usage

wernerkrauss:
SilverStripe CMS uses flysystem as its filesystem abstraction layer.

I explained the small difference to the default flysystem adapters and listed the available adapter wrappers for SilverStripe.

I also noticed, that some tailwind bg colors are redundant and removed them.


frankdejonge:
Hi @wernerkrauss, thanks you for the contribution. The bg colors are purged in the build process, so it should be fine to have them in the config as it doesn't affect the dist size.

Issue #791, opened 2017-05-07
is_readable issues with Samba/UNC paths


SylvesterNielsen:
When i create a filesystem using the Local adapter using a samba share/UNC path or mounted drive name i get an error at initialization due to is_readable.
There has been a few bug reports about the issue, but i find no fixes unfortunately.

The adapter and flysystem actually does work, but as the constructor throws an exception we can't use it. 

So i would love to know if it could be possible to have the is_readable part of the constructor as a optional step. I have tested by simply removing the is_readable check from my local version and all seems to work without any issues. Read/write/delete/list.

There's a couple of PHP bugreports about the issue, but as far as i can see no fixes.

* https://bugs.php.net/bug.php?id=49620
* https://bugs.php.net/bug.php?id=62199 
* http://stackoverflow.com/questions/10818770/php-is-readable-fails-on-readable-samba-directory

Surt:
Same problem here. It's a shame. I'm trying to create a docker application that writes on SAMBA paths. The is_readable (from the php bug) is causing the app to not work, but the system CAN write and read. (I can do it with `file_put_contents` for example. The problem is the "is_read"

broberts-mrmc:
For those who have encountered this "UNC path & is_readable()" issue using Flysystem, I wanted to share our solution in case it's helpful.

Since Flysystem's architecture uses the adapter pattern (https://flysystem.thephpleague.com/docs/architecture/), it's inherently extensible in way that can be used to solve this problem.  Basically, to side-step the issue with PHP's built-in function `is_readable()` and UNC/Samba paths, we created a custom "local" adapter that's identical to the one that ships with Flysystem except for the problematic call to `is_readable()`, and placed it within our namespace.

I've attached the source.  Be sure to change the namespace declaration at the top (line 3) to suite your project's namespace/directory structure, and provide an instance of that adapter to Flysystem's `Filesystem()` constructor.

Custom Local Adapter: [Local.zip](https://github.com/thephpleague/flysystem/files/3571714/Local.zip)

PR #1093, opened 2019-11-07
Built-In PHP Function Problematic in Some Cases

broberts-mrmc:
In certain PHP releases, there's a known issue (https://bugs.php.net/bug.php?id=62199) in PHP's built-in `is_readable()` function and SMB file shares. The PHP built-in `is_writable()` function (alias `is_writeable()`) appears similarly affected.

In certain cases, these functions incorrectly return `FALSE` for a given UNC file path even though the target is, in fact, readable in the case `is_readable()`, or writable in the case of `is_writable()`.

Until such a time as this bug is fixed, I'd like to suggest a small work-around (or something like it) for the most reliable use of UNC paths in the `League\Flysystem\Adapter\Local` class.

Reference:
* https://www.php.net/manual/en/function.is-readable.php
* https://www.php.net/manual/en/function.is-writable.php
* https://www.php.net/manual/en/function.is-writeable.php
* https://bugs.php.net/bug.php?id=62199
* https://github.com/thephpleague/flysystem/issues/791

### DIFF
--- a/src/Adapter/Local.php
+++ b/src/Adapter/Local.php
@@ -77,7 +77,17 @@ class Local extends AbstractAdapter
         $this->permissionMap = array_replace_recursive(static::$permissions, $permissions);
         $this->ensureDirectory($root);
 
-        if ( ! is_dir($root) || ! is_readable($root)) {
+        // If UNC path...
+        if (substr($root, 0, 2) === str_repeat($this->pathSeparator, 2)) {
+
+            $readable = file_exists($root);
+
+        } else {
+
+            $readable = is_readable($root);
+        }
+
+        if ( ! is_dir($root) || $readable === false) {
             throw new LogicException('The root path ' . $root . ' is not readable.');
         }
 


frankdejonge:
If you're experiencing this issue I recommend to work around it in your project, this can be done by extending the adapter and using that instead of the one shipped. It's not the responsibility to work around PHP version specific issues, there's many types of issues like this over the life-time of the project and accounting for all of them is not a responsibility the project should bear.

frankdejonge:
I understand it may be a dissatisfying resolution to your problem, but I hope you understand my position and see that it's quite easy to work around on a project level.

broberts-mrmc:
> I understand it may be a dissatisfying resolution to your problem, but I hope you understand my position and see that it's quite easy to work around on a project level.

Yes, we're already using the work-around method described (i.e. our own custom `Local` adapter), but was hoping a workaround/fix would make it into the project itself.  Of course, I understand/respect your stance on it.

broberts-mrmc:
> I understand it may be a dissatisfying resolution to your problem, but I hope you understand my position and see that it's quite easy to work around on a project level.

A custom `Local` adapter is easy enough, though the downside is that other fixes/changes to the one that's shipped with the project, if desired in the custom adapter, must be back ported, so to speak.

PR #1094, opened 2019-11-13
Local adapter umask MUST be removed

MockingMagician:
umask can lead to unexpected behavioir on multithreaded webservers. 

See notes on https://www.php.net/manual/en/function.umask.php :

>Avoid using this function in multithreaded webservers. It is better to change the file permissions with chmod() after creating the file. Using umask() can lead to unexpected behavior of concurrently running scripts and the webserver itself because they all use the same umask.

### DIFF
--- a/src/Adapter/Local.php
+++ b/src/Adapter/Local.php
@@ -131,6 +131,7 @@ class Local extends AbstractAdapter
 
     /**
      * @inheritdoc
+     *
      * @throws Exception
      */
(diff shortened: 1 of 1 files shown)


frankdejonge:
What problem are you running into?

frankdejonge:
Closing due to the lack of response.

GrahamCampbell:
I think the issue they are running into is that PHP caches the is_dir and is_file stuff, which can cause race conditions?

PR #1095, opened 2019-11-14
Improve FTP connection healthcheck speed

Greg0:
Listing files can be very slow. `NOOP` command is enough in most cases and it is very fast.
NOOP reference - http://ftpguide.com/NOOP.htm

I profiled my code by blackfire and that change increased speed 10 times. Usage of `ftp_rawlist` is bottleneck.

### DIFF
--- a/src/Adapter/Ftp.php
+++ b/src/Adapter/Ftp.php
@@ -531,12 +531,8 @@ class Ftp extends AbstractFtpAdapter
     public function isConnected()
     {
         try {
-            return is_resource($this->connection) && ftp_rawlist($this->connection, $this->getRoot()) !== false;
+            return is_resource($this->connection) && ftp_exec($this->connection, 'NOOP');
         } catch (ErrorException $e) {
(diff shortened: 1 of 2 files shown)


Greg0:
If you finding this solution good, tell in comments then I will fix tests

frankdejonge:
Looks good, if you fix the tests I can release it this week 👍

Greg0:
Finally I used `ftp_raw` instead of `ftp_exec` because of SITE commands isn't very widely supported so can cause compatibility issues on some servers. Pipelines for PHP 7.3 doesn't pass but I think it is not caused by my changes. It fail because deprecations in PHP I guess. 

Sorry for so many commits they should be stashed on merge. 

Waiting for your reply.

ikari7789:
@Greg0 my team has also been facing this issue with downloads taking several hours when it could take an hour or so. We tested this approach ourselves and found in some cases that the NOOP did not necessarily keep the connection alive and had to force kill and reopen the FTP connection on a disconnection.

PR #1096, opened 2019-11-15
Automated native_function_invocation fixes

draco2003:
Reproduce with:
php php-cs-fixer --rules=native_function_invocation fix ./ --allow-risky=yes
(php-cs-fixer from : https://cs.symfony.com/)

Details:
https://veewee.github.io/blog/optimizing-php-performance-by-fq-function-calls/

### DIFF
--- a/phpstan.php
+++ b/phpstan.php
@@ -6,14 +6,14 @@ if (PHP_VERSION_ID < 70100) {
 
 $phpstanURL = 'https://raw.githubusercontent.com/phpstan/phpstan-shim/0.11.8/phpstan.phar';
 
-if (file_exists('./phpstan.phar') === false) {
+if (\file_exists('./phpstan.phar') === false) {
     echo "phpstan.phar doesn't exist, downloading from $phpstanURL\n";
-    $phpstanFile = fopen($phpstanURL, 'rb');
-    file_put_contents('./phpstan.phar', $phpstanFile);
-    chmod('./phpstan.phar', 0755);
-    fclose($phpstanFile);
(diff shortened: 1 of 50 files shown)


frankdejonge:
Hi @draco2003, this is not possible because the library depends on the ability to hijack this to run tests and fixate responses. Filesystems are an ugly business. Also, these optimization makes sense for high number of calls in the same place, Filesystem usage is not one of the areas where this applies.

draco2003:
Hey @frankdejonge thanks for the quick reply. 
Sorry about the over aggressive additions of the \ 
In my tests it only optimized the ones that it was supposed to. 
Somewhere in my automation across PRs it is doing a lot more. 

It should have only done it for the functions that can be optimized.
(https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/master/src/Fixer/FunctionNotation/NativeFunctionInvocationFixer.php#L353)

Agree that filesystems (and really any I/O) is going to overshadow improvements if anything slows down. 

Also an alternative syntax (just wasn't part of the automation i used) is the: 
`use function is_array;`

Which has the same optimization, without the forward slash everywhere if that style would be preferred.

GrahamCampbell:
I don't think this really adds many benefits, and maybe is a breaking change actually, since it stops people from being able to mock out flystem's calls to global functions (which they shouldn't be doing, but might be!).

GrahamCampbell:
> this is not possible because the library depends on the ability to hijack this to run tests and fixate responses

Yeh, and indeed flysystem's own tests need this. ;)

frankdejonge:
Closing this since we need the hijacking behavior.

Issue #117, opened 2014-03-28
How ot move Directory to directory recurcive


ghost:
Hello,
how copie a full directory to another please ?

i have this 

dir/underDir1

and i want this to the final :+1: 

anotherDir/dir/underdir1

and of course move all files on the final directory ?
 a function like this :+1: 
moveDirectory($sourceDirectory,estinationDirectory);

what do u think please?

Spholt:
For anyone coming to this late, the answer is to use:

`Filesystem:rename('path/to/old', 'path/to/new'):`

The documents are not clear that this function can actually be used for directories as well as files. Any files within your directory will also be pulled across to the new location.

Hope this helps :)

PR #1097, opened 2019-11-21
Updating API doc to reflect directory operations

igorsantos07:
Given #117, at least since 2014 the docs were missing clearer information on whether it's possible to delete/rename/move directories as well as files.

### DIFF
--- a/docs/usage/filesystem-api.md
+++ b/docs/usage/filesystem-api.md
@@ -125,7 +125,7 @@ param         | description                                   | type
 
 ---
 
-## Check if a file exists
+## Check if a file or directory exists
 
 ```php
 $exists = $filesystem->has($path);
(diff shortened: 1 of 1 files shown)


docs/usage/filesystem-api.md
```diff
@@ -168,29 +168,29 @@ param         | description                                   | type
...
 param         | description                                   | type
 ------------- | --------------------------------------------- | -----------
-`$from`       | location of a file                            | `string`
+`$from`       | location of a file/directory                  | `string`
 `$to`         | new location                                  | `string`
 `$response`   | success boolean                               | `bool`
 
 ---
 
-## Copy Files
```
frankdejonge: We should add a note here that this is not consistent behaviour across adapters (for local it works, for AWS it doesn't for example.

cancan101:
Not sure if this got reverted / changed in new versions, but the current docs do not make clear if move can be used on directories: https://flysystem.thephpleague.com/docs/usage/filesystem-api/:
image
and if not are there any APIs to move / rename a directory?

PR #1102, opened 2019-12-04
Allowed some empty dirname in emulateObjectDirectories() method

ewake:
https://www.php.net/manual/en/function.empty.php#103756

### DIFF
--- a/src/Util.php
+++ b/src/Util.php
@@ -291,13 +291,13 @@ class Util
             $listedDirectories[] = $object['path'];
         }
 
-        if (empty($object['dirname']) && !is_numeric($object['dirname'])) {
+        if (!isset($object['dirname']) || trim($object['dirname']) === '') {
             return [$directories, $listedDirectories];
         }
 
         $parent = $object['dirname'];
 
-        while ( (! empty($parent) || is_numeric($parent)) && ! in_array($parent, $directories)) {
+        while (isset($parent) && trim($parent) !== '' && ! in_array($parent, $directories)) {
             $directories[] = $parent;
             $parent = static::dirname($parent);
         }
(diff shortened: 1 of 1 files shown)


src/Util.php
```diff
@@ -291,13 +291,13 @@ protected static function emulateObjectDirectories(array $object, array $directo
             $listedDirectories[] = $object['path'];
         }
 
-        if (empty($object['dirname'])) {
+        if (empty($object['dirname']) && !is_numeric($object['dirname'])) {
```
GrahamCampbell: What's wrong with a numeric directory name?

GrahamCampbell: Oh, I see. Maybe just change the whole thing to `if (isset($object['dirname']) && $object['dirname'] !== null) {`. Is much clearer what the intent is then.

ewake:
Hi there, i encountered problems when `$object['dirname']` is `'0'`, and I used this solution https://www.php.net/manual/en/function.empty.php#103756 .
I change it with your advice if you think it is better..

frankdejonge:
Hi @ewake, thank you for your contribution. The assertion you're making check if for a known thing to exclude, rather than checking if the value is "wrong". Could you update it if it's set and if it's not an empty string. That's basically what we want to guard against. I can do it myself to, but if you like to contribute that's really welcome as well 👍

ewake:
> Hi @ewake, thank you for your contribution. The assertion you're making check if for a known thing to exclude, rather than checking if the value is "wrong". Could you update it if it's set and if it's not an empty string. That's basically what we want to guard against. I can do it myself to, but if you like to contribute that's really welcome as well

Hi Frank, I updated the pull request and I hope I understood what you mean..
Otherwise my contribution will be limited to reporting the problem, that I hope someone can solve better than me :wink:

frankdejonge:
Released as 1.0.59

PR #1106, opened 2019-12-08
Fixed another PHP 7.4 bug

GrahamCampbell:
The `$metadata` may be `false`, in which case, we should probably treat it as a `dir`, the same as what `get` in `Filesystem` does.


frankdejonge:
@GrahamCampbell please ping when you're done and I'll make sure it'll be tagged.

GrahamCampbell:
This is the only other case of a bad array access to metadata that I could see. :)

Issue #1113, opened 2019-12-16
Deprecation notice when parsing permission value 33188


DuckThom:
### Bug Report

|    Q        |   A
|------------ | ------
| BC Break    | not sure
| Version     | 1.0.61

#### Summary

Parsing a permission value of 33188 (octal 0644) (as returned by `phpseclib/phpseclib`) by `\League\Flysystem\Adapter\AbstractFtpAdapter::normalizePermissions`, as of PHP 7.4, passing this value to `octdec()` will result in a `Deprecated: Invalid characters passed for attempted conversion, these have been ignored` notice.

After reading https://github.com/phpseclib/phpseclib/issues/1329 I feel like octdec is not the right function for this returned number and looking at the `fileperms` function from PHP (https://www.php.net/manual/en/function.fileperms.php), converting this number to it's octal representation can be done with `substr(sprintf('%o', fileperms($filename)), -4);`.

As tested on my server:

```
$ ls -l 18028852_10240_04122018_144957.PDF
-rw-r--r-- 1 ftp ftp 64583 Dec  4  2018 18028852_10240_04122018_144957.PDF

$ php -a
Interactive mode enabled

php > echo fileperms('18028852_10240_04122018_144957.PDF');
33188
php > echo substr(sprintf('%o', fileperms('18028852_10240_04122018_144957.PDF')), -4);
0644
```

#### How to reproduce

Have a file with `-rw-r--r--` permissions on a remote server and let it go through the `normalizePermissions` method.

DuckThom:
I've done some more testing and it looks like the `normalizePermissions` method is receiving the value `33188` instead of its string representation `-rw-r--r--`, I think I got something working locally, I'll make a PR for reviewing

PR #1114, opened 2019-12-16
Fix deprecation message

DuckThom:
The `normalizePermissions` method can receive permissions in their string format (ex. `-rw-r--r--`) but also in decimal format `33188`. This change will add support for the decimal format as previously it would be first truncated to 3188, which would then be interpreted as octal number and converted to decimal. After PHP 7.4, parsing a non-octal number with `octdec` will make PHP emit a Deprecation notice. 

Fixes: #1113

### DIFF
--- a/src/Adapter/AbstractFtpAdapter.php
+++ b/src/Adapter/AbstractFtpAdapter.php
@@ -557,6 +557,10 @@ abstract class AbstractFtpAdapter extends AbstractAdapter
      */
     protected function normalizePermissions($permissions)
     {
+        if (is_numeric($permissions)) {
+            return $permissions & 0777;
+        }
+
         // remove the type identifier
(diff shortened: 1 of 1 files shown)


frankdejonge:
@DuckThom what kind of server are you talking against? Do you know the distro and the sftp flavour?

DuckThom:
The sftp server is an openssh server with internal-sftp as subsystem.

```OpenSSH_7.4p1 Debian-10+deb9u7, OpenSSL 1.0.2t  10 Sep 2019 ```

The permission value 33188 is coming from here https://github.com/phpseclib/phpseclib/blob/63e4c9a4a9ccba7d695a577994e14ef8a1cc8f2b/phpseclib/Net/SFTP.php#L2817 

As a side note, PHPs `fileperms` function also returns permissions in this decimal format

PR #1116, opened 2019-12-20
improved coerency between interfaces

mabuonomo:
Hi, I've modified two interfaces:

- FilesystemInterface need to expose the getAdapter() method

- AdapterInterface need to expose the methods included in the AbstractAdapter. All the adapter extends  the AbstractAdapter, but the getAdapter() of Filesystem class return only an AdapterInterface without several important methods like getPathPrefix()

### DIFF
--- a/src/FilesystemInterface.php
+++ b/src/FilesystemInterface.php
@@ -281,4 +281,11 @@ interface FilesystemInterface
      * @return $this
      */
     public function addPlugin(PluginInterface $plugin);
+
+    /**
+     * Get the Adapter.
+     *
+     * @return AdapterInterface adapter
+     */
+    public function getAdapter();
 }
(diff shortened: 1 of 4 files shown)


yellow1912:
I like this PR, it's just weird that those methods are there and public but not shown when you use IDEs.

GrahamCampbell:
This would be a major breaking change, however.

src/FilesystemInterface.php
```diff
@@ -281,4 +281,11 @@ public function get($path, Handler $handler = null);
      * @return $this
      */
     public function addPlugin(PluginInterface $plugin);
+
+    /**
+     * Get the Adapter.
+     *
+     * @return AdapterInterface adapter
+     */
+    public function getAdapter();
```
GrahamCampbell: 👎 The flysystem class should not expose its adapter.

mabuonomo: why? without this method, i need to cast the FilesystemInterface into a Filesystem object

frankdejonge:
I disagree with this position. Interfaces are for consumption. Internals should be exposed via DI if needed, not extracted from their consumers. In v2 the majority of these methods will be removed. In the future please discuss first, before sending PR’s.

PR #1119, opened 2019-12-29
Robust getStreamSize

GrahamCampbell:


frankdejonge:
@GrahamCampbell please change it to return `null` instead of throwing an exception and add this test case for the Util class:

```php
    public function testStreamSizeForUrl()
    {
        $stream = \fopen('https://flysystem.thephpleague.com/img/flysystem.svg', 'r');
        $this->assertNull(Util::getStreamSize($stream));
        fclose($stream);
    }
```

GrahamCampbell:
Sure. For v2's design, I wonder how you would think about using the `Maybe` pattern, for robustness. Since, people probably calling this code will end up treating null as 0.

GrahamCampbell:
https://github.com/schmittjoh/php-option is a long standing PHP implementation of this, used by phpdotenv under the hood.

frankdejonge:
@GrahamCampbell for 2.0 the behaviour will probably be to omit any size optimisations as well, but the interface won't be public like now.

frankdejonge:
Released as 1.0.62 (this PR build fails because of the missing branch), master is green.

PR #1122, opened 2020-01-04
Fixed prefer lowest tests on PHP 8.0

GrahamCampbell:
(I hope... let's see if travis agrees)


GrahamCampbell:
Worked. Just warnings now.

PR #1129, opened 2020-02-12
Fixed deletion iteration bug in Local::deleteDir()

modevelops:
When removing files during an RecursiveIteratorIterator iteration, the iteration is perturbed, producing incorrect results (skipping files). Storing the files in an array before and iterating the array afterwards solves the problem.

Similar problems reported here: 
https://github.com/emscripten-core/emscripten/issues/2528
https://forge.typo3.org/issues/88524

### DIFF
--- a/src/Adapter/Local.php
+++ b/src/Adapter/Local.php
@@ -410,11 +410,16 @@ class Local extends AbstractAdapter
 
         /** @var SplFileInfo $file */
         foreach ($contents as $file) {
+            $files[] = $file;
+        }
+    
+        /** @var SplFileInfo $file */
+        foreach ($files as $file) {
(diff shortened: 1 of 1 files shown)


modevelops:
Hello @frankdejonge,

is my pull request denied? I am developer at Gambio (all-in-one ecommerce solution) and this bug occurs on several servers of our customers using our software with flysystem. It would be great, if this bug will be solved so we don't have to patch it ourselves. If you need more information feel free to ask.

Hibbelharry:
Hi @frankdejonge,
can you please comment on this matter?

frankdejonge:
@Hibbelharry I'll get to it when I get to it.

PR #1131, opened 2020-02-16
Doc suggestions

driesvints:
Here's just a few suggestions. I removed the backticks in the title because those looked a little bit weird.

### DIFF
--- a/v2/advanced/upgrade-to-2.0.0.md
+++ b/v2/advanced/upgrade-to-2.0.0.md
@@ -12,8 +12,8 @@ were removed.
 
 ### Upgrade your dependencies
 
-Firstly you'll need to upgrade Flysystem itself. You can do this by requiring `~2.0`
-instead of  `~1.0`. The same needs to be done for all the adapter too, in the same action.
+Firstly you'll need to upgrade Flysystem itself. You can do this by requiring `^2.0`
+instead of  `^1.0`. The same needs to be done for all the adapter too, in the same action.
 
 ## Removed functionality.
 
--- a/v2/usage/directory-listings.md
+++ b/v2/usage/directory-listings.md
@@ -39,7 +39,7 @@ $allPaths = $filesystem->listContents('/some/path')
     ->toArray();
 ```
 
-# Storage attributes
+## Storage attributes
 
 Directory listings contain storage attributes, objects that expose information about
 the items contained in a (part of a) filesystem.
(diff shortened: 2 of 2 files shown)


frankdejonge:
@driesvints what's the reason for the `~` => `^` change? From what I get from the composer docs, the `~` is just fine, or am I missing something?

GrahamCampbell:
It tells users that flysystem is following semver.

frankdejonge:
@GrahamCampbell the other one is semver too though? I'm confused.

frankdejonge:
https://getcomposer.org/doc/articles/versions.md#tilde-version-range- is even states it's the most useful for semver projects. The `^` seems more for people who want to get at least a certain patch version as well, so for an x.x.x version, not a x.x one.

GrahamCampbell:
Only `^` is "the semver operator". For example, `~0.4` or `~4.2.1` will not have the effect of "give me the latest version compatible with the API at 0.4 or 4.2.1, respectively".

frankdejonge:
Are you sure, @GrahamCampbell, I read the docs in a different way.

from the docs:

The ~ operator is best explained by example: ~1.2 is equivalent to >=1.2 <2.0.0, while ~1.2.3 is equivalent to >=1.2.3 <1.3.0. As you can see it is mostly useful for projects respecting semantic versioning. A common usage would be to mark the minimum minor version you depend on, like ~1.2 (which allows anything up to, but not including, 2.0). Since in theory there should be no backwards compatibility breaks until 2.0, that works well. Another way of looking at it is that using ~ specifies a minimum version, but allows the last digit specified to go up.

GrahamCampbell:
Yes, I am sure. For the most clarity, look at their tests. ;)

GrahamCampbell:
My first example was 0.x range on purpose, since the behaviour is the same on 1.x.

frankdejonge:
That shit is not clearly explained in the composer docs then ... hmm.

GrahamCampbell:
I think it is in the docs? It says:

> For pre-1.0 versions it also acts with safety in mind and treats ^0.3 as >=0.3.0 <0.4.0.

GrahamCampbell:
Where as `~` does nothing special, so will let you have 0.5.0, given `~0.3`.

frankdejonge:
Gotcha, so for this upgrade and further on it doesn't matter. Thanks for the clarification anyway 👍

driesvints:
@frankdejonge part because of the reasons stated by Graham above but mainly because it's the more commonly used convention.

PR #1134, opened 2020-02-24
catch error exceptions on reconnect

kohlerdominik:
Hello there

We found an issue with the FTP-adapter with SSL-encrypted connection.

**Our use-case**
we run workers which download files from an FTP-Server periodically. There are several list-files and downloads per minute. The worker keeps its connection up after start until stop/restart, which can be between several days and several months.

But very occasionally, we se in our error log `ftp_close(): SSL read failed `. We assume this happens when the SSL-connection times out after some days, because it’s never to be seen for at least some hours after restart.

**Our fix**
For some days now we use the fix now provided in our PR and don’t see this issue anymore, which confirms above assumption.

We add error handling to the `getConnection()`-method. In our opinion, it doesn’t make sense that this method allows several tries to connect, but doesn’t catch an exception when disconnect fails, especially due connection loss.

### DIFF
--- a/src/Adapter/AbstractFtpAdapter.php
+++ b/src/Adapter/AbstractFtpAdapter.php
@@ -643,7 +643,19 @@ abstract class AbstractFtpAdapter extends AbstractAdapter
 
         while ( ! $this->isConnected() && $tries < 3) {
             $tries++;
-            $this->disconnect();
+
+            // when there was an unexpected connection loss (especially some
+            // SSL errors), disconnect will throw an ErrorException.
+            // But we try to reconnect here, so it's better to catch this error
+            // once and allow a retry nonetheless.
+            try {
+                $this->disconnect();
+            } catch (\ErrorException $e) {
+                if ($tries > 1) {
+                    throw $e;
+                }
+            }
+
             $this->connect();
         }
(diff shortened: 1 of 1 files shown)


frankdejonge:
Can you check latest master, the errors are suppressed there, which should prevent this. It seems like you're using some mechanism to convert warnings into exceptions, is that correct?

kohlerdominik:
Hi Frank

Thanks for your response. I will update to the fixed Version today and provide feedback in a couple of days.

Actually, I didn't even notice that it’s a warning originally. We use Laravel Framework, which does what you suspected: It converts the warning to an Exception ([source](https://github.com/laravel/framework/blob/7.x/src/Illuminate/Foundation/Bootstrap/HandleExceptions.php#L35-L52)).

kohlerdominik:
Good Morning Frank

After several days in Production I can confirm that your fix for suppressing warnings works in our environment. Stability is unaffected and our error log is clean now. Therefore i can reccomend releasing this change.

Thank you for the quick solution of our issue.

frankdejonge:
@kohlerdominik 1.0.66 was just tagged. I'll close this issue.

Issue #1140, opened 2020-03-08
\League\Flysystem\Adapter\Local::copy does not copy directories


flangfeldt:
### Bug Report

|    Q        |   A
|------------ | ------
| Version     | 1.0.63
| PHP version | 7.3.13
| Server OS | Centos 7.7.1908

#### Summary

\League\Flysystem\Adapter\Local::copy does not copy directories. The description of the API states it copies Files and Directories, but with the Local Adapter only Files are copied. See https://flysystem.thephpleague.com/v1/docs/usage/filesystem-api/#copy-files-or-directories.

#### How to reproduce

```
$fromPath = '1';
$toPath = '2';

$handle = new Filesystem(
    new Local(
        "/opt/test/",
        LOCK_EX,
        Local::DISALLOW_LINKS,
        [
            "file" => [
                "public" => 0777,
                "private" => 0700
            ],
            "dir" => [
                "public" => 0777,
                "private" => 0700,
            ]
        ]
    )
);

if (is_object($handle)) {
    try {
        if ($handle->has($fromPath)) {
            $handle->copy($fromPath, $toPath);
        }
    } catch (\Exception $ex) {
        echo $ex->getMessage();
    }
}
```

This results in the error message:

``` 
copy(): The first argument to copy() function cannot be a directory
```

frankdejonge:
This was an error in the docs, I've corrected it.

flangfeldt:
Ok, that's a shame. Seems like a great addition. 
Now I have to Frankenstein it outside of the Adapter. 

Thanks for the super fast response anyway.

PR #1141, opened 2020-03-08
Add directory support to the Local::copy method

flangfeldt:
This is solving the missing directory support for the Local::copy method as outlined in #1140.


frankdejonge:
@flangfeldt hi, this behaviour is not compatible with the other adapters. Since the main promise is portability, this cannot be supported like this. I've corrected the documentation.

PR #1144, opened 2020-03-14
Use iterable instead of generator.

frankdejonge:


src/DirectoryListing.php
```diff
@@ -62,7 +63,9 @@ public function getIterator(): Generator
      */
     public function toArray(): array
     {
-        return iterator_to_array($this->listing, false);
+        return $this->listing instanceof Traversable
```
GrahamCampbell, line 66: Is this the same as `!is_array($this->lising)`. Can something ever be iterable and not an array or a traversable?

frankdejonge: For static analysis this is the clearer setup. The `iterable` pseudo type is just a little weird. The `!is_array` does pretty much the same but casting to array is easy. I think there's pretty much no difference between either implementation, right?

PR #1145, opened 2020-03-18
Revert "ignore test-functions.php"

jderusse:
Methods in this file are used by FilesystemAdapterTestCase (which is not ignored)

This changed prevent 3rd to use the AbstractAdapter for testing their adapter.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -9,7 +9,6 @@ src/PHPSecLibV2 export-ignore
 .gitignore export-ignore
 build/ export-ignore
 mocked-functions.php export-ignore
-test-functions.php export-ignore
 test_files/ export-ignore
 **/*Test.php export-ignore
 **/*Stub.php export-ignore


frankdejonge:
Hi @jderusse, instead of reverting this, I've created an additional package which includes the test utilities. Can you checkout that package and see if it works for you? https://github.com/thephpleague/flysystem-adapter-test-utilities (it's a -dev release now so mind that when you're requiring it).

frankdejonge:
I'm spotting a incorrect test file location, fixing it now.

jderusse:
It works thank => https://github.com/async-aws/aws/pull/323

But maintaing the same adapter for both 1.x and 2.x branches is not trivial. Do you plan to release a 1.x branch of `thephpleague/flysystem-adapter-test-utilities` that don't requires `league/flysystem": "2.0.0"`?

frankdejonge:
@jderusse what problem are you running into? The test utils are only needed when integrating with Flysystem 2.x and it depends on the internals of that package. If you want to support two versions of Flysystem, I suggest either releasing them in separate branches or repositories. This is how I solved this issue for the normal AWS SDK in the past (version 2 and 3) and the soon to be released pphpseclib3 will follow the same pattern.

Issue #1151, opened 2020-04-03
Documentation error 


TerraSkye:
When i was browsing thourght the upgrade docs i fount out that 
under the heading *Writes are now deterministic*
the new methods  are called `write` and `writeStream`
yet the example below that says `writeWrite`

https://flysystem.thephpleague.com/v2/docs/advanced/upgrade-to-2.0.0/

Lamicz:
I have a small error in this block:

- $fileExists = $filesystem->has($path); 
+ $fileExists = $filesystem->fileExist($path);

The new method is called "fileExists"

frankdejonge:
Thanks, I fixed it.

PR #1152, opened 2020-04-03
#1151  upgrade docs fix

TerraSkye:
updated the documents to reflect the correct method name


frankdejonge:
Thank you @TerraSkye 👍

PR #1153, opened 2020-04-15
Added markdown MIME Type

n123456789k:
added `.md` as `text/markdown`


n123456789k:
RFC for MIME Type: https://tools.ietf.org/html/rfc7763

colinodell:
As the maintainer of league/commonmark, I support this addition :wink:

OskarStark:
What about adding `*.markdown`, too?

src/Util/MimeType.php
```diff
@@ -102,6 +102,7 @@ class MimeType
         'txt' => 'text/plain',
         'text' => 'text/plain',
         'log' => 'text/plain',
+        'md' => 'text/markdown',
```
OskarStark, line 106:
```suggestion
        'markdown' => 'text/markdown',
        'md' => 'text/markdown',        
```

n123456789k:
Added `*.markdown` @OskarStark

GrahamCampbell:
NB This is already available in flysystem v2 which uses https://github.com/thephpleague/mime-type-detection/blob/master/src/GeneratedExtensionToMimeTypeMap.php.

n123456789k:
@GrahamCampbell Can this be merged as a fix to the v1 stable rather than waiting on v2 release.

Also, no idea why it's failing since I've added the other `.markdown` 🤷

GrahamCampbell:
> Can this be merged as a fix to the v1 stable rather than waiting on v2 release.

GrahamCampbell:
> Can this be merged as a fix to the v1 stable rather than waiting on v2 release.

I would have thought so.

> Also, no idea why it's failing since I've added the other .markdown 🤷

It's because of the trailing whitespace you added:

image

n123456789k:
@GrahamCampbell Whitespace removed 👍

frankdejonge:
Thanks @narwy 👍

frankdejonge:
Released as 1.0.67

PR #1154, opened 2020-04-16
Add own __call magic method of Filesystem class

samuelhgf:
This PR implements the possibilite of call a method available in the adapter that is not available in the Filesystem. 

Example: Let's say a want to call the method getThumbnail in DropboxFilesystem. Because it is not in the Filesystem interface it will throw a BadMethodCallException. But I can call the method getAdapter() to get the custom adapter and then call the method that is only in the CustomAdapter.

It is a bit confusing but it worth it.

If any thing is wrong, even the whole PR, please trying yo implement that feature.

Thanks!

### DIFF
--- a/src/Filesystem.php
+++ b/src/Filesystem.php
@@ -2,6 +2,7 @@
 
 namespace League\Flysystem;
 
+use BadMethodCallException;
 use InvalidArgumentException;
 use League\Flysystem\Adapter\CanOverwriteFiles;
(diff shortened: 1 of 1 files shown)


frankdejonge:
Hi, if you need to access the adapter, just access the adapter. The filesystem class is not a transparent proxy for calls. If you require functionality that are vendor-specific, it's also better just to not use Flysystem for this. Depending on the actual explicit dependencies is always better in those cases.

PR #1163, opened 2020-05-12
Added ico => image/x-icon mime type

adamhopkinson:
Needed for some code I'm working on - all other image types I need are represented, but not ico files.


GrahamCampbell:
This should be added in the correct position, alphabetically, rather than being dumped at the end.

adamhopkinson:
@GrahamCampbell I did consider this, but the array is neither order by key nor value so I assumed my addition should go at the end.

frankdejonge:
Thanks @adamhopkinson, it's released as 1.0.68.

adamhopkinson:
Thanks @frankdejonge

PR #1171, opened 2020-06-04
Fix name of AwsS3V3Adapter in documentation 

sebo:
It seems that the adapter name on AwsS3V3 is different than described in the documentation. See here https://github.com/thephpleague/flysystem-aws-s3-v3/blob/2.0.0-alpha.2/AwsS3V3Filesystem.php. Led to some confusion on my side. Might be helpful to change it in the documentation then.


frankdejonge:
Thank you 👍

PR #1183, opened 2020-08-08
Version: 1.1.0

frankdejonge:


composer.json
```diff
@@ -21,12 +21,13 @@
         }
     ],
     "require": {
-        "php": ">=5.5.9",
-        "ext-fileinfo": "*"
+        "php": ">=7.2.5",
```
GrahamCampbell: Maybe go with `^7.2.5 || ^8.0`?

tests/ForcedCopyPluginTests.php
```diff
@@ -5,12 +5,12 @@
 
 class ForcedCopyPluginTests extends TestCase
 {
-    use \PHPUnitHacks;
+    // use \PHPUnitHacks;
```
GrahamCampbell: These should probably be deleted from the code. They remain in the git history, if we wanna get them back? ;)

tests/FtpTests.php
```diff
@@ -805,40 +810,44 @@ public function testRawlistFail()
...
         $adapter->connect();
     }
 
     /**
      * @depends testInstantiable
-     * @expectedException RuntimeException
      */
     public function testLoginFailSsl()
     {
+        if (substr(PHP_VERSION, 0, 2) === '8.') {
```
GrahamCampbell: `PHP_MAJOR_VERSION > 7`?

GrahamCampbell: Can also be implemented with a `@requires PHP < 8` annotation.

tests/FtpTests.php
```diff
@@ -866,20 +875,20 @@ public function testItCanSetSystemType()
...
         $adapter->listContents();
     }
 
     /**
      * @depends testInstantiable
-     * @expectedException \RuntimeException
      */
     public function testItThrowsAnExceptionWhenAnInvalidUnixListingIsFound()
     {
+        $this->expectException(RuntimeException::class);
```
GrahamCampbell, line 890: Why is there an assert equals on the last line. Is that not reachable?

GrahamCampbell: Also, the reason PHPUnit introduced expectException and ditched the annotation was so that people could put this just above the line of code they expected to crash, rather than at the start of the whole test method. The idea being then that the code cannot crash earlier than expected, and also unreachable code is meant to be less likely to be written in error. ;)

GrahamCampbell (changes requested):
Left some minor comments. :)

tests/FtpTests.php
frankdejonge: This was just a super old "copy-paste-adjust" from a PR, about 4 years ago.

tests/FtpTests.php
frankdejonge: Adjusted 👍

frankdejonge:
@GrahamCampbell thanks for the review 👍 much appreciated. I've pushed changed to handle them.

composer.json
```diff
@@ -21,12 +21,13 @@
         }
     ],
     "require": {
-        "php": ">=5.5.9",
-        "ext-fileinfo": "*"
+        "php": "^7.2.5 || ^8.0",
+        "ext-fileinfo": "*",
+        "league/mime-type-detection": "^1.3"
```
GrahamCampbell, line 26: https://github.com/thephpleague/mime-type-detection/pull/1 will be needed so that `league/mime-type-detection` doesn't block installation on PHP 8. :)

GrahamCampbell (changes requested):

composer.json
frankdejonge: I've merged your PR and tagged 1.4.0, so we're good 👍

src/Util/MimeType.php
```diff
@@ -2,194 +2,35 @@
...
 namespace League\Flysystem\Util;
 
-use ErrorException;
-use finfo;
+use League\MimeTypeDetection\FinfoMimeTypeDetector;
+use League\MimeTypeDetection\GeneratedExtensionToMimeTypeMap;
+use League\MimeTypeDetection\MimeTypeDetector;
 
 /**
  * @internal
```
GrahamCampbell, line 10: This class is internal, so nobody can call the useDetector stuff?

frankdejonge: People can use it if they need to, however, it falls outside of the BC promise, so it's at their own risk.

frankdejonge: It's all going to be different in V2 anyway, where this is just a normal injected dependency, not this global scope iffiness.

GrahamCampbell: Fair enough. :)

GrahamCampbell (approved):
🚢

PR #1221, opened 2020-11-03
Tighten visibility typehint, always string.

frankdejonge:


Nyholm (approved):
I agree with this. It is better to be strict with the types

PR #1222, opened 2020-11-03
Add information about invalid filesystem when using MountManager.

armetiz:


Nyholm (approved):
I like this. 

Thank you

PR #1223, opened 2020-11-03
Adding AsyncAws adapter

Nyholm:
This will fix #1220

PR #1224, opened 2020-11-03
Adding documentation for AsyncAws adapter

Nyholm:
Related to #1223

### DIFF
--- a/_data/menu_v2.yml
+++ b/_data/menu_v2.yml
@@ -19,6 +19,7 @@ Usage:
 Adapters:
     Local: '/v2/docs/adapter/local/'
     AWS S3: '/v2/docs/adapter/aws-s3-v3/'
+    AsyncAws S3: '/v2/docs/adapter/async-aws-s3/'
     FTP: '/v2/docs/adapter/ftp/'
     SFTP: '/v2/docs/adapter/sftp/'
     InMemory: '/v2/docs/adapter/in-memory/'
--- a/v2/adapter/async-aws-s3.md
+++ b/v2/adapter/async-aws-s3.md
@@ -1,0 +1,65 @@
+---
+layout: default
+title: AsyncAws S3 Adapter
+permalink: /v2/docs/adapter/async-aws-s3/
+---
+
+Interacting with Aws S3 through Flysystem can be done
+by using the `League\Flysystem\AsyncAwsS3\AsyncAwsS3Adapter`.
+
+Read more about AsyncAws's S3Client in [their documentation](https://async-aws.com/clients/s3.html).
+
+## Simple usage:
+
+```php
+$client = new AsyncAws\S3\S3Client();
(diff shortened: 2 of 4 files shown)


frankdejonge (approved):

_data/menu_v2.yml
```diff
@@ -19,6 +19,7 @@ Usage:
 Adapters:
     Local: '/v2/docs/adapter/local/'
     AWS S3: '/v2/docs/adapter/aws-s3-v3/'
+    AsyncAws S3: '/v2/docs/adapter/async-aws-s3/'
```
OskarStark, line 22: I know the class is named `AsyncAws`, but the flysystem adapter for S3 is named `AWS`, so I propose to rename the adapter name to:

```suggestion
    AsyncAWS S3: '/v2/docs/adapter/async-aws-s3/'
```

v2/adapter/async-aws-s3.md
```diff
@@ -0,0 +1,65 @@
...
+    )
+);
+
+// The FilesystemOperator
+$filesystem = new League\Flysystem\Filesystem($adapter);
+```
+
+## Support for large files:
+
+If you want to upload files larger than 5GB you need to use the `SimpleS3Client`.
```
OskarStark, line 55: When reading this sentence, my question would be _why_? What does the Async adapter make different?

Nyholm: Thank you. 

The `SimpleS3Adapter` is taking care of the complexity of multipart upload. That logic is not in the `S3Client` and should not be in the FlysystemAdapter. 

One could suggest that we should always use the SimpleS3 adapter. However, I think uploading 5GB files is very uncommon and I would rather save people the "trouble" to download the `SimpleS3Client` when they dont need to. 

-----

The SimpleS3Client has two features: 
1. It supports large file uploads out of the box. 
2. It creates a way easier interface for the user. `upload()`, `getUrl()`, `download()` etc. That is good for people that is interacting with S3 directly, ie not using Flysystem.

_data/menu_v2.yml
Nyholm: It should be "AsyncAws" or "async-aws" depending on the context. That is something we decided just because we think it looks better. 

I know it differs from "AWS".

v2/adapter/async-aws-s3.md
OskarStark: Great explanation, thank you. Maybe we could extend the sentence a bit to make the "why" more clear

Nyholm: Thank you

OskarStark: great, thanks

OskarStark (approved):

Nyholm:
Thank you for merging.

PR #1225, opened 2020-11-03
Add support for PHPUnit 9

Nyholm:
PHPUnit 9 is the only version with PHP8 support.


Nyholm:
Thank you for merging

PR #1227, opened 2020-11-05
Add PHP 8 support

ElGigi:
Add support of PHP 8 in `composer.json` file.
Only one problem : `friendsofphp/php-cs-fixer` package is not compatible with PHP 8


Nyholm:
We are already testing on php8. 

I’m 👍

PR #1234, opened 2020-11-20
Fix private dir permission default for fromArray call

glensc:
Perhaps the defaults should be private class constants, to avoid sync copy-paste error again.


glensc:
Not sure, the FTP checks seem to be flaky, and not caused by this PR changes.

On the first push it failed with 7.4 (7.2, 7.3, 8.0 are ok):
- https://github.com/thephpleague/flysystem/actions/runs/374013248

On the second push it failed with 7.2 (and 7.3, 7.4 are ok, 8.0 cancelled):
- https://github.com/thephpleague/flysystem/actions/runs/374104763

PR #1235, opened 2020-11-20
Migrate phpunit XML configuration

glensc:
Running PHPUnit, emitted a notice:

```
Warning:       Your XML configuration validates against a deprecated schema.
Suggestion:    Migrate your XML configuration using "--migrate-configuration"!
```

Converted with:
```
$ ./vendor/bin/phpunit --migrate-configuration
```

With a manual adjustment to keep existing indenation.

### DIFF
--- a/.github/workflows/quality-assurance.yml
+++ b/.github/workflows/quality-assurance.yml
@@ -4,12 +4,18 @@ on:
   push:
     paths:
       - src/**/*.php
+      - phpunit.xml.dist
+      - phpunit.php
+      - mocked-functions.php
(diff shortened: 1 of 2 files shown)


OskarStark (approved):

frankdejonge:
Hi @glensc,

The lower end for PHPUnit is still 8.5, which has the inverse effect of this. Either way it won't be perfect. Since I expect backwards compatibility will be far more likely than forward compatibility, I'll leave the configuration as is for now. When we increase the minimum PHP version (along with the frameworks) then we can make the migration.

PR #1240, opened 2020-12-02
Fix composer require command README

Iandenh:
The better version of:
thephpleague/flysystem-aws-s3-v3#233


frankdejonge:
Thank you @Iandenh 👍

PR #1254, opened 2021-01-11
Added details of installing the in memory adapter

toby-griffiths:
The v1 docs included a note on how to install, but was missing here, so added.


Nyholm (changes requested):
Thank you for this PR. 

I think we should add it on all adapters or no adapters. Just adding in for the InMemory Adapter makes no sense. 

Could you update the PR please?

frankdejonge:
I've merged this, PR's for other pages are very welcome.

PR #1259, opened 2021-01-18
Fix time-related tests failing in 2021

rgson:
Two tests related to Unix file listings are failing now in 2021 due to
the expectation that a timestamp without an explicit year is from 2020.

The expectation is adjusted to always use the current year. This
solution was already used for a different file in the same dataset.

### DIFF
--- a/tests/FtpTests.php
+++ b/tests/FtpTests.php
@@ -662,7 +662,7 @@ class FtpTests extends TestCase
                     [
                         'type' => 'dir',
                         'path' => 'folder',
-                        'timestamp' => 1606226340,
+                        'timestamp' => DateTime::createFromFormat('M d H:i', 'Nov 24 13:59')->getTimestamp(),
                     ],
                     [
                         'type' => 'file',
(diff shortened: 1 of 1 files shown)


BackEndTea (approved):

frankdejonge:
Thank you, @rgson 👍

SeniorSupportEngineer:
I don't think this is the correct solution. 
Because this year we had a issue that a file from 21 dec 2021, got a timestamp of 21 dec 2022. 
Because of the normalizeUnixTimestamp gets a time and not the year. Then your codes adds the $year = date('Y');
Now I need to make my own fix for this.

frankdejonge:
@SeniorSupportEngineer the FTP implementations do not supply the year, so it's not information we can use.

SeniorSupportEngineer:
@frankdejonge But your code makes the assumption that then the year is now by doing date('Y'); Now I need to undo that in someway, don't know how yet.

frankdejonge:
@SeniorSupportEngineer it's that or not supporting any type of timestamp. Choices were made, you can always choose not to trust the input, but flysystem cannot get better information about this, unfortunately.

PR #1262, opened 2021-01-26
Use $config parameter in copy operation

lboynton:
I noticed that when copying a file, the `$config` param is not used. This is a problem if you require the object to be encrypted, for example. This fixes the problem by merging the config with the config that the class is instantiated with.

### DIFF
--- a/src/AwsS3V3/AwsS3V3Adapter.php
+++ b/src/AwsS3V3/AwsS3V3Adapter.php
@@ -393,6 +393,12 @@ class AwsS3V3Adapter implements FilesystemAdapter
             );
         }
 
+        $options = $this->createOptionsFromConfig($config);
+
+        if ($metadataDirective = $config->get('MetadataDirective')) {
+            $options['MetadataDirective'] = $metadataDirective;
+        }
+
         try {
             $this->client->copy(
                 $this->bucket,
@@ -400,7 +406,7 @@ class AwsS3V3Adapter implements FilesystemAdapter
                 $this->bucket,
                 $this->prefixer->prefixPath($destination),
                 $this->visibility->visibilityToAcl($visibility),
-                $this->options
+                $options
             );
         } catch (Throwable $exception) {
             throw UnableToCopyFile::fromLocationTo($source, $destination, $exception);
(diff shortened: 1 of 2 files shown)


Nyholm (changes requested):
Thank you for this PR. 

Could you add a test too please?

lboynton:
> Thank you for this PR.
> 
> Could you add a test too please?

I have added a test, how do I run AWS tests? It doesn't look like Github actions runs any.

Nyholm:
The tests are running. See https://github.com/thephpleague/flysystem/pull/1262/checks?check_run_id=1810393450#step:13:1

To run the Aws tests locally you need to define some environment variables: https://github.com/thephpleague/flysystem/blob/317a3951ffca189da1c9774c32567d1973619fef/src/AwsS3V3/AwsS3V3AdapterTest.php#L80-L83

Nyholm:
> This is a problem if you require the object to be encrypted, for example.

I read form the tests, that you want to copy a file and add encryption at the same time? Is that really correct? The end result of `copy()` will not be a copy of the original file.

lboynton:
Originally yes, because I want to use a bucket policy that enforces encryption. However in the test I have used an example of modifying the content type of a file through a move operation, as this is easier to make assertions on in the test code.

Nyholm:
What I'm wondering is, could you download the file from s3 with no encryption, get the stream a d then with a new Filesystem object (with encryption) upload the stream. 

The copy function is for cloning a file. If you modify the properties (or content in the case of encryption) it is not a clone anymore.

lboynton:
Yes I could. But this change just exposes functionality that the AWS S3 CLI provides itself: https://docs.aws.amazon.com/cli/latest/reference/s3/cp.html

Happy to close this as a "won't fix" if this is not desired, but it seems logical to me to provide the same functionality. If this is preferred, the `$config` parameter should probably be removed since it's currently unused and would stop people like me from trying to set it.

frankdejonge:
@lboynton Hi,

I'm open to this PR but I would like to make one change, the explicit `if` statement + option setting is not needed as there is already a mechanism to include items from the config. Instead, please add `'MetadataDirective'` to the AVAILABLE_OPTIONS constant, this will make sure the option is copied over from the config object to the options array.

frankdejonge:
I'll amend the changes myself, about to tag a new release so this can be shipped as well then.

lboynton:
Thanks @frankdejonge 

Just to explain my reasoning for doing it that way - I believe `MetadataDirective` is only valid when doing certain actions like copy? So may not be valid for all S3 API operations. But I only looked at it briefly, so could be wrong.

Issue #1250, opened 2021-01-04
MountManager for v2


EvgenyBarinov:
Could you make MountManager more flexible for v2?
Proposed ways:
1. Convert private methods to protected to allow extending and modification
2. Optimize MountManager
- add public methods to 
  -  mount (add) fileSystem
  - check filesystem exists by key

As I can see it was available in v1 but in v2 it was cut...

OskarStark:
Could you please elaborate a bit on your use case? Thanks

EvgenyBarinov:
I've need to prepare lib to build required fileSystems list by provided configuration (for Spiral framework)
At one point of view it will be more comfortable to build MountManager and add all required fileSystems one by one to existing MountManager.
From another point of view I can use it at initialization step with creating MountManager from builded fileSystems list.

I tend to think that first way is more preferable for me.
In current implementation (with private method) MountManager class can be useless for users despite the fact it is very powerful helper in system with several file systems.

For example I would like to prepare some specific handling for copying files from one fileSystem to another. MountManager has all required functionality and I would like to extend it with my specific manager and prepare required functionality.
But I can't do it in current implementation - private methods forbid it for me...

P.S. I am at investigation stage and prepare some basic architecture for this lib.

EvgenyBarinov:
@OskarStark is it really required to keep all methods of MountManager with private access?
As I can see private in libs is not the best solution...

No developer can take into account any scenario of usage which can be actual for another developer.

EvgenyBarinov:
One more example of usage.
When you are using some bootloaders and DI you need to pass MountManager in boot method as argument but your current implementation blocks this method by the reason init filesystems provided only via constructor - it is not applicable in this case. In this case I need some public method for init server

EvgenyBarinov:
Basic example you can find here
https://github.com/thephpleague/flysystem/pull/1263

I am not sure what kind of problem with ftp tests...

Eugentis:
I've created custom implementation like MountManager
Please close the issue

PR #1263, opened 2021-01-26
[feature]: optimize MountManager to make it extendable

EvgenyBarinov:
This modification can help to reuse MountManager or use it in more flexible way.
Using separator in const can be used to provide another pattern for filepath storages

### DIFF
--- a/src/MountManager.php
+++ b/src/MountManager.php
@@ -4,12 +4,14 @@ declare(strict_types=1);
 
 namespace League\Flysystem;
 
-class MountManager implements FilesystemOperator
+class MountManager implements MountManagerInterface, FilesystemOperator
 {
+    protected const FILESYSTEM_PATH_SEPARATOR = '://';
+
     /**
      * @var array<string, FilesystemOperator>
      */
-    private $filesystems = [];
+    protected $filesystems = [];
 
     /**
      * MountManager constructor.
@@ -18,7 +20,25 @@ class MountManager implements FilesystemOperator
      */
     public function __construct(array $filesystems = [])
     {
-        $this->mountFilesystems($filesystems);
+        foreach ($filesystems as $key => $filesystem) {
+            if ($this->isFilesystemExists($key)) {
+                continue;
+            }
+
+            $this->mountFilesystem($key, $filesystem);
+        }
+    }
+
+    public function mountFilesystem(string $key, FilesystemOperator $filesystem): void
+    {
+        $this->guardAgainstInvalidMount($key, $filesystem);
+
+        $this->filesystems[$key] = $filesystem;
+    }
+
+    public function isFilesystemExists(string $key): bool
+    {
+        return array_key_exists($key, $this->filesystems);
     }
 
     public function fileExists(string $location): bool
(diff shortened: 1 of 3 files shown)


src/MountManager.php
```diff
@@ -4,12 +4,14 @@
 
 namespace League\Flysystem;
 
-class MountManager implements FilesystemOperator
+class MountManager implements MountManagerInterface, FilesystemOperator
 {
+    protected const FILESYSTEM_PATH_SEPARATOR = '://';
```
Nyholm, line 9: Could you help me understand in what scenario this needs to be changed?

Nyholm (changes requested):
Thank you for this PR. 

Can you explain **why** these modifications are needed? Why do you need to extend MountManager? 

This PR is also missing tests.

Nyholm:
I have read #1250 and Im not sure I understand the scenario. 

In general, one should always keep things private. One should use composition over inheritance. I also prefer `final`.

src/MountManager.php
Eugentis: Current implementation is very strict about uri format but in some system you can use some other separator for server and filepath. It depends on project.

As I can see the most flexible way can be some regex patter to store some specific pattern with required parts.
For example current pattern can be described in pattern like [this](https://regex101.com/r/2sPIri/1) but in some projects you need to keep more information in uri - project name for example or smth like this

I understand this pattern is not ideal now and can be modified but it is more flexible as I can see and can be used in your method `\League\Flysystem\MountManager::determineFilesystemAndPath` for example

Eugentis:
> I have read #1250 and Im not sure I understand the scenario.
> 
> In general, one should always keep things private. One should use composition over inheritance. I also prefer `final`.

If we are talking about some project it is suitable but if you would like to provide reusable components you should provide more flexible components.

Current implementation of MountManager is not reusable by next reasons:
1. Private doesn't allow to extend it
2. Current implementation can't provide dynamic mounting new filesystems - it is restrict mounting management after MountManager creation. By this reason MountManager is not suitable for some dynamic mounting.
* it can be a problem in frameworks with DI - you need to prepare some initialization methods instead of mapping by some interfaces

Now I am working on component for providing work with multiple system depends on provided configuration and your current implementation suitable for me only by 70-80% but I need to use some modifications provided by different business logic requirements. I would like to use some specific component based on MountManager but it was restricted by private.
From my point of view it is bad practice if I will have class similar to MountManager in most of codebase...

Nyholm:
> If we are talking about some project it is suitable but if you would like to provide reusable components you should provide more flexible components.

A component can be flexible without inheritance. 

> Private doesn't allow to extend it

I see that as a good thing. 

> Current implementation can't provide dynamic mounting new filesystems

That is also a good thing. Services should be immutable. 

> it can be a problem in frameworks with DI - you need to prepare some initialization methods instead of mapping by some interfaces

A good DI will prefer initialization methods. And you do have the `FilesystemOperator` interface...

------------

It sounds like you want to wedge in some use of the MountManager in a way it is not intended to. This makes your code feel bad and instead of fixing your architecture, you want to make the MountManager worse by calling in "more flexible". 

I understand your frustration, I also don't think you will see my point of view. I recommend you to copy the `MountManager` and modify it so it fits your specific needs. 

I vote for closing this PR.

Eugentis:
> Thank you for this PR.
> 
> Can you explain **why** these modifications are needed? Why do you need to extend MountManager?
> 
> This PR is also missing tests.

Tests were added. Mount Manager coverage 100%

Eugentis:
New functionality added for https://github.com/thephpleague/flysystem/issues/1266 for v2

Eugentis:
I suppose in nearest future you will add similar functionality in MountManager. In v1 this class was more friendly and your users will request to make it more flexible to use at least.
Take a look at 1266 for example

Nyholm:
> New functionality added for #1266 for v2

Adding more features in this PR will not make it easer to merge. A general rule is "One feature => One PR".

frankdejonge:
I'm not in favour of adding this functionality in the library. If you need a mount manager which has these capabilities I would advice to implement it yourself. The mount manager used to have these capabilities and I have negative experiences with that setup. For 2.0 there was a request for the mount manager. If you need a mount manager with dynamic filesystems I suggest to create a factory where you can resolve a mount manager with the needed filesystem instead.

Eugentis:
vf

Eugentis:
Ok
Please close this PR

PR #1272, opened 2021-02-06
[CS] Be consistent with blank lines

Nyholm:
I want to avoid multiple blank lines and places where we don't have any blank lines. This is purely for looks.


Nyholm:
Thank you for merging

PR #1273, opened 2021-02-06
Disconnect FTP connection

### LINKED ISSUE #1256: Missing disconnect method for FTP connections in V2 (?)
Is there a way to disconnect an open FTP connection? In V1 there was a [method](https://github.com/thephpleague/flysystem/blob/1.x/src/Adapter/Ftp.php#L236) for this purpose. Unfortunately, it seems to be removed completely... Also, as the connection property is private without any kind of setter/getter, I can't find a simple way to get through to it. Is there a proper workaround?

Nyholm:
This will fix #1256

When the FTP adapter is destructed, then we should disconnect the FTP connection. 

I don't understand why my tests are failing. It is like `ftp_close` have no effect...

### DIFF
--- a/src/Ftp/FtpAdapter.php
+++ b/src/Ftp/FtpAdapter.php
@@ -93,6 +93,17 @@ class FtpAdapter implements FilesystemAdapter
     }
 
     /**
+     * Disconnect FTP connection on destruct.
+     */
+    public function __destruct()
+    {
+        if ($this->connection !== false) {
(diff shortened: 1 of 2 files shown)


frankdejonge:
@Nyholm I figured it out, the test tooling and the reflection held a reference to the object, which is why the destructor didn't fire. I've altered the implementation and it's green now. Thanks for this! 👍

PR #1275, opened 2021-02-12
InMemoryAdapter: Wrong filesize.

Warxcell:
This is just the test case.

Basically `fileSize` returns wrong fileSize, as can be seen with this simple test:

```php
var_dump(filesize($replacement));
var_dump(mb_strlen(file_get_contents($replacement)));
```
Note: `$replacement` is path to file I added in repo.

the output is 

```php
int(22518)
int(17858)
```


src/InMemory/InMemoryFilesystemAdapterTest.php
```diff
@@ -264,6 +264,25 @@ public function checking_for_metadata(): void
         $this->assertStringStartsWith('image/svg', $adapter->mimeType(self::PATH)->mimeType());
     }
 
+    /**
+     * @test
+     */
+    public function checking_for_metadata2(): void
+    {
+        mock_function('time', 1234);
```
Nyholm, line 270: Why is this needed?

src/InMemory/InMemoryFilesystemAdapterTest.php
```diff
@@ -264,6 +264,25 @@ public function checking_for_metadata(): void
         $this->assertStringStartsWith('image/svg', $adapter->mimeType(self::PATH)->mimeType());
     }
 
+    /**
+     * @test
+     */
+    public function checking_for_metadata2(): void
```
Nyholm: Please be more descriptive.

Nyholm (changes requested):
Im not sure what the test is trying to prove. 

Im also very hesitant that content length should always be the same as file size.. Dont files have headers?

The `mb_strlen` will also count characters like `ä` as one character when it indeed is takes twice the space as `a`.

src/InMemory/InMemoryFilesystemAdapterTest.php
Warxcell: I don't know. I just copied the original test and changed the file and metadata that is being checked.

src/InMemory/InMemoryFilesystemAdapterTest.php
Warxcell: Original test is named like that, I just added 2 because it's with second/another file.

Nyholm: That is my point. Name the test better. Dont just add "2".

src/InMemory/InMemoryFilesystemAdapterTest.php
Nyholm: It is good practice to know what the added code does.

Warxcell:
> Im not sure what the test is trying to prove.
> 
> Im also very hesitant that content length should always be the same as file size.. Dont files have headers?
> 
> The `mb_strlen` will also count characters like `ä` as one character when it indeed is takes twice the space as `a`.

Test is trying to prove that `fileSize` doesn't return actual filesize AND furthermore it DOES NOT behave the same way depending if `function_exists('mb_strlen')`, since the result of `mb_strlen !== strlen`

In the end if I do the following:

```
$file = '/path/to/file.jpg';
$local = new \League\Flysystem\Filesystem(new \League\Flysystem\Local\LocalFilesystemAdapter(sys_get_temp_dir()));
$memory = new \League\Flysystem\Filesystem(new \League\Flysystem\InMemory\InMemoryFilesystemAdapter());

$local->write('file.jpg', file_get_contents($file));
$memory->write('file.jpg', file_get_contents($file));

$localFileSize = $local->fileSize('file.jpg');
$memoryFileSize = $memory->fileSize('file.jpg');
if ($localFileSize !== $memoryFileSize) {
    var_dump($localFileSize);
    var_dump($memoryFileSize);
    die('not equal');
}
```

the results is: 

```php
int(22518)
int(17858)
not equal
```

And you are telling me that THIS IS EXPECTED BEHAVIOUR?!

It's the very same file!

Nyholm:
> Test is trying to prove that fileSize doesn't return actual filesize AND furthermore it DOES NOT behave the same way depending if function_exists('mb_strlen'), since the result of mb_strlen !== strlen

But why do you include `lastModified` and `mimeType` in the test?

Also, why would you assume that they should be the same filesize? 

In OSX when I open file information, I get two file sizes, one  with the actual size of the contents and one with the actual space it takes up on the disk. They are not the same, Im not 100% sure why so I can only speculate. But you seam to be convinced that they must always be the same value.. that is why I asked, why are you so convinced?


Screenshot 2021-02-12 at 14 35 42

Warxcell:
> > Test is trying to prove that fileSize doesn't return actual filesize AND furthermore it DOES NOT behave the same way depending if function_exists('mb_strlen'), since the result of mb_strlen !== strlen
> 
> But why do you include `lastModified` and `mimeType` in the test?
> 
> Also, why would you assume that they should be the same filesize?
> 
> In OSX when I open file information, I get two file sizes, one with the actual size of the contents and one with the actual space it takes up on the disk. They are not the same, Im not 100% sure why so I can only speculate. But you seam to be convinced that they must always be the same value.. that is why I asked, why are you so convinced?
> 
> Screenshot 2021-02-12 at 14 35 42

I expect them to be same filesize, because it's very same file, stored on different places. But the file itself is same.

Nyholm:
> I expect them to be same filesize, because it's very same file.

You seam to miss my point. Please read my comment again.

Warxcell:
> > I expect them to be same filesize, because it's very same file.
> 
> You seam to miss my point. Please read my comment again.

Yeah, the adapter could save file in optimized way - but the abstraction should behave same, isn't it? So if I upload one file to several locations I expect it's fileSize to stay the same in the end. 

And furthermore I repeat: `mb_strlen !== strlen`.


```
var_dump(strlen(file_get_contents($file)));
var_dump(mb_strlen(file_get_contents($file)));
```

int(22518)
int(17858)

If my instance doesn't have `mb_strlen` and  tomorrow I install it, suddenly files will have different `fileSize`.

src/InMemory/InMemoryFilesystemAdapterTest.php
Warxcell: Ok, refactored test with data provider.

Nyholm:
I am out on very thin ice now. Im sure I will get some details wrong, but my concept is correct. 

Say that you have 2100 bytes of data that you need to store on a disk. The first disk has a block size of 256 bytes. That means it needs 9 blocks to store the data. 9*256 = 2304. The file will take 2304 bytes on disk.

The second disk has a block size of 1024 bytes. That means you need 3 blocks to store the data. 3 * 1024 = 3072. The same file will take 3072 bytes on the disk. 

It is the same file stored in multiple places. 

----------

This is the difference between the memory adapter (which counts the exact size in memory) and the filesystem adapter (which returns the size reported from the filesystem). 

---------

> And furthermore I repeat: mb_strlen !== strlen.

I dont understand what you mean with this. Yes, it is only natural that two different functions return different results. See the manual if you are not sure what `mb_strlen` is doing: https://www.php.net/manual/en/function.mb-strlen.php

Warxcell:
> I am out on very thin ice now. Im sure I will get some details wrong, but my concept is correct.
> 
> Say that you have 2100 bytes of data that you need to store on a disk. The first disk has a block size of 256 bytes. That means it needs 9 blocks to store the data. 9*256 = 2304. The file will take 2304 bytes on disk.
> 
> The second disk has a block size of 1024 bytes. That means you need 3 blocks to store the data. 3 * 1024 = 3072. The same file will take 3072 bytes on the disk.
> 
> It is the same file stored in multiple places.
> 
> This is the difference between the memory adapter (which counts the exact size in memory) and the filesystem adapter (which returns the size reported from the filesystem).
> 
> > And furthermore I repeat: mb_strlen !== strlen.
> 
> I dont understand what you mean with this. Yes, it is only natural that two different functions return different results. See the manual if you are not sure what `mb_strlen` is doing: https://www.php.net/manual/en/function.mb-strlen.php

```
mkfs -t ext4 -b 1280 /dev/nvme0n1p5
mkfs -t ext4 -b 4096 /dev/nvme0n1p4

mount /dev/nvme0n1p4 test1
mount /dev/nvme0n1p5 test2

cp image2.jpg test1/
cp image2.jpg test2/

ls -l test1/
-rw-r--r-- 1 root root *22518* Feb 12 15:59 image2.jpg


ls -l test2/
-rw-r--r-- 1 root root *22518* Feb 12 15:59 image2.jpg

du -h test1/image2.jpg 
24K     test1/image2.jpg

du -h test2/image2.jpg 
22K     test2/image2.jpg
```

Allocated size on DISK is different, But the size of the file it self IS SAME = 22518.

Nyholm:
So.. everything is fine? 

In your your original post, you reacted that this code had different output: 

```
var_dump(filesize($replacement));
var_dump(mb_strlen(file_get_contents($replacement)));
```
and as [you discovered](https://github.com/thephpleague/flysystem/pull/1275#issuecomment-778214940), the content length and the file size on disk is different. 

If you are using FilesystemAdapter, it will use the filesystem to ask what size the file is. If you are using InMemoryAdapter, it will do `mb_strlen` or `strlen` on the file contents to give you the size.

Warxcell:
> So.. everything is fine?
> 
> In your your original post, you reacted that this code had different output:
> 
> ```
> var_dump(filesize($replacement));
> var_dump(mb_strlen(file_get_contents($replacement)));
> ```
> 
> and as [you discovered](https://github.com/thephpleague/flysystem/pull/1275#issuecomment-778214940), the content length and the file size on disk is different.
> 
> If you are using FilesystemAdapter, it will use the filesystem to ask what size the file is. If you are using InMemoryAdapter, it will do `mb_strlen` or `strlen` on the file contents to give you the size.

In my opinion it should always return same thing, as you can see `ls` or `filesize` php function returns same file size, no matter of blocks of the disks. `du` returns different allocated space ,depending on blocks. 
Adapters should work like `ls` or `filesize` PHP function.

Nyholm:
> In my opinion it should always return same thing

How do you suggest to fix that? Say the `InMemoryAdapter`. Here is how the file size is calculated: https://github.com/thephpleague/flysystem/blob/5ae7a6dcb7c1d47cff0bc7199c1554b777849c40/src/InMemory/InMemoryFile.php#L59-L64

How would you make it work like `ls` or `filesize` PHP function?

Warxcell:
> > In my opinion it should always return same thing
> 
> How do you suggest to fix that? Say the `InMemoryAdapter`. Here is how the file size is calculated:
> 
> https://github.com/thephpleague/flysystem/blob/5ae7a6dcb7c1d47cff0bc7199c1554b777849c40/src/InMemory/InMemoryFile.php#L59-L64
> 
> How would you make it work like `ls` or `filesize` PHP function?

Simplest option would be to like that:
```php
 public function fileSize(): int 
 { 
     return strlen($this->contents); 
 }
```

Nyholm:
That would not make it the same as the `FilesystemAdapter`. So it would not fix the issue.

Warxcell:
> That would not make it the same as the `FilesystemAdapter`. So it would not fix the issue.

What you mean? What's the connection to the interface?

Nyholm:
You said: "In my opinion it should always return same thing,"

Then you provided a "fix" to make then return the same thing. 

Im telling you that your fix will not make the `InMemoryAdapter` return the same filesize as `LocalFilesystemAdapter`.

Warxcell:
> You said: "In my opinion it should always return same thing,"
> 
> Then you provided a "fix" to make then return the same thing.
> 
> Im telling you that your fix will not make the `InMemoryAdapter` return the same filesize as `LocalFilesystemAdapter`.

It will.

Take as example this one:

```php
var_dump(filesize($file));
var_dump(strlen(file_get_contents($file)));
var_dump(mb_strlen(file_get_contents($file)));
```

```
int(22518)
int(22518)
int(17858)
```

Did you get my point of why `InMemoryAdapter` can behaves differently.

Nyholm:
So I finally undertand what you mean. You are saying that the `InMemoryAdapter` is returning the content length and not the number of bytes. 

If that is correct, the PR #1276 will address that.

PR #1276, opened 2021-02-12
InMemoryAdapter must return the number of bytes

Nyholm:
Currently, it returns the length of the content. 

This will fix #1275

```php
echo mb_strlen('täst'); // => 4
echo strlen('täst'); // => 5
```

We are always interested the byte length.


frankdejonge:
Thank you @Nyholm 👍

Nyholm:
Thank you for merging

PR #1277, opened 2021-02-12
Add AzureBlobStorageAdapter.php

Warxcell:
Original PR: https://github.com/thephpleague/flysystem-azure-blob-storage/pull/25


Nyholm:
The environment variables `FLYSYSTEM_AZURE_ACCOUNT_KEY` and `FLYSYSTEM_AZURE_ACCOUNT_NAME` are currently not defined in the Github actions workflow. They also do not exists as secrets. 

However, Github secrets are not available in PRs, only on jobs running in a non-fork. One need to make sure tests dont fail without credentials and also with credentials. 

I think the fastest way forward is to you to signup for a free Azure account and make sure the tests are green locally. I know @frankdejonge will double check the tests locally before he merge this PR anyways.

Warxcell:
How is handled the case when the adapter doesn't support visibility? In V1 it does/return nothing, but in V2 tests expect adapters to always support it thus failing.

Warxcell:
Ping @Nyholm

Nowicki:
Maybe you could overwrite or skip tests. For example:
```php
/**
 * @test
 */
public function overwriting_a_file(): void
{
    $this->runScenario(function () {
        $this->givenWeHaveAnExistingFile('path.txt', 'contents');
        $adapter = $this->adapter();

        $adapter->write('path.txt', 'new contents');

        $contents = $adapter->read('path.txt');
        $this->assertEquals('new contents', $contents);
    });
}

/**
 * @test
 */
public function setting_visibility(): void
{
    $this->markTestSkipped('Visibility is not supported');
}

/**
 * @test
 */
public function failing_to_set_visibility(): void
{
    $this->expectException(VisibilityException::class);
    
    $this->runScenaion(function() {
        $adapter = $this->adapter();
        $this->givenWeHaveAnExistingFile('path.txt', 'contents');

        $adapter->setVisibility('path.txt', Visibility::PRIVATE);
    });
}

/**
 * @test
 */
public function failing_to_check_visibility(): void
{
    $this->expectException(VisibilityException::class);
    
    $this->runScenaion(function() {
        $adapter = $this->adapter();
        $this->givenWeHaveAnExistingFile('path.txt', 'contents');
        
        $adapter->visibility('path.txt')->visibility();
    });
}
```

OskarStark:
Thanks for your work on this feature!

It looks like a rebase is needed

Warxcell:
> Thanks for your work on this feature!
> 
> It looks like a rebase is needed

Yeah, will do it these days. Also need to skip tests for permissions/visibility.

Warxcell:
Conflicts are resolved and I marked few tests as skipped. Unfortunately I don't have Azure Blob account currently I can test with.

OskarStark:
@Warxcell I will try to find some time and test this with a AzureBlobStorage and let you know if everything is fine 👍

src/AzureBlobStorage/AzureBlobStorageAdapter.php
```diff
@@ -0,0 +1,284 @@
...
+            $options->setContentType($mimeType);
+        }
+
+        return $options;
+    }
+
+    private function normalizeBlobProperties($path, BlobProperties $properties): FileAttributes
+    {
+        return new FileAttributes(
+            $path,
```
snoob: What do you think about $this->client->getBlobUrl($this->container, $path) here ? to get the blob full path ?

src/AzureBlobStorage/AzureBlobStorageAdapter.php
```diff
@@ -0,0 +1,284 @@
...
+            return $response->getContentStream();
+        } catch (Throwable $exception) {
+            throw UnableToReadFile::fromLocation($path);
+        }
+    }
+
+    public function listContents(string $path, bool $deep = false): iterable
+    {
+        if (strlen($path) > 0) {
+            $path = rtrim($path, '/') . '/';
```
snoob: The trailing slash makes the filter by blob in the current container impossible. What do you think about remove it ?

src/AzureBlobStorage/AzureBlobStorageAdapter.php
```diff
@@ -0,0 +1,284 @@
...
+    {
+        try {
+            $options = new ListBlobsOptions();
+            $options->setPrefix($path . '/');
+            $listResults = $this->client->listBlobs($this->container, $options);
+            foreach ($listResults->getBlobs() as $blob) {
+                $this->client->deleteBlob($this->container, $blob->getName());
+            }
+        } catch (Throwable $exception) {
+            UnableToDeleteDirectory::atLocation($path, '', $exception);
```
snoob: throw is missing

src/AzureBlobStorage/AzureBlobStorageAdapter.php
```diff
@@ -0,0 +1,284 @@
...
+
+    private function getMetadata($path): FileAttributes
+    {
+        return $this->normalizeBlobProperties(
+            $path,
+            $this->client->getBlobProperties($this->container, $path)->getProperties()
+        );
+    }
+
+    public function getSize($path): FileAttributes
```
snoob: duplicate with function fileSize(). Why did you implement this one ?

Warxcell: Sorry, I basically copied the old one, and refactored it. It was leftover. :) 
Removed it.

src/AzureBlobStorage/AzureBlobStorageAdapter.php
Warxcell: Fixed.

src/AzureBlobStorage/AzureBlobStorageAdapter.php
Warxcell: This is copied from the older adapter. https://github.com/thephpleague/flysystem-azure-blob-storage/blob/master/src/AzureBlobStorageAdapter.php#L199

src/AzureBlobStorage/AzureBlobStorageAdapter.php
Warxcell: It's implemented as it was in the old adapter: https://github.com/thephpleague/flysystem-azure-blob-storage/blob/master/src/AzureBlobStorageAdapter.php#L291

src/AzureBlobStorage/AzureBlobStorageAdapter.php
snoob: Ok thank you

src/AzureBlobStorage/AzureBlobStorageAdapter.php
snoob: Ok thank you

snoob:
@Warxcell On a project, i have copy/pasted your class and covered it with unit test (i have fully mocked Azure API). Are you interested with my test class ?

src/AzureBlobStorage/AzureBlobStorageAdapter.php
snoob: No problem

Warxcell:
> @Warxcell On a project, i have copy/pasted your class and covered it with unit test (i have fully mocked Azure API). Are you interested with my test class ?

Not sure what's the practice here. There are integrational test, but I can add it, no problem.

snoob:
> > @Warxcell On a project, i have copy/pasted your class and covered it with unit test (i have fully mocked Azure API). Are you interested with my test class ?
> 
> Not sure what's the practice here. There are integrational test, but I can add it, no problem.

I think integration tests are more relevant for flysystem adapters but i also don't know what the project philosophy is

src/AzureBlobStorage/.gitattributes
```diff
@@ -0,0 +1,8 @@
+* text=auto
+
+tests/
+.gitignore
+.gitattributes
+.travis.yml
+phpunit.xml
+
```
Nyholm: Please update this file. See other packages. Ie Ftp or AwsS3V3

src/AzureBlobStorage/.gitignore
```diff
@@ -0,0 +1,5 @@
+/vendor/
```
Nyholm: This file is not needed

src/AzureBlobStorage/AzureBlobStorageAdapter.php
```diff
@@ -0,0 +1,275 @@
...
+        } catch (Throwable $exception) {
+            throw UnableToDeleteFile::atLocation($path, '', $exception);
+        }
+    }
+
+    public function read(string $path): string
+    {
+        $response = $this->readStream($path);
+
+        return stream_get_contents($response);
```
Nyholm, line 92: Is the stream always rewinded?

src/AzureBlobStorage/AzureBlobStorageAdapter.php
```diff
@@ -0,0 +1,275 @@
...
+            foreach ($response->getBlobs() as $blob) {
+                $name = $blob->getName();
+
+                if ($path === '' || strpos($name, $path) === 0) {
+                    yield $this->normalizeBlobProperties($name, $blob->getProperties());
+                }
+            }
+
+            if (!$deep) {
+                foreach ($response->getBlobPrefixes() as $blobPrefix) {
```
Nyholm, line 125: Is a "blob prefix" the same as a directory?

src/AzureBlobStorage/AzureBlobStorageAdapter.php
```diff
@@ -0,0 +1,275 @@
...
+        'ContentEncoding',
+    ];
+    /** @var BlobRestProxy */
+    private $client;
+    /** @var string */
+    private $container;
+    /** @var MimeTypeDetector */
+    private $mimeTypeDetector;
+    /** @var int */
+    private $maxResultsForContentsListing;
```
Nyholm, line 51: Is a "max result" really needed? Why would I use a smaller value like 5 for this?

src/AzureBlobStorage/AzureBlobStorageAdapter.php
```diff
@@ -0,0 +1,275 @@
...
+        try {
+            $response = $this->client->getBlob($this->container, $path);
+
+            return $response->getContentStream();
+        } catch (Throwable $exception) {
+            throw UnableToReadFile::fromLocation($path);
+        }
+    }
+
+    public function listContents(string $path, bool $deep = false): iterable
```
Nyholm, line 107: Why do you specify a default value for `$deep`?

Nyholm (commented):
Thank you. 

I have a few questions/comments.

src/AzureBlobStorage/AzureBlobStorageAdapter.php
Warxcell: I guess optimization. [It was done like that in V1](https://github.com/thephpleague/flysystem-azure-blob-storage/blob/master/src/AzureBlobStorageAdapter.php#L45) so I just ported it as-is.

src/AzureBlobStorage/AzureBlobStorageAdapter.php
Warxcell: [If was done like that in V1, so I just ported it as-is.](https://github.com/thephpleague/flysystem-azure-blob-storage/blob/master/src/AzureBlobStorageAdapter.php#L193)

src/AzureBlobStorage/AzureBlobStorageAdapter.php
Warxcell: Should be, as it's brand new object. [In V1 is the same way.](https://github.com/thephpleague/flysystem-azure-blob-storage/blob/master/src/AzureBlobStorageAdapter.php#L164)

src/AzureBlobStorage/AzureBlobStorageAdapter.php
Warxcell: https://github.com/thephpleague/flysystem-azure-blob-storage/blob/master/src/AzureBlobStorageAdapter.php#L222

OskarStark:
Could you please update the PR title and the PR header to the infos from the original PR? Thanks

snoob:
@OskarStark What is the exactly the purpose of this PR ?  Migrate the v1 azure adapter like it was ? Or can we add some features ?

For example the method createDirectory is empty.
I implemented it on a project  :

```
public function createDirectory(string $path, Config $config): void
    {
        try {
            $this->client->createContainer($path);
        } catch (Throwable $exception) {
            if ($exception instanceof ServiceException && Response::HTTP_CONFLICT === $exception->getCode()) {
                throw UnableToCreateDirectory::alreadyExists($path, $exception);
            }
            throw UnableToCreateDirectory::anyReason(
                \sprintf('Unable to create container "%s".', $path),
                $exception
            );
        }
    }
```

OskarStark:
> @OskarStark What is the exactly the purpose of this PR ? Migrate the v1 azure adapter like it was ? Or can we add some features ?

From my POV adding features is fine. @Warxcell are you willing to add and test the `createDirectory()` method?

EDIT:
and the `deleteDirectory()` method

snoob:
@OskarStark There is also a lack on the method deleteDirectory : you can only delete the current container (passed in the constructor) or a sub container. 

With this simplier implementation, you can delete any container : 

```
public function deleteDirectory(string $path): void
  {
      $this->client->deleteContainer($path);
  }
```

Warxcell:
If we add containers as directories I think I should refactor the adapter itself and remove `string $container` as constructor argument, as it changes the philosophy of the how the Adapter works.

OskarStark:
In this case I would first go with you integration which is almost ready and add the 2 features afterwards

Warxcell:
> In this case I would first go with you integration which is almost ready and add the 2 features afterwards

But that would mean BC breaks. I think we can add 2 adapters here. This one works as-is, the other one will treat directories as containers.

snoob:
> If we add containers as directories I think I should refactor the adapter itself and remove `string $container` as constructor argument ?

+1 for me ! But the DX will be degraded a bit as the developer will have to pass $container . '/' . $blob on most method calls. Anyway this can be solve by a decoration to append it automatically.

OskarStark:
So adding it directly would result in one adapter and a better DX?

Warxcell:
actually it's mandatory to refactor it, because create/remove dir will actually do only that. These newly created containers will be never referenced again, because read/write/etc always uses `$this->container`

snoob:
> So adding it directly would result in one adapter and a better DX?

1. Having it in the constructor : 
- Pros : 
  - Better DX 
- Cons : 
  - Method deleteDirectory and createDirectory are not consistent with others methods as the path argument will have another behavior
  - You have to define a service to manipulate multiple container in the same project (one service per container)

2. NOT having it in the constructor : 
- Pros : 
  - Every methods are consistent 
  - You can manipulate different containers 
- Cons : 
  - DX lowered as ou have to pass $container . '/' . $blob on most method calls (Can be solved by a decoration project side). The decoration is not needed if you push blog at the root. This could be explained in the documentation

snoob:
> because create/remove dir will actually do only that. These newly created containers will be never referenced again, because read/write/etc always uses `$this->contai`

good point

Warxcell:
I would vote for 2 adapters. I don't really know all the use cases, I have single project with 2 containers `private_file` and `public_files`. And I have 2 Flysystem instances (1 for private_files and 1 for public_files) and it works great, because they are isolated from each other. 

The question is how will the directories be treated?

`dir1/dir2/file.png` Results in what? `container: dir1` and `blob: dir2/file.png` or container: `dir1/dir2` and `blob: file.png`?

snoob:
@Warxcell 

The solution without the container in constructor is more flexible you can do anything (with a bit of code)

The old behavior can be obtained like this no ?
```
class MyPublicAzureBlobStorageAdapter implements FilesystemAdapter
{
    private AzureBlobStorageAdapter $decorated;
    
    public function __construct(AzureBlobStorageAdapter $decorated, string $container)
    {
    }

    public function fileExists(string $path): bool
    {
        return $this->decorated->fileExists($container . '/ . $path);
    }
}

```

Warxcell:
> @Warxcell
> 
> The solution without the container in constructor is more flexible you can do anything (with a bit of code)
> 
> The old behavior can be obtained like this no ?
> 
> ```
> class MyPublicAzureBlobStorageAdapter implements FilesystemAdapter
> {
>     private AzureBlobStorageAdapter $decorated;
>     
>     public function __construct(AzureBlobStorageAdapter $decorated, string $container)
>     {
>     }
> 
>     public function fileExists(string $path): bool
>     {
>         return $this->decorated->fileExists($container . '/ . $path);
>     }
> }
> ```

Depends on the answer of question:

dir1/dir2/file.png Results in what? container: dir1 and blob: dir2/file.png or container: dir1/dir2 and blob: file.png?

snoob:
> > @Warxcell

> Depends on the answer of question:
> 
> dir1/dir2/file.png Results in what? container: dir1 and blob: dir2/file.png or container: dir1/dir2 and blob: file.png?

Oh sorry. 
For me a / should identify a container. It's the current behavior on listContent method if i'am not wrong so container: dir1/dir2 and blob: file.png.

Moreover, it's more consistent with other adapter right ?

Warxcell:
> Moreover, it's more consistent with other adapter right ?

Actually no. Currently it's consistent as other adapters: https://github.com/thephpleague/flysystem/blob/2.x/src/AwsS3V3/AwsS3V3Adapter.php#L108

AWS S3 works with single `bucket`.

Warxcell:
Ok I made changes that should solve both issues.

src/AzureBlobStorage/Resolved.php
```diff
@@ -0,0 +1,30 @@
+<?php
+
+declare(strict_types=1);
+
+namespace League\Flysystem\AzureBlobStorage;
+
+class Resolved
```
snoob: * ResolvedContainerPath ?
* ResolvedAzureLocation ?

WDYT?

Warxcell: Breaks atomicity?

src/AzureBlobStorage/AzureBlobStorageAdapter.php
```diff
@@ -0,0 +1,303 @@
...
+    {
+        if (strlen($path) > 0) {
+            $path = rtrim($path, '/').'/';
```
snoob: I know you ported it like that but the trailing slash doesn't not allow to search in the current container. It will be a nice feature to allow it no ?

src/AzureBlobStorage/Resolved.php
snoob: Ok

src/AzureBlobStorage/AzureBlobStorageAdapter.php
Warxcell: So basically I should leave it as ` $path = rtrim($path, '/');` ?

snoob: On my project, i completely deleted these 3 lines (105 to 107) to leave completely free to do everything finding a blob or a container. If you always remove the trailing slash you won't be able to search in a container ?

Warxcell: ok I see. Is it possible to register free azure account so I can test it?

Warxcell:
actually supporting multiple container is even harder than I thought initially, cause `listContents` now needs to iterate all containers if `path === ''`. I'm not sure I want to support this complexity. @snoob do you want to take over and finish multiple containers support? 

Also what happens if `write` receives `path === 'file.jpg'` ? Where that goes? It basically have no container. I've put `defaultContainer` in that case, but.... it feels like workaround.

In same time listContents resolve path to container and inner path should resolve to null to be able to list all containers, but when writing a file it should resolve in specific backup container.

Examples:

write('file.jpg') => which container is that?
listContents('') => what should this list? all files from all containers?

Furthermore: there are no directories in azure until file is actually there, so tests will always fail. 
we could fake it by creating file with empty contents and then read the file and if it's empty == return directory, else return file, but this is also hacky.

Warxcell:
I'm done with that. I have it locally on my project and it's basically working and I'm happy with it.

I won't do anything else here. If you want merge it, if not... whatever.

Anyone can takeover this.

snoob:
I'am in holidays right now but i can give a try when i comeback (around the 15th August). If you consider that the PR is mergeable like that it can be an improvement in a second one. 

Thank you for trying @Warxcell

torchello:
Thanks everybody for working on this! @snoob are you still going to work on/finalize this PR? What are the remaining tasks?

snoob:
@torchello Sorry for the late answer.

With the current implementation, some methods can't be implemented or will have some weird behavior : 
* createDirectory 
* deleteDirectory

This is because the container is in the constructor https://github.com/thephpleague/flysystem/pull/1277#issuecomment-882386265

We discussed about remove it and @Warxcell gave a try but it was more difficult that it looked

@OskarStark what i should do ? 
The current code works but if we merge it and decide to remove the container from the construct after it will lead to a BC break (except if we add the comptability layer).
So i can give a try.  

Or we create a second adapter ?

OskarStark:
This decision should be made by @Nyholm or @frankdejonge

Nyholm:
I would prefer to have the container in the constructor. That makes is similar to AWS which has the S3 bucket in the constructor. 

> With the current implementation, some methods can't be implemented or will have some weird behavior :
> 
> createDirectory
> deleteDirectory

If Azure does not support empty directories. Then I guess it is fine to leave the body of `createDirectory` empty.

OskarStark:
Or throw an exception that this is not provided?

snoob:
@Nyholm WDTY about something like this ? This is not finalized i need to provide a way to configure the ContainerResolver

https://github.com/snoob/flysystem/commit/1b35354f2630dff7f8ce2ecb11af7f5cc5cea974

Nyholm:
> Or throw an exception that this is not provided?

I don't think so. That would break LSP. You may still use the Azure container just as the directory existed. Right?

> @Nyholm WDTY about something like this ? This is not finalized i need to provide a way to configure the ContainerResolver

Hm.. I appriciate the flexibility. But Im not sure why it is needed. Ie, the AWS S3 adapter has done pretty well without this extra complexity. I dont think that commit should be in the initial version of the adapter. But let's bring it up for discussion when this PR is merged.

Warxcell:
I guess I will make the adapter extendible, so anyone can extend it and
override parts of it.
Its a solution in the middle. So snoob can override it and implement the
container thing.

На нд, 19.09.2021 г., 9:58 ч. Tobias Nyholm ***@***.***>
написа:

snoob:
> But let's bring it up for discussion when this PR is merged.

@Nyholm If we merge it, the container resolver won't be mergeable before 3.x right ? (because of BC break?)

I personally have the use as i work for a multitenant projet (symfony application with flysystembundle) and the container depends of the current customer. So it should be resolved at runtime

Warxcell:
I added again path resolver, but implemented only static one. So it's up to you do implement it however you want in your app. What do you think now?

src/AzureBlobStorage/StaticContainerPathResolver.php
```diff
@@ -0,0 +1,22 @@
...
+    private $container;
+
+    public function __construct(string $container)
+    {
+        $this->container = $container;
+    }
+
+    public function resolve(string $path): Path
+    {
+        return new Path($this > $this->container, $path);
```
maxhelias:
```suggestion
        return new Path($this->container, $path);
```

OskarStark:
Any news on this PR

@frankdejonge can you please approve the workflow run? Thanks

src/AzureBlobStorage/AzureBlobStorageAdapter.php
frankdejonge: this is the "per page" setting

src/AzureBlobStorage/AzureBlobStorageAdapter.php
frankdejonge: pretty much in azure land

PR #1281, opened 2021-03-05
New ArrayAdapter adapter to add namespace capabality

ElGigi:
Creation of a not a really new adapter but an array of adapter.
Each adapter has name, and can be called by syntax "@name".

```php
$adapter = new ArrayAdapter(
    [
        'main' => new League\Flysystem\Local\LocalFilesystemAdapter('.../directory1'),
        'memory' => new League\Flysystem\InMemory\InMemoryFilesystemAdapter(),
    ],
    'main'
);

$filesystem->read('@main/my-file.txt'); // Read on local filesystem adapter
$filesystem->read('@memory/my-file.txt'); // Read on memory adapter
$filesystem->read('/my-file.txt'); // Read on default namespace, so the local filesystem adapter
```

When copy/move methods are called on two namespaces, it's used readStream/writeStream methods to copy or move files.


ElGigi:
I preferred create a pull request than a independant package because i think it's a good basic functionnality for library.

frankdejonge:
@ElGigi Hi, is there anything in the mount manager that does not fit your requirements? It looks quite similar?

ElGigi:
My bad! I didn't found it in documentation.

So can be change visibility of method MountManager::determineFilesystemAndPath() to protected, to extends class and choose separator and how namespace are written?

Nyholm:
Isn't the "namespace" a feature of the adapter itself and not the manager?

ElGigi:
Yes it's the approach of the ArrayAdapter.
I looked at the MountManager, it's necessary to instantiate as many flysystem as adapter. In the ArrayAdapter, only one flysystem instance necessary.

I can update ArrayAdapter to have the same syntax as MountManager.

Nyholm:
I think it is a misunderstanding. 

> > So can be change visibility of method MountManager::determineFilesystemAndPath() to protected, to extends class and [...] how namespace are written?
>
> Isn't the "namespace" a feature of the adapter itself and not the manager?

To specify path/namespace is part of the adapter it self. That should not be the responsibility of the manager. I dont think extending the `MountManager` for this reason is a good idea and it should be avoided. You should configure your adapters instead.

PR #1299, opened 2021-04-22
Setting metadata when uploading an object into a GCS bucket

### LINKED ISSUE #2: Impossible to set storage object metadata
Would you be able to add possibility to set a metadata?

I'm new to this but my tests shows, that something like that is properly working:
https://github.com/thephpleague/flysystem-google-cloud-storage/blob/2.x/GoogleCloudStorageAdapter.php#L88

```php
    'predefinedAcl' => $this->visibilityHandler->visibilityToPredefinedAcl($visibility),
    'metadata' => $config->get('metadata') ?: [],
]
```

mmetayer:
Allow setting metadata when writing to a GCS bucket. Specifically, this would allow defining the object's mime type through the `metadata['contentType']` entry (if omitted, GCS will serve the object as `application/octet-stream`)

Fixes https://github.com/thephpleague/flysystem-google-cloud-storage/issues/2

### DIFF
--- a/src/GoogleCloudStorage/GoogleCloudStorageAdapter.php
+++ b/src/GoogleCloudStorage/GoogleCloudStorageAdapter.php
@@ -22,6 +22,8 @@ use League\Flysystem\UnableToRetrieveMetadata;
 use League\Flysystem\UnableToSetVisibility;
 use League\Flysystem\UnableToWriteFile;
 use League\Flysystem\Visibility;
+use League\MimeTypeDetection\FinfoMimeTypeDetector;
+use League\MimeTypeDetection\MimeTypeDetector;
 use Throwable;
 
 class GoogleCloudStorageAdapter implements FilesystemAdapter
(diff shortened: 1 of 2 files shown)


src/GoogleCloudStorage/GoogleCloudStorageAdapter.php
```diff
@@ -78,13 +78,15 @@ private function upload(string $path, $contents, Config $config): void
     {
         $prefixedPath = $this->prefixer->prefixPath($path);
         $visibility = $config->get(Config::OPTION_VISIBILITY, $this->defaultVisibility);
+        $metadata = $config->get('metadata', []);
```
mmetayer, line 102: should we restrict the metadata entries to the ones listed in https://cloud.google.com/storage/docs/json_api/v1/objects/insert#request-body, or should it be the responsibility of the developer?

mmetayer: Also, should we try to guess the uploaded file's mime type, as in https://github.com/thephpleague/flysystem/blob/65bf14d6ac610bb0054d43a82dca83d07cc34032/src/AwsS3V3/AwsS3V3Adapter.php#L150?

mmetayer: @frankdejonge could you give me your opinion on this when you have time?

Nyholm: > should we restrict the metadata entries to the ones listed

I don't think we should. Let the server validate this instead.

Nyholm: Using `metadata` like this would mean you have one set of metadata per filesystem. Ie, you cannot upload a text file then an image and expect them both to have the correct content type. 

Right?

Nyholm (commented):
Thank you for this PR.

src/GoogleCloudStorage/GoogleCloudStorageAdapter.php
mmetayer: Right. Do you think this is an issue? Also, we could try to determine the file's mime type, but that would require injecting/instantiating a MimeTypeDetector

odombrovskyi-dev:
Hi guys, any updates on this?

mmetayer:
@frankdejonge @nyholm is there anything I can do to move things forward? Would you like me to update this PR?

mmetayer:
I ended up adding a commit to try to guess the mime type if not provided, just like in the AWSS3V3 adapter. What do you think?

ValentinMumble (approved):

ValentinMumble:
Hello everybody, would anyone with write access have news on this?

mmachatschek:
Just a bump for the stale bot so it's not getting closed. This needs review of one of the maintainers.

PR #1302, opened 2021-04-30
Remove rawurlencode() from CopySource argument

derrabus:
I currently have the problem that I'm unable to copy files on a DigitalOcean S3 bucket using Flysystem and the AsyncAWS adapter. Digging into this issue, I suspected that the source file name is URL-encoded twice in the HTTP requests. In fact, if I remove the `rawurlencode()` call, it works like a charm.


Nyholm (approved):
AWS does not care if we run `rawurlencode()` or not. 

The header will already be encoded. I guess AWS is nice enough to take care of the double encoding while Digital Ocean is more strict.

Nyholm:
Wohoo. Thank you for merging

derrabus:
Thank you very much!

PR #1306, opened 2021-05-04
Removed generic exception from phpdoc

### LINKED ISSUE #1304: Remove FilesystemException from annotations
`FilesystemException` should not be annotated in `@throws`.

If you catch e.g. `UnableToWriteFile` from `FilesystemWriter#write()` you cover all errors which can occur in method `write()`. There is no need to annotate the interface as well. You can still catch whole `FilesystemException` as it is inferred from `UnableToWriteFile`.

The problem is that IDE complains now that you should catch also `FilesystemException`:

image

dakur:
Closes #1304

### DIFF
--- a/src/FilesystemAdapter.php
+++ b/src/FilesystemAdapter.php
@@ -7,13 +7,12 @@ namespace League\Flysystem;
 interface FilesystemAdapter
 {
     /**
-     * @throws FilesystemException
+     * @throws UnableToCheckFileExistence
      */
     public function fileExists(string $path): bool;
 
     /**
      * @throws UnableToWriteFile
-     * @throws FilesystemException
      */
     public function write(string $path, string $contents, Config $config): void;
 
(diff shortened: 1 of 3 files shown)


src/FilesystemAdapter.php
```diff
@@ -7,102 +7,86 @@
 interface FilesystemAdapter
 {
     /**
-     * @throws FilesystemException
+     * @throws UnableToCheckFileExistence
```
Nyholm, line 10: I like that you added this line, but dont remove the `FilesystemException`.

Nyholm (changes requested):
I dont think it is a good idea to remove `FilesystemException` from everywhere. The exception can be thrown as stated in https://github.com/thephpleague/flysystem/issues/1304#issuecomment-831944486. 

Also, this brings no clear advantage. 

Im happy with a PR that just adds `UnableToCheckFileExistence` on `fileExists()`

dakur:
> 
> 
> This issue has been automatically marked as stale because it has not had recent activity. It will be closed if no further activity occurs. Thank you for your contributions.

I plan to reply

dakur:
Sorry, I'm not able to finish this. But the idea was sketched in #1304 and https://github.com/thephpleague/flysystem/issues/1304#issuecomment-832474936 even though some of changes in this PR [might not be appropriate](https://github.com/thephpleague/flysystem/pull/1306#discussion_r631588942).

PR #1314, opened 2021-05-20
[2.x] Add extra metadata to DirectoryAttributes (Consistent with FileAttributes)

erikn69:
Please help.

I'm migrating custom adapter to V2 and i need extra metadata for directories, not breaking changes, with test and consistent with FileAttributes

This fix me a bug with laravel 9.x, @Nyholm suggestions?

### DIFF
--- a/src/DirectoryAttributes.php
+++ b/src/DirectoryAttributes.php
@@ -90,7 +90,7 @@ class DirectoryAttributes implements StorageAttributes
             $attributes[StorageAttributes::ATTRIBUTE_PATH],
             $attributes[StorageAttributes::ATTRIBUTE_VISIBILITY] ?? null,
             $attributes[StorageAttributes::ATTRIBUTE_LAST_MODIFIED] ?? null,
-            $attributes[StorageAttributes::ATTRIBUTE_EXTRA_METADATA] ?? [],
+            $attributes[StorageAttributes::ATTRIBUTE_EXTRA_METADATA] ?? []
         );
     }
 
--- a/src/DirectoryAttributes.php
+++ b/src/DirectoryAttributes.php
@@ -28,11 +28,17 @@ class DirectoryAttributes implements StorageAttributes
      */
     private $lastModified;
 
-    public function __construct(string $path, ?string $visibility = null, ?int $lastModified = null)
+    /**
+     * @var array
+     */
+    private $extraMetadata;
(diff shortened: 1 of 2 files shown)


Nyholm:
Thank you. 

Could you elaborate on the use case? What kind of attribute do you store?

Is this fixing a bug in Laravel 9 using your custom adapter or is it fixing a bug in Laravel 9 for everyone?

PaolaRuby:
> Thank you.
> 
> Could you elaborate on the use case? What kind of attribute do you store?
> 
> Is this fixing a bug in Laravel 9 using your custom adapter or is it fixing a bug in Laravel 9 for everyone?

@Nyholm I'm upgrading a google drive adapter to V2, in this case i use the id from google like extra metadata for files, directories also has unique ids, without extra metadata there it's broke the adapter, i try  to create custom DirectoryAttributes class but it broke league test cases for V2, so i extend original DirectoryAttributes, and then pass all test, but after testing on laravel 9 stop working with directories

erikn69:
> Thank you.
> 
> Could you elaborate on the use case? What kind of attribute do you store?
> 
> Is this fixing a bug in Laravel 9 using your custom adapter or is it fixing a bug in Laravel 9 for everyone?

@Nyholm I'm upgrading a google drive adapter to V2, in this case i use the id and the parent id from google like extra metadata for files, directories also has unique ids, without extra metadata there, it's broke the adapter, i try  to create custom DirectoryAttributes class but it brokes league test cases for V2 `League\Flysystem\AdapterTestUtilitiesFilesystemAdapterTestCase.php`, so i extend original `League\Flysystem\DirectoryAttributes`, and then pass all test, but after when i start testing on laravel 9 stop working with directories, so custom DirectoryAttributes is the problem because works when i filter only files, maybe extra meta data can help to others customs adapter

**UPDATE:** In laravel 9 using `jsonSerialize` i can get the data from directories
```
$files->filter(fn (StorageAttributes $attributes) => $attributes->isDir())
        ->map(fn (StorageAttributes $attributes) => $attributes->jsonSerialize())
        ->toArray()
```
but using without `jsonSerialize`  as  `->map(fn (StorageAttributes  $attributes) => $attributes)` doesn't work

erikn69:
Fixed for **PHP-CS-Fixer**

Issue #1318, opened 2021-06-07
FilesystemReader interface missing config on read and readStream


KDederichs:
### Bug Report

It's actually a mix of bug report and feature request.
While searching to use SSE-C encryption with the AsyncAws adapter I noticed that you can set the required options on write but not on read.
As per S3 documentation you are required to transmit the required parameters on write AND on read though.

Which brings me to the bug report/feature request:
`FilesystemReader` needs to be able to pass a config object on at least `read` and `readStream` cause otherwise you'd be able to write encrypted files with flysystem that are unreadable due to missing parameters.

|    Q        |   A
|------------ | ------
| BC Break    |  yes? 
| Version     | 2.0.4

#### Summary

See above.

Also it doesn't break BC for anything IN the the adapters itself as you'd just update them to merge an empty array but people might experience issues with the interface on custom adapters so it's a possible BC break for them.

#### How to reproduce

Try to use SSE-C with AsyncAws

PR #1319, opened 2021-06-07
[2.x] Refactor read interfaces to allow config

KDederichs:
Add optional config parameter to read interfaces to cover cases like descibed in #1318

### DIFF
--- a/src/AdapterTestUtilities/ExceptionThrowingFilesystemAdapter.php
+++ b/src/AdapterTestUtilities/ExceptionThrowingFilesystemAdapter.php
@@ -66,18 +66,18 @@ class ExceptionThrowingFilesystemAdapter implements FilesystemAdapter
         $this->adapter->writeStream($path, $contents, $config);
     }
 
-    public function read(string $path): string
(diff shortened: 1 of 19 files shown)


PaolaRuby:
Breaks all custom adapters, maybe default options setter on Aws Adapter, or arg on constructor?

KDederichs:
You can't really set it for the whole adapter imo since it's possible for each object to have a different key.

KDederichs:
Another solution that won't break existing custom interfaces would be to maybe add a new 'EncryptedFilesystem' interface and then implement the encryption/decryption logic directly on the filesystem adapters that support it?
Maybe that would be a better solution.

KDederichs:
Ok I introduced that solution in #1325 so tell me which is the one you guys would prefer.

Nyholm:
Thank you for this PR. Im afraid I cannot merge this. Modifying the method signature of the interface is a BC break. 

Thank you @PaolaRuby for the review. I agree with you. 
I'll close this PR and review the alternative in #1325.

PR #1320, opened 2021-06-10
Allow to mount/unmount filesystems

ElGigi:
Expose methods of MountManager class:

- `MountManager::mountFilesystem(string $key, FilesystemOperator $filesystem): void` Mount new filesystem in the manager
- `MountManager::unmountFilesystem(string $key): void` Unmount filesystem from manager
- `MountManager::hasFilesystem(string $key): bool` Know if a filesystem is mounted

Alternative naming proposal:

- `MountManager::mount(...)`
- `MountManager::unmount(...)`
- `MountManager::isMounted(...)`

### DIFF
--- a/src/MountManager.php
+++ b/src/MountManager.php
@@ -249,9 +249,37 @@ class MountManager implements FilesystemOperator
         }
     }
 
-    private function mountFilesystem(string $key, FilesystemOperator $filesystem): void
+    /**
+     * Mount filesystem.
+     *
+     * @param string $key
+     * @param FilesystemOperator $filesystem
+     */
+    public function mountFilesystem(string $key, FilesystemOperator $filesystem): void
     {
         $this->filesystems[$key] = $filesystem;
+    }
+
+    /**
+     * Unmount filesystem.
+     *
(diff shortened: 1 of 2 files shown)


frankdejonge:
This creates unpredictable usage. There have been requests for this previously, but I have not had any good reasons apart from 'because you can' type of arguments. If you have a case that cannot be solved by constructing a new mount manager, please let me know.

ElGigi:
In my case, I mount framework paths, to find for example the user configurations.
In a second time, i need to mount the new paths from user config.

It’s right, I can construct a new manager, but I’m not sure about memory consumption optimization.

Or if it's possible to change visibility of private methods of MountManager to protected, in my case, I will can make dynamic mounts, without mount unnecessaries paths in a context.

Nyholm:
> It’s right, I can construct a new manager, but I’m not sure about memory consumption optimization.

Im closing this PR because of the reasons Frank mentioned and because you have a workaround. 

Feel free to continue the discussion with the exact memory overhead the workaround causes. I doubt that it is more than a few bytes.

PR #1324, opened 2021-06-22
Allow fetching extra metadata

### LINKED ISSUE #1323: Allow fetching of extraMetadata
### Feature Request

|    Q        |   A
|------------ | ------
| New Feature | yes
| BC Break    | yes

#### Scenario / Use-case

Many backends provide more metadata. We currently populate `FileAttributes` with the data, but don't provide a way to access it. While we can't provide interfaces into any of that data in a way that would allow cross adapter use, we can just provide that data. That way when using the Flysystem, if you either know the specific filesystem adapter that will be used, or can use the metadata with a fallback, you can make for a more efficient system. For example, a few systems return the md5sum in the metadata, this can be used to check if a file matches an existing file before updating the file. It's easy enough to check the returned metadata for that field, and if not present, then code a fallback method (it may be more costly to replace the file if it hasn't changed).

I believe this would be a BC due to the addition of methods to the interfaces.

#### Summary

Expose the `extraMetadata` attribute to the end user.

timwsuqld:
Fix #1323 

Allow fetching of extra metadata so that more customised code can be written that isn't adapter specific

### DIFF
--- a/src/AsyncAwsS3/AsyncAwsS3Adapter.php
+++ b/src/AsyncAwsS3/AsyncAwsS3Adapter.php
@@ -254,6 +254,17 @@ class AsyncAwsS3Adapter implements FilesystemAdapter
         return $attributes;
     }
 
+    public function extraMetadata(string $path): FileAttributes
+    {
+        $attributes = $this->fetchFileMetadata($path, FileAttributes::ATTRIBUTE_EXTRA_METADATA);
+
+        if (null === $attributes->extraMetadata()) {
+            throw UnableToRetrieveMetadata::extraMetadata($path);
+        }
+
+        return $attributes;
+    }
+
     public function listContents(string $path, bool $deep): iterable
     {
         $prefix = trim($this->prefixer->prefixPath($path), '/');
--- a/src/AwsS3V3/AwsS3V3Adapter.php
+++ b/src/AwsS3V3/AwsS3V3Adapter.php
@@ -349,6 +349,17 @@ class AwsS3V3Adapter implements FilesystemAdapter
         return $attributes;
     }
 
+    public function extraMetadata(string $path): FileAttributes
+    {
+        $attributes = $this->fetchFileMetadata($path, FileAttributes::ATTRIBUTE_EXTRA_METADATA);
+
+        if ($attributes->extraMetadata() === null) {
+            throw UnableToRetrieveMetadata::extraMetadata($path);
+        }
+
+        return $attributes;
+    }
+
     public function listContents(string $path, bool $deep): iterable
     {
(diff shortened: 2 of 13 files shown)


Nyholm:
Adding methods to an interface is a BC break. I cannot merge this.

4c0n:
@Nyholm @frankdejonge He also explains that in the issue: https://github.com/thephpleague/flysystem/issues/1323

I would kind of like this feature as well, as after "upgrading" to version 2 I lost the ability of checking the S3 object version metadata amongst other metadata, are there any plans on starting a new branch/work on a new major version so we can overcome the BC break?

We could also go the less pretty route and not update the interface, but add the methods to the adapters for example.

Let's talk about this :smile:

m-vo:
@Nyholm Would you be fine with adding a method to the `Filesystem` like so?

```php
public function extraMetadata(string $path): array
{
    if (!method_exists($this->adapter, 'extraMetadata')) {
        return [];
    }

    return $this->adapter->extraMetadata($this->pathNormalizer->normalizePath($path))->extraMetadata();
}
```

With the adapter pattern you're typically only interfacing with the single default implementation, effectively making the adapters a black box. I don't see how your're otherwise able to get the metadata that the `StorageAttributes` classes can already hold.

An alternative with less magic would be to add a new interface for adapters supporting an `extraMetadata` function and test against this interface in the `Filesystem`.

PaolaRuby:
@m-vo make a PR

m-vo:
> @m-vo make a PR

Not if there is no intention for it to get merged?

fredsal:
This would have been great for v3

fredsal:
??

aedart:
Ah... Can we perhaps re-work this to support additional meta-data, without having to introduce a breaking change?
A possible solution would be to extract the the `extraMetadata()` into its own interface, so that adapters that support it can chose to implement it, e.g.

```php
interface HasMetaData
{
    /**
     * @throws UnableToRetrieveMetadata
     * @throws FilesystemException
     */
   public function extraMetadata(string $path): FileAttributes;
}
```

As for the filesystem implementation, it does not have to beak it's interface, but simply offer a public method:

```php
class Filesystem implements FilesystemOperator
{
    // ...previous not shown ....

    public function extraMetadata(string $path): array
    {
        $adapter = $this->adapter;
        
        if ($adapter instanceof HasMetaData) {
            return $this->adapter->extraMetadata($this->pathNormalizer->normalizePath($path))->extraMetadata();    
        }
        
        // Default to empty array
        return [];
    }
}
```

Then, perhaps, in a future major version you could choose to merge the "HasMetaData" interface into the appropriate interfaces.

PR #1325, opened 2021-06-22
[2.x] Encrypted Filesystem Adapter

KDederichs:
An alternative solution to #1319 (#1318)which would not break BC by introducing a new `EncryptedFilesystemAdapter` interface which allows operation on encrypted files on filesystems that support it.

Depending on which solution you think is better I'll close this or the other PR and focus on that one.

### DIFF
--- a/src/AsyncAwsS3/AsyncAwsS3Adapter.php
+++ b/src/AsyncAwsS3/AsyncAwsS3Adapter.php
@@ -15,6 +15,7 @@ use AsyncAws\SimpleS3\SimpleS3Client;
 use Generator;
 use League\Flysystem\Config;
 use League\Flysystem\DirectoryAttributes;
+use League\Flysystem\EncryptedFilesystemAdapter;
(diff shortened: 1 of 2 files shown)


frankdejonge:
Hi @KDederichs,

Might I suggest an alternative method for this. There is an interface what you can use to wrap the S3 client, you can use this to decorate the original client which allows you to transparently add encryption similar to how you do now. The benefit of this is that the current implementation does not need to incur any complexity and secondly you have a general encrypted S3 client. If there is anything preventing you from using that route, I'd be happy to explore options with you.

KDederichs:
What interface would that be?
The problem in the first place is the existing implementation allows for saving it in encrypted form using write/writeStream with the right parameters but there is no way to decrypt it again since read can not accept parameters...

frankdejonge:
That would be the interface provided by the AWS SDK named `Aws\S3\S3ClientInterface`.

PR #1330, opened 2021-07-18
Fixes #1313: Add a sorting function to the directory listing.

### LINKED ISSUE #1313: Directory listing is not sorted
### Question

I started testing the upgrade from 1.0 to 2.0.8. after allot of code updates I got it working, only one thing I can't fix is the change of the directory listing. 

It looks like a random sorting of the filenames, while in the old version it was [a - z] and now sometimes it's [z - a] or [a - z] or random.
I need it to be [a - z] because we import data in that order, because of the updates are in that order presented on, lets say, a sFtp.

Is this a bug or can this be updated.

Thank you in advance for your answers.

frankdejonge:
Directory iterators and other PHP local filesystem methods do not return sorted responses. In v1 we would always sort the output. Sorting is not compatible with generators, so it cannot be the default. This PR introduces a sortByPath method to the directory listing to sort it, which also turns the directory listing into a fully accumulated array response, which takes up more memory. This is the trade-off that people will have to make.

This will fix #1313

### DIFF
--- a/src/DirectoryListing.php
+++ b/src/DirectoryListing.php
@@ -55,7 +55,7 @@ class DirectoryListing implements IteratorAggregate
         $listing = $this->toArray();
 
         usort($listing, function (StorageAttributes $a, StorageAttributes $b) {
-            return strcasecmp($a->path(), $b->path());
+            return $a->path() <=> $b->path();
         });
 
         return new DirectoryListing($listing);
--- a/src/DirectoryListing.php
+++ b/src/DirectoryListing.php
@@ -50,6 +50,17 @@ class DirectoryListing implements IteratorAggregate
         return new DirectoryListing($generator);
     }
 
+    public function sortByPath(): DirectoryListing
+    {
+        $listing = $this->toArray();
+
+        usort($listing, function (StorageAttributes $a, StorageAttributes $b) {
+            return strcasecmp($a->path(), $b->path());
+        });
+
+        return new DirectoryListing($listing);
+    }
+
     /**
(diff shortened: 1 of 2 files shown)


Nyholm (approved):
Since you opted to use `strcasecmp()` instead of the `<=>`, I assume you want case to be ignored. That is currently not reflected in the test. 

Could you update the test to include `/b/c/A.txt` too?

toaomatis:
I agree with @Nyholm that the spaceship operator would be desirable for me as well.
Maybe add an (default) argument to the function, so the user can choose whether to use the case-sensitive or case-insensitive directory listing. 

We are using a Linux-bases OS for our software, zo case-sensitivity is a big thing!

Nyholm:
I’m fine with both case-sensitive and case-insensitive. But to make it optional seams to be too much. 

I have no preference, I just said we should add a test for whatever is decided.

frankdejonge:
For this one I looked at what we did with Flysystem 1.x and copied that so it would have the same behaviour. I do not actually have very strong preferences. Perhaps we can look at how other filesystems (or views thereof) do.

toaomatis:
@frankdejonge there is no need for further investigation. I appreciate the work already done on this one.

For me the case-insensitive sorting is fine (for now ;) ). When I run into problems in the future I'll open a new Issue / PR for a case-sensitive compare.

frankdejonge:
Converted it to use the spaceship operator 👍

toaomatis:
Thanks!

Issue #1338, opened 2021-08-03
UTF-8 activation fail in some server


baptistemeunier:
### Bug Report

Hi,

On some server the comand `OPTS UTF8 ON` respond with `202 UTF8 mode is always enabled. No need to send this command.`.

If you request UTF8 in your adapter config, the connection will fail for those server.

#### How to reproduce

* Find a server who return `202 UTF8 mode is always enabled` (maybe filezilla-server ?).
* Set `utf8` to `true` in your adapter config
* Try a connection

What's happen : 
* Conection failled : `Could not set UTF-8 mode for connection: localhost::21`

What's expected :
* Connection success (because UTF8 is already set by server)

frankdejonge:
Both PRs have been merged and released, so this issue is now fixed. Thank you for reporting it and sending in the PRs.

PR #1340, opened 2021-08-03
[2.x] Allow utf8 connections when server is already utf8 

baptistemeunier:
* Allow 202 response code as valid UTF-8 activation
* Add tests to match this new configuration


baptistemeunier:
Relative to #1338

PR #1349, opened 2021-08-25
Add mount manager docs to v2

Seldaek:
Carried over docs from v1, but using (hopefully correctly) the v2 APIs.


frankdejonge:
Thank you 👍

Issue #1356, opened 2021-09-21
[Google Cloud Storage] It should be possible not to use the visibility handler/visibility settings.


jkuma:
### Bug Report

Report of issue (https://github.com/thephpleague/flysystem-google-cloud-storage/issues/4) from `thephpleague/flysystem-google-cloud-storage`

|    Q        |   A
|------------ | ------
| BC Break    | no
| Version     | 2.0.6

#### Summary

GCP storage buckets now suggest not to use ACL but bucket wide permissions when possible. Unfortunately trying to upload a file to such bucket throws an error:

Cannot insert legacy ACL for an object when uniform bucket-level access is enabled. Read more at https://cloud.google.com/storage/docs/uniform-bucket-level-access

That was a problem for me, I ended up making a copy of the adapter since it was unable to extend it and just commented out the following line:

'predefinedAcl' => $this->visibilityHandler->visibilityToPredefinedAcl($visibility),

it would be good to actually allow null $defaultVisibility, $dafaultAcl with null works when doing uploads.

#### How to reproduce

When using GoogleCloudStorage adapter and enabling uniform bucket-level access.

PR #1357, opened 2021-09-21
Issue #1356: It should be possible not to use the visibility handler/visibility settings.

jkuma:
### Context

See related issue #1356 

The [Uniform bucket-level access](https://cloud.google.com/storage/docs/uniform-bucket-level-access) doesn't accept the predefinedAcl option to be set when performing actions on objects. I tried all default acl options, even null, and it ends up with following error: _Cannot insert legacy ACL for an object when uniform bucket-level access is enabled_

### Proposed solution

In order to ensure backward compatibility, it should be nice to have a new constant to bypass this settings.


frankdejonge:
@jkuma thank you for your contribution, it's released as 2.3.0

Issue #1358, opened 2021-09-21
[Google Cloud Storage] Impossible to set an empty prefix


jkuma:
### Bug Report

This is a report of (https://github.com/thephpleague/flysystem-google-cloud-storage/issues/3) from `thephpleague/flysystem-google-cloud-storage`

|    Q        |   A
|------------ | ------
| BC Break    | no
| Version     | 2.0.6

#### Summary

There is no way to list first level (root) contents of bucket, as prefix is always overridden with the slash '/'.

#### How to reproduce

When using Google Cloud Storage adapter and setting an empty prefix.

Explanation with example:
googleapis/google-cloud-php#910

PR #1359, opened 2021-09-21
Issue #1358: Impossible to set an empty prefix for Google Cloud Storage adapter.

jkuma:
### Context

Please have a look at issue #1358.

### Proposed solution.

Solution proposed is to not automatically set **prefix** option when listing content. This option is only set when a prefix exists.


frankdejonge:
@jkuma thank you for your contribution, it's released as 2.3.0

PR #1368, opened 2021-10-18
Fix: get the correct visibilities when listing the contents of the filesystem

meyerbaptiste:
According to the [documentation](https://www.php.net/manual/en/splfileinfo.getperms#example-3683) of `SplFileInfo::getPerms()`, the permissions were not parsed correctly.
This resulted in incorrect visibilities when listing the contents of the filesystem.

### DIFF
--- a/src/Local/LocalFilesystemAdapter.php
+++ b/src/Local/LocalFilesystemAdapter.php
@@ -104,7 +104,7 @@ class LocalFilesystemAdapter implements FilesystemAdapter
         );
         error_clear_last();
 
-        if (($size = @file_put_contents($prefixedLocation, $contents, $this->writeFlags)) === false) {
+        if (@file_put_contents($prefixedLocation, $contents, $this->writeFlags) === false) {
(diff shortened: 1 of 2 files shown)


erikn69:
Looks like write uses `FILE:0777,0755;` `DIR:0777,0755` intead of `FILE:0644,0600;` `DIR:0755,0700` on [PortableVisibilityConverter.php](https://github.com/thephpleague/flysystem/blob/2.x/src/UnixVisibility/PortableVisibilityConverter.php)
```php
There was 1 failure:

1) League\Flysystem\Local\LocalFilesystemAdapterTest::retrieving_visibility_while_listing_directory_contents
Failed asserting that two strings are equal.
--- Expected
+++ Actual
@@ @@
-'public'
+'private'

/home/runner/work/flysystem/flysystem/src/Local/LocalFilesystemAdapterTest.php:313
/home/runner/work/flysystem/flysystem/vendor/phpunit/phpunit/phpunit:98
```

PR #1371, opened 2021-10-26
Fixes #1370 create PrefixedFilesystem

### LINKED ISSUE #1370: Remove all prefixer implementation from Adapters, move it to FilesystemPrefixedDecorator.
### Bug Report

It's very annying to have prefix implemented per adapter. 
The use-case is: on file write => create CloudFront Invalidation.
I'm doing that by creating Decorator that Decorates FilesystemOperator, and on write methods - do API call to create invalidation.

The problem is that prefix is implemented in underlying adapter. So if I have AwsS3 adapter with prefix `xxx`, and I do

`$filesystem->write('some-file.jpg');`

my decorator doesn't know that this is in "directory" `xxx` and sends wrong invalidation request to CloudFront.

Better way to do this is by creating PrefixedFilesystemDecorator. This way solves the issue by having

- PrefixedFilesystemDecorator
- - MyDecoratorWithInvalidator
- - - Real Filesystem Operator.

Rather that simple approach, currently i'm doing a lot of if checks with `CompilerPass` and alot of if and extract prefix, depending on the adapter. 

Furthermore it will reduce the complexity of each Adapter and will DRY out the prefix implementation from each of them and will have consistent behaviour accross all adapters, and also more SOLID

|    Q        |   A
|------------ | ------
| BC Break    | yes
| Version     |  ALL

Warxcell:
TODOs:
- Rework the bundle to add consistent behaviour to add prefix.

For LocalFilesystemAdapter and FtpAdapter I think it's fine to have internally prefixer.

### DIFF
--- a/src/AsyncAwsS3/AsyncAwsS3Adapter.php
+++ b/src/AsyncAwsS3/AsyncAwsS3Adapter.php
@@ -18,6 +18,7 @@ use League\Flysystem\DirectoryAttributes;
 use League\Flysystem\FileAttributes;
 use League\Flysystem\FilesystemAdapter;
 use League\Flysystem\PathPrefixer;
+use League\Flysystem\PrefixedFilesystem;
 use League\Flysystem\StorageAttributes;
 use League\Flysystem\UnableToCheckFileExistence;
 use League\Flysystem\UnableToCopyFile;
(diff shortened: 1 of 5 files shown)


frankdejonge:
Closing this, I do not like the DX of this. In the future, it helps to discuss before sending in a PR to prevent having a PR from being closed after putting in the work.

Warxcell:
What's wrong with DX? I think it's much better like that. It reduces individual adapter complexity, reduces test complexity (no need to test prefix over and over again, individually). Adapters have 2 concerns now - prefixing and actual storage reads/writes - otherwise they would have single concern. Furthermore if some adapter doesn't have implemented prefix, you can always have prefix, because of the decorator at file system level.

PR #1375, opened 2021-11-05
FilesystemAdapterTestCase: always close stream handlers after use

Slamdunk:
I had some issues with this two dandling streams left opened within the tests.


frankdejonge:
Thank you 👍

PR #1396, opened 2022-01-05
Added mup options to config array

ToshY:
Original issue: https://github.com/thephpleague/flysystem-aws-s3-v3/issues/269


frankdejonge:
Hi @ToshY, any reason why this array is expressed separately from the normal options? I'm askind because it seems to always merged be together.

ToshY:
@frankdejonge The only reason I separated them was because I thought it makes it clearer that those options are specifically for multipart uploads. I have no objections to just merging them into the normal options.

frankdejonge:
@ToshY thanks for letting me know, I'll merge and release soon 👍

PR #1407, opened 2022-01-17
Feature: add azure blob storage

frankdejonge:


keatliang2005:
imo this should be dedicated adaptor package , and should not be directly merge at flysytem itself 

example do it something like this instead
https://github.com/thephpleague/flysystem-aws-s3-v3/tree/3.x

frankdejonge:
Hi @keatliang2005, those other adapters are also maintained from this repo, they are published as a sub-split to their own repo for distribution.

keatliang2005:
ah after peaking into 3.x branch now i see clearer. i was upgrading to Laravel 9, and it's using flysystem v3 and azure blob storage adaptor is missing for v2|v3 

now i just wait this to be merged into the v3 branch 

thanks for the awesome work there.

jbraband:
Is there anything I can do to help with any remaining tasks on this PR? 
With all checks having passed, is merging it just a matter of time?

jbraband:
thanks for the merge @frankdejonge !

frankdejonge:
👍 I'll release it later today

Issue #1415, opened 2022-01-31
Undefined constant Servdebt\\SlimCore\\Filesystem\\S3\\AsyncAwsS3Adapter::EXTRA_METADATA_FIELDS


jupitern:
### Bug Report

I get the error bellow after update to 3.0.* when calling function fileSize() with AsyncAwsS3Adapter

Undefined constant Servdebt\\SlimCore\\Filesystem\\S3\\AsyncAwsS3Adapter::EXTRA_METADATA_FIELDS
at vendor/league/flysystem-async-aws-s3/AsyncAwsS3Adapter.php 
line 444

```php
foreach (static::EXTRA_METADATA_FIELDS as $field) {
```

works changing to self::

frankdejonge:
@jupitern are you extending that adapter by any chance?

jupitern:
yes. forgot to mention that. sorry.

christoph-kluge:
Same here but custom extended class. The temporary solution from @jupitern works like a charm.

In case the const stays private maybe moving the accesor from `static` to `self`?

```
- foreach (static::EXTRA_METADATA_FIELDS as $field) {
+ foreach (self::EXTRA_METADATA_FIELDS as $field) {
```

PR #1416, opened 2022-02-01
Use self instead of static

Nyholm:
This might be a fix for #1415

But Im not sure..


frankdejonge:
@Nyholm we could do that, or allow people to extend. I'd have to write a BC policy for that, stating it's "up to them". In that case, instead of switching this, we'd actually make that const protected. Not sure about that tho. What do you think?

frankdejonge:
I've fixed this by altering the visibility instead. This way people can also change the fields they want to populate.

PR #1301, opened 2021-04-26
SftpAdapter: fix issue with `$connection->rawlist()` returning false

dmaicher:
Fixes https://github.com/thephpleague/flysystem/issues/1300

PR #1428, opened 2022-03-02
SftpAdapter fix - check if the listing is not an array

X-Coder264:
`$connection->rawlist` returns `array|false`. There was a check missing before trying to foreach `$listing` which resulted in 
 
`foreach() argument must be of type array|object, bool given` error.

The same check is already present in the PhpseclibV2 sftp adapter.

### DIFF
--- a/src/PhpseclibV3/SftpAdapter.php
+++ b/src/PhpseclibV3/SftpAdapter.php
@@ -274,6 +274,10 @@ class SftpAdapter implements FilesystemAdapter
         $location = $this->prefixer->prefixPath(rtrim($path, '/')) . '/';
         $listing = $connection->rawlist($location, false);
 
+        if (false === $listing) {
+            return [];
+        }
+
         foreach ($listing as $filename => $attributes) {
             if ($filename === '.' || $filename === '..') {
                 continue;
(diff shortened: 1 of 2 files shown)


dmaicher:
For PhpseclibV2 I [added a test](https://github.com/thephpleague/flysystem/pull/1301/files#diff-08d7c8b2e8002044cd05d7654dd4e20e6154617d7dc594b77ab7c435ec747987R200). Might be good to add one for this change too?

X-Coder264:
@dmaicher True, test added. Thanks for pointing me to that test for the PhpseclibV2 adapter.

src/PhpseclibV3/SftpAdapter.php
```diff
@@ -274,6 +274,10 @@ public function listContents(string $path, bool $deep): iterable
         $location = $this->prefixer->prefixPath(rtrim($path, '/')) . '/';
         $listing = $connection->rawlist($location, false);
 
+        if (false === $listing) {
+            return [];
```
dmaicher:
```suggestion
            return;
```

seems better since we are returning a `\Generator`?

frankdejonge: agreed

dmaicher (approved):
:+1:

PR #1432, opened 2022-03-11
Use prefixDirectoryPath in deleteDirectory of AsyncAwsS3Adapter

Seldaek:


frankdejonge:
@Seldaek cheers, will wait for CI and tag right after 👍

Seldaek:
No rush, it's just a cosmetic change I think, functionally seems to be the same.

frankdejonge:
Ah yeah, that's correct. Anywho, still pushing out a fix soon because of an SFTP bugfix so it'll go out regardless! Thanks for this :)

PR #1437, opened 2022-04-01
[BUGFIX] Ensure top-level directory is not returned in directory listing.

frankdejonge:


jesseleite:
Thank you again for your time on this @frankdejonge, did not expect it so fast. Have a great week ❤️

PR #1444, opened 2022-04-13
Fix #1397

### LINKED ISSUE #1397: Directory listing fails with hardcoded options
### Bug Report

|    Q        |   A
|------------ | ------
| BC Break    | no
| Version     | 1.1.9

#### Summary

After successfully connecting to an FTP server, I am unable to get the listing of files in the directory. When I manually go into the code and clear the options variable in method `ftpRawlist` the listing comes back successfully.

These options are hardcoded in `listDirectoryContents` on line 505 : `$options = $recursive ? '-alnR' : '-aln';`

I am using flysystem through Laravel (8.70.2) storage. 

#### How to reproduce

I am not sure if this is unique to this particular server, or what is causing the issue. When I connect to the FTP server (220-FileZilla Server 1.1.0) manually and try those command they error as well:

```
ftp> ls -a
227 Entering Passive Mode (209,249,197,15,73,115)
450 Couldn't open the file
ftp> ls -aln
227 Entering Passive Mode (209,249,197,15,76,133)
450 Couldn't open the file
ftp> ls -ln
227 Entering Passive Mode (209,249,197,15,77,221)
450 Couldn't open the file
```

If I do an `ls` with no arguments, the listing comes back successfully.

This update fixes it for me, I can submit as a PR if this makes sense:
```
    /**
     * @return bool
     */
    protected function isFileZillaServer()
    {
        $response = ftp_raw($this->connection, 'SYST');

        return stripos(implode(' ', $response), 'FileZilla') !== false;
    }

    /**
     * The ftp_rawlist function with optional escaping.
     *
     * @param string $options
     * @param string $path
     *
     * @return array
     */
    protected function ftpRawlist($options, $path)
    {
        $connection = $this->getConnection();

        if ($this->isPureFtpd) {
            $path = str_replace([' ', '[', ']'], ['\ ', '\\[', '\\]'], $path);
        }

        if ( $this->isFileZillaServer()) {
            $options = '';
        }

        return ftp_rawlist($connection, $options . ' ' . $this->escapePath($path));
    }
```

marvinlinthorst:


src/Ftp/FtpAdapter.php
```diff
@@ -141,6 +146,18 @@ private function isPureFtpdServer(): bool
...
+   
+    private function isFileZillaServer(): bool
+    {
+        if ($this->isFileZillaServer !== null) {
+            return $this->isFileZillaServer;
+        }
+        
+        $response = ftp_raw($this->connection, 'SYST');
+
+        return $this->isFileZillaServer = stripos(implode(' ', $response), 'FileZilla') !== false;
```
astehlik, line 158: The autodetection is a great idea. I'm not sure how this needs to be communicated because it might be a breaking change for oder FileZilla server versions that supported options.

Also: a check for pyftpdlib could be added if possible (see https://github.com/thephpleague/flysystem/issues/1397#issuecomment-1026137563).

src/Ftp/FtpAdapter.php
```diff
@@ -69,6 +69,11 @@ class FtpAdapter implements FilesystemAdapter
      * @var bool|null
      */
     private $isPureFtpdServer;
+    
+    /**
+     * @var bool|null
+     */
+    private $isFileZillaServer;
```
astehlik, line 76: Since other servers seem to be affected by this (like pyftpdlib, see https://github.com/thephpleague/flysystem/issues/1397#issuecomment-1026137563) a more generic option name like `useRawListOptions` (which defaults to `true`) could be better.

src/Ftp/FtpAdapter.php
```diff
@@ -141,6 +146,18 @@ private function isPureFtpdServer(): bool
 
         return $this->isPureFtpdServer = stripos(implode(' ', $response), 'Pure-FTPd') !== false;
     }
+   
+    private function isFileZillaServer(): bool
```
astehlik, line 150: Similar as above for the field name, e.g. `isServerSupportingListOptions()`

astehlik:
Also: the check at 

https://github.com/thephpleague/flysystem/blob/12f43f5800b3e92dca3d75dab535d210bc02bf53/src/Ftp/FtpAdapter.php#L378

needs to be overwritten to `true`, when the server does not support LIST options.

PR #1445, opened 2022-04-13
made flags configurable for file_put_contents

Virtuoid:
Hello,

in the new version you guys are trying to lock files before writing.

The NFS (Versions 2 and 3) protocol does not support file locking, but the NFS environment supports an ancillary protocol called NLM, which originally stood for "Network Lock Manager." When an NFS filesystem on an NFS client gets a request to lock a file, instead of an NFS remote procedure call, it generates an NLM remote procedure call.

To fix this our purpose is to allow clients to choose if use lock flag or not.

This requires a change also in /Illuminate/Filesystem/FilesystemManager.php adding the 'write_flag' to the allowed configurations.

Hope this helps, regards

### DIFF
--- a/src/Config.php
+++ b/src/Config.php
@@ -10,6 +10,7 @@ class Config
 {
     public const OPTION_VISIBILITY = 'visibility';
     public const OPTION_DIRECTORY_VISIBILITY = 'directory_visibility';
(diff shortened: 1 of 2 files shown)


frankdejonge:
The flag is configurable in the constructor, this PR is not needed as far as I'm concerned.

PR #1446, opened 2022-04-14
Avoid type errors when public key is not retrieved

lcobucci:
When dealing with unstable connections, phpseclib may return `false` when retrieving the public key.  That causes `null` to be passed down to `base64_decode()` and triggers a TypeError due to strict types.

This exits early, preventing triggering the problem.
However, testing is a bit tricky since we need an unstable connection (or stub things).

### DIFF
--- a/src/PhpseclibV2/SftpConnectionProvider.php
+++ b/src/PhpseclibV2/SftpConnectionProvider.php
@@ -143,7 +143,12 @@ class SftpConnectionProvider implements ConnectionProvider
             return;
         }
 
-        $publicKey = $connection->getServerPublicHostKey() ?: 'no-public-key';
+        $publicKey = $connection->getServerPublicHostKey();
+
+        if ($publicKey === false) {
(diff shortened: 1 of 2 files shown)


src/PhpseclibV2/SftpConnectionProvider.php
```diff
@@ -143,7 +143,12 @@ private function checkFingerprint(SFTP $connection): void
             return;
         }
 
-        $publicKey = $connection->getServerPublicHostKey() ?: 'no-public-key';
+        $publicKey = $connection->getServerPublicHostKey();
+
+        if ($publicKey === false) {
+            throw UnableToEstablishAuthenticityOfHost::becauseTheAuthenticityCantBeEstablished($this->host);
```
lcobucci, line 149: I feel bad about not providing any tests here (which would have caught the issue).
Having this method private and always instantiating `SFTP` make it slightly difficult to inject a stubbed connection to make sure we cover this behaviour.

Would love having ideas that don't involve using something like https://github.com/shopify/toxiproxy to mess up with the connection we use for tests.

PR #1451, opened 2022-04-21
Fix SFTP retry mechanism

### LINKED ISSUE #1450: [SFTP Adapter (phpseclib v3)] Retry mechanism doesn't work
### Bug Report

|    Q        |   A
|------------ | ------
| BC Break    | no
| Version     | 2.4.4

#### Summary

In phpseclibv3 we have exceptions when dealing with unreliable connections, which isn't handled well by `SftpConnectionProvider` and cause the retry mechanism to be bypassed.

https://github.com/thephpleague/flysystem/blob/29eb78cac0be0c22237c5e0f6f98234d97037d79/src/PhpseclibV3/SftpConnectionProvider.php#L114-L153

#### How to reproduce

Testing and reproducing this requires a flaky connection, here's a snippet that isolates the problem: https://3v4l.org/KmXsL

lcobucci:
As of phpseclib v3 `RuntimeException` objects are thrown, causing the whole retry logic to be bypassed.

This makes sure to re-attempt to connect when timeout or connection issues happen.

FIxes #1450

### DIFF
--- a/src/PhpseclibV3/SftpConnectionProvider.php
+++ b/src/PhpseclibV3/SftpConnectionProvider.php
@@ -4,6 +4,7 @@ declare(strict_types=1);
 
 namespace League\Flysystem\PhpseclibV3;
 
+use League\Flysystem\FilesystemException;
 use phpseclib3\Crypt\Common\AsymmetricKey;
 use phpseclib3\Crypt\PublicKeyLoader;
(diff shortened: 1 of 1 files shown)


lcobucci:
I'll plan sometime to add tests to cover the issues I fixed 👍

PR #1452, opened 2022-04-21
Verify SFTP retries when dealing with peer resets 

lcobucci:
This verifies that Flysystem SFTP adapter goes through all the configured tries before throwing an `UnableToConnectToSftpHost` exception.

Reverting the fix applied in https://github.com/thephpleague/flysystem/pull/1451 makes the test to fail due to an unexpected `RuntimeException`.

Applying this for phpseclib v2 is a bit tricky, though, due to the lack of knowledge on what has happened in the flow (e.g. `false` is returned when a connection cannot be established or when an authentication error happened). To make it work, the adapter for that version would need to verify if there's a connection or not all the time - which doesn't sound very good to me.

### DIFF
--- a/src/AdapterTestUtilities/ToxiproxyManagement.php
+++ b/src/AdapterTestUtilities/ToxiproxyManagement.php
@@ -1,0 +1,75 @@
+<?php
+
+declare(strict_types=1);
+
+namespace League\Flysystem\AdapterTestUtilities;
+
+use GuzzleHttp\Client;
+
+/**
+ * This class provides a client for the HTTP API provided by the proxy that simulates network issues.
+ *
+ * @see https://github.com/shopify/toxiproxy#http-api
+ *
+ * @phpstan-type RegisteredProxies 'ftp'|'sftp'|'ftpd'
+ * @phpstan-type StreamDirection 'upstream'|'downstream'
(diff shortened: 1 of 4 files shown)


lcobucci:
@frankdejonge I'm not sure how these changes affect the `FTP` (not `SFTP`) tests... is it possible we're dealing with flaky stuff (considering the other tests have passed)?

frankdejonge:
@lcobucci yeah, those FTP containers are very unstable. I've implemented automatic retry mechanisms, but even then sometimes it's not good enough... I'll rerun the failed ones.

src/AdapterTestUtilities/ToxiproxyManagement.php
```diff
@@ -0,0 +1,75 @@
+<?php
+
+declare(strict_types=1);
+
+namespace League\Flysystem\AdapterTestUtilities;
+
+use GuzzleHttp\Client;
+
+/**
+ * This class provides a client for the HTTP API provided by the proxy that simulates network issues.
```
lcobucci, line 10: This sentence can definitely be improved... I mean "provides (...) provided" is a little weird. Feel free to change it @frankdejonge, I won't be able to push new commits today

frankdejonge: 👍

PR #1467, opened 2022-05-18
Fix static call a private constant EXTRA_METADATA_FIELDS 

kayw-geek:
- Fix  issues https://github.com/thephpleague/flysystem-aws-s3-v3/issues/252
- Replace the old command `ListObject` https://docs.aws.amazon.com/AmazonS3/latest/API/API_ListObjectsV2.html


frankdejonge:
This fix was resolved a different way. Thank you for your contribution 👍

PR #1471, opened 2022-05-31
Local Adapter: Lock files during read when LOCK_EX flag is set

MichaelRoosz:
Currently the Local Adapter will lock files with the "LOCK_EX" flag by default during write.
However, during read locks are completely ignored.

This PR will set read (shared) locks during reads.
This will fix reading corrupted data while a write is happening.

I encountered this issue (corrupted files) while using the php-cache library, which is using flystem for read and write operations ( https://github.com/php-cache/cache/blob/master/src/Adapter/Filesystem/FilesystemCachePool.php#L19 ).

Note: in readStream() the lock is obtained but never released, but this should be fine, because it will be automatically released during "fclose()".

### DIFF
--- a/src/Adapter/Local.php
+++ b/src/Adapter/Local.php
@@ -179,6 +179,10 @@ class Local extends AbstractAdapter
         $location = $this->applyPathPrefix($path);
         $stream = fopen($location, 'rb');
 
+        if ($this->writeFlags & LOCK_EX && $stream !== false) {
+            flock($stream, LOCK_SH);
+        }
+
         return ['type' => 'file', 'path' => $path, 'stream' => $stream];
     }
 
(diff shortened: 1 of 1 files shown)


src/Adapter/Local.php
```diff
@@ -220,8 +224,24 @@ public function update($path, $contents, Config $config)
...
         $location = $this->applyPathPrefix($path);
+
+        if ($this->writeFlags & LOCK_EX) {
+            $fileHandle = fopen($location, 'rb');
+
+            if ($fileHandle === false) {
+                return false;
+            }
+
+            flock($fileHandle, LOCK_SH);
```
frankdejonge, line 235: a false return needs to be handled, resulting in the appropriate exception (UnableToReadFile)

PR #1482, opened 2022-06-09
fix: Moving and copying of directories using the InMemoryFilesystemAdapter

### LINKED ISSUE #1477: InMemoryFilesystemAdapter cannot handle moving of folders
### Bug Report

I am migrating currently from v1 to v3 and using in the tests the InMemoryFilesystemAdapter. It looks like the move function currently does only handle files and not directory. In a testcase we move folders around and this seems to fail.

|    Q        |   A
|------------ | ------
| BC Break    | no
| Version     | 3.0.20

#### Summary

Create a folder with the InMemoryFilesystemAdapter and try to move it.

#### How to reproduce

- Create a folder with a file in it
- Move that folder to somewhere else (rename)
- UnableToMove exception is thrown

aragon999:
Closes #1477 

Not sure if the handling of the errors is desired or not. Let me know if there should be something changed.

### DIFF
--- a/src/InMemory/InMemoryFilesystemAdapter.php
+++ b/src/InMemory/InMemoryFilesystemAdapter.php
@@ -19,6 +19,8 @@ use League\MimeTypeDetection\MimeTypeDetector;
 
 use function array_keys;
 use function rtrim;
+use function str_starts_with;
+use function strlen;
 use function strpos;
 
 class InMemoryFilesystemAdapter implements FilesystemAdapter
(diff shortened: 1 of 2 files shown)


shyim:
Hey @frankdejonge,

can we help you in some way to get this merged :) like adding more phpunit tests

tinect:
@frankdejonge could you please provide any feedback?

mitelg:
hey @frankdejonge 

can we do something to get this forward? :slightly_smiling_face: 

We have an adaption of this class for three years now in our code, just to fix this bug :see_no_evil: https://github.com/shopware/shopware/blob/trunk/src/Core/Framework/Adapter/Filesystem/MemoryFilesystemAdapter.php

PR #1488, opened 2022-06-20
Deprecation sftp v2 in favor v3

maxhelias:
Following https://github.com/thephpleague/flysystem-sftp/issues/132

If this is too early or unwanted please feel free to close these

### DIFF
--- a/phpunit.xml.dist
+++ b/phpunit.xml.dist
@@ -16,4 +16,9 @@
             </exclude>
         </whitelist>
     </filter>
+    <groups>
+        <exclude>
+            <group>legacy</group>
+        </exclude>
+    </groups>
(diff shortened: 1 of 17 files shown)


frankdejonge:
Hi @maxhelias,

If you can remote the error triggers then this is ok to merge, I don't like that style of deprecation. We could instead use a `@deprecated` class-level tag. Let me know if you want to contribute it, alternatively I can do this change myself. Either way is fine by me :)

maxhelias:
@frankdejonge I change for `@deprecated` tag, are you ok with this deprecated message ?

frankdejonge:
Thanks for this @maxhelias 👍

PR #1498, opened 2022-06-30
Reenable phpstan

jnoordsij:
When working on another PR, I noticed that phpstan is disabled in the current GitHub workflow. Since there were already some errors occuring on the current code state, I figured it might be useful to reenable this on the `3.x` branch with current PHP versions.


jnoordsij:
Fixed by https://github.com/thephpleague/flysystem/pull/1571

PR #1517, opened 2022-07-05
Adding `ContentMD5` option to `AVAILABLE_OPTIONS` array

buglinjo:
This proposed change is adding `ContentMD5` option to `AVAILABLE_OPTIONS` array. Without this option in the array it is impossible to upload file on file lock enabled S3.


buglinjo:
Also, would be nice `AVAILABLE_OPTIONS` not to be a constant and be user settable with setter.

PR #1523, opened 2022-07-14
Set correct permissions wrt. visibility for implicitly created directories

nagmat84:
`Local::ensureDirectory` should use the configured visibility if available. I discovered this problem while working on a Laravel 8 project which still uses Flysystem v1. This bug fix does probably not completely fix the problem, but only for writing file streams.
If a file is copied or moved a non-existing parent directory is sill created with `'public'` visibility, but this fix suffices for our project and fixing all possible cases seemed to be too much work.

### DIFF
--- a/src/Adapter/Local.php
+++ b/src/Adapter/Local.php
@@ -99,7 +99,7 @@ class Local extends AbstractAdapter
     {
         if ( ! is_dir($root)) {
             $umask = umask(0);
-	        $visibility = $config instanceof Config ? $config->get('visibility', 'public') : 'public';
+	        $visibility = $config instanceof Config ? $config->get('directory_visibility', 'public') : 'public';
 
             if ( ! @mkdir($root, $this->permissionMap['dir'][$visibility], true)) {
                 $mkdirError = error_get_last();
--- a/src/Adapter/Local.php
+++ b/src/Adapter/Local.php
@@ -99,7 +99,7 @@ class Local extends AbstractAdapter
     {
         if ( ! is_dir($root)) {
             $umask = umask(0);
-	        $visibility = $config === null ? 'public' : $config->get('visibility', 'public');
+	        $visibility = $config instanceof Config ? $config->get('visibility', 'public') : 'public';
 
             if ( ! @mkdir($root, $this->permissionMap['dir'][$visibility], true)) {
                 $mkdirError = error_get_last();
--- a/src/Adapter/Local.php
+++ b/src/Adapter/Local.php
@@ -95,12 +95,13 @@ class Local extends AbstractAdapter
      *
      * @throws Exception in case the root directory can not be created
      */
-    protected function ensureDirectory($root)
+    protected function ensureDirectory($root, ?Config $config = null)
     {
         if ( ! is_dir($root)) {
             $umask = umask(0);
+	        $visibility = $config === null ? 'public' : $config->get('visibility', 'public');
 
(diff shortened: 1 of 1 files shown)


src/Adapter/Local.php
```diff
@@ -95,12 +95,13 @@ public function __construct($root, $writeFlags = LOCK_EX, $linkHandling = self::
      *
      * @throws Exception in case the root directory can not be created
      */
-    protected function ensureDirectory($root)
+    protected function ensureDirectory($root, ?Config $config = null)
     {
         if ( ! is_dir($root)) {
             $umask = umask(0);
+	        $visibility = $config === null ? 'public' : $config->get('visibility', 'public');
```
szepeviktor:
```suggestion
	        $visibility = $config instanceof Config ? $config->get('visibility', 'public') : 'public';
```

szepeviktor: _Explicit is better than implicit._ :)

nagmat84: Are you maintainer of this package? I thought @frankdejonge is. I had a quick look on the rest of the file and the way how it is written seems to be in line with that. I understand your approach and style which is perfectly fine. Without doubt it is more explicit than what I have written. However, I am wondering whether you are competent for this package and whether your suggestion is authoritative for this package.

szepeviktor: Okay-okay.

I run away.... 🏃🏻‍♂️

nagmat84: Sorry, if you took offenses. This was not my intention. I feel flattered that you take interest into my work and I appreciate your inputs.

However, in my experience code reviews become unstructured and confusing quickly, if they are not kept restricted to those who eventually also make the verdict whether a PR gets accepted or not. In Germany, we have the saying "Zu viele Köche verderben den Brei". I believe it is "too many cooks spoil the broth" in English.

In particular, this is true if the maintainers have a different concept of what good code loos like than those who also take part in the review process voluntarily (even if it is with the best intentions). I don't think this will be the case here, because the PR is rather small and straight forward.

szepeviktor: > In particular, this is true if the maintainers have a different concept

My high school physics teacher used to say 👨🏻‍🏫 _Don't scratch your left ear with your right hand!_
So if you want to call a method on a nullable object, check whether it is an instance.

szepeviktor: ... I'm secretly hoping that Frank will agree.

nagmat84:
@frankdejonge What are the chances to get the PR out of the door soon?

src/Adapter/Local.php
```diff
@@ -95,12 +95,13 @@ public function __construct($root, $writeFlags = LOCK_EX, $linkHandling = self::
      *
      * @throws Exception in case the root directory can not be created
      */
-    protected function ensureDirectory($root)
+    protected function ensureDirectory($root, ?Config $config = null)
     {
         if ( ! is_dir($root)) {
             $umask = umask(0);
+	        $visibility = $config instanceof Config ? $config->get('visibility', 'public') : 'public';
```
frankdejonge:
```suggestion
	        $visibility = $config instanceof Config ? $config->get('directory_visibility', 'public') : 'public';
```

frankdejonge:
This is not a security related issue. Flysystem 1.x is in security issues only mode. Since this has been default behaviour, I'm not going to change it.

PR #1527, opened 2022-07-28
Enable use of $config option in MergeManager::move()

gravitiq-cm:
Updated MergeManager::move() to use the $config parameter, by passing it through to moveAcrossFilesystems()
Added $config parameter to MergeManager::moveAcrossFilesystems() and pass it through to MergeManager::copy()

### DIFF
--- a/src/MountManager.php
+++ b/src/MountManager.php
@@ -223,7 +223,7 @@ class MountManager implements FilesystemOperator
             $destinationPath,
             $source,
             $destination
-        ) : $this->moveAcrossFilesystems($source, $destination);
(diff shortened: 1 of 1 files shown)


gravitiq-cm:
Note: before this change I couldn't see a way to move a file to an S3 bucket without an ACL policy. I was getting this error: 

>     Code:    AccessControlListNotSupported
>     Message: The bucket does not allow ACLs"

But there was no way to set "visibility: private" when calling move(), hence the changes in this PR.

maxhelias (approved):

Issue #1528, opened 2022-08-03
[ZipArchive] Prefix issue under windows


Tofandel:
### Bug Report

I'm reopening https://github.com/thephpleague/flysystem-ziparchive/issues/20 here as that's a subsplit and not monitored (maybe consider removing the PR section and issues of the git project)

This is a follow up after some digging up of https://github.com/pierrejoye/php_zip/issues/32

This package adds a `/` prefix (leading slash) to all files it adds to the archive https://github.com/thephpleague/flysystem-ziparchive/blob/3.x/ZipArchiveAdapter.php#L51

This is not valid according to the [zip specs](https://pkware.cachefly.net/webdocs/casestudies/APPNOTE.TXT) this creates an issue where the zip content is not visible in the windows zip explorer

>    4.4.17 file name: (Variable)
> 
>        4.4.17.1 The name of the file, with optional relative path.
>        The path stored MUST NOT contain a drive or
>        device letter, or a leading slash.  All slashes
>        MUST be forward slashes '/' as opposed to
>        backwards slashes '\' for compatibility with Amiga
>        and UNIX file systems etc.  If input came from standard
>        input, there is no file name field.  
> 

This effectively makes the ZipArchiveAdapter unusable as it does not create a valid zip file

|    Q        |   A
|------------ | ------
| BC Break    | yes
| Version     | 3.2.0

#### How to reproduce

Use ZIpArchive to create an archive
Open the created Zip file, it will be empty in the Windows explorer

Matth--:
Same issues when trying to use `listContents('', true) with a zip that got downloaded somewhere. The unit tests seem to only cover leading slashes when creating and reading from a zip folder.

PhilETaylor:
I can confirm this bug. To replicate on a Mac just right click a file in finder and select "Compress", the resulting zip file when read with listContents produces a blank array. 

downgrading to 3.1.1 fixes the issue for me for now, 3.2.0 is unusable in production because of this.

frankdejonge:
Almost back from holiday, will attend to after.

PhilETaylor:
Holidays are important, absolutely no pressure for a fix for me (I just downgraded for now)! Don't feel any pressure at all please !

frankdejonge:
Version 3.2.1 was just released to address this, if it's fixed for you, please close the issue 🙏

PhilETaylor:
I can confirm that 3.2.1 has fixed the issue I was having.

[...]

PR #1530, opened 2022-08-05
[ZipArchive] Do not prefix directories when creating/reading an archive

Matth--:
The unit tests only  tested scenarios where archives where created and folder had leading slashes. This is not according to zip spec. This PR trims leading slashes in folder names when creating the folder and tries to read files without a prefixed slash after opening an archive.


Matth--:
This closes #1528

PhilETaylor:
This is already fixed in 3.2.1 which was released yesterday

Matth--:
@PhilETaylor The 3.2.1 branch still uses a prefixed `/` when creating an archive. This PR drops the prefixed `/`. 

Maybe I will still need to add some test cases to make sure that zip archives created with previous versions still work. There are however no real E2E tests with zip archives. So I might add that with some random archives

PhilETaylor:
Ah ok. My use case was reading only. Sorry

PR #1533, opened 2022-08-16
Fix prefix-related issues in WebDAV adapter

mtorromeo:
Hi,
using the prefix option with the WebDAV adapter currently has multiple issues and works only for simple ascii prefixes in paths that already exist on the destination server.

Basically, all paths are encoded BEFORE adding the prefix. This means that the prefix is never encoded, and a simple space character in the prefix is enough to generate errors for any operation.

The second issue is that the functionality for auto-creating the directories on the server does not add the prefix before iterating over its split components, and so it fails to create the directory tree when the prefix itself does not exist on the server.

Thanks


frankdejonge:
@mtorromeo thank you for your contribution!

PR #1537, opened 2022-08-22
Add method to get connection provider

ikudosi:


frankdejonge:
Flysystem is not a DI or Service Locator. You can construct your own connection provider, or use the one you've passed to the constructor.

PR #1543, opened 2022-09-04
feat: add `StaticInMemoryFilesystemAdapter`

kbond:
This can be useful when functionally testing requests with a `InMemoryFilesystemAdapter` that is reset between requests.

If acceptable, I can create a doc PR.

### DIFF
--- a/src/InMemory/StaticInMemoryFilesystemAdapter.php
+++ b/src/InMemory/StaticInMemoryFilesystemAdapter.php
@@ -1,0 +1,117 @@
+<?php
+
+namespace League\Flysystem\InMemory;
+
+use League\Flysystem\Config;
+use League\Flysystem\FileAttributes;
+use League\Flysystem\FilesystemAdapter;
+
+class StaticInMemoryFilesystemAdapter implements FilesystemAdapter
+{
+    /** @var array<string, InMemoryFilesystemAdapter> */
+    private static array $filesystems = [];
+
+    public function __construct(private string $name = 'default')
+    {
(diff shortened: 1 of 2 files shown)


frankdejonge:
Could you describe a scenario in which this static state, based on the same key is needed? I don't really understand the use-case. I do get the case for wanting to reset the in-memory store, but I wonder if adding the reset function to the in memory instance class wouldn't achieve that without the need for static/global state? Regardless of the use-case, the test suite seems to not cover the cases where writing to two separate instances of this class with the same key has cross-instance effects.

kbond:
> Could you describe a scenario in which this static state, based on the same key is needed? I don't really understand the use-case.

Suppose a Symfony app where you have two Flysystem filesystem services: _public_ and _private_. In your test environment, you'd like these to be in-memory to speed up your tests. The problem with just the current in-memory adapter is it would be reset between kernel reboots (making multiple requests in a functional test). With the _static_ in-memory adapter, you would create them using the _public_ and _private_ names. Now files created would be persisted between reboots. In your `tearDown()` method, you'd run `StaticInMemoryFilesystemAdapter::reset()` to ensure each test resets any files created.

Hope that describes it's use well enough. Maybe I'm missing another way to achieve (this is the solution I came up with to solve the problem).

> the test suite seems to not cover the cases where writing to two separate instances of this class with the same key has cross-instance effects.

Added a test to cover this.

frankdejonge:
@kbond thanks! That makes sense :)

PR #1544, opened 2022-09-04
feat: add `ReadonlyFilesystemAdapter`

kbond:
Allows wrapping an adapter to disallow any _write_ operations. Throws appropriate `FilesystemOperationFailed` if write operation attempted.

If acceptable, I can create a doc PR.


frankdejonge:
Tests were failing due to a phpseclib regression, affected version is blocked in the base branch.

PR #1547, opened 2022-09-07
Use visibility argument when ::copy() in AWS adapters

fnash:
Avoids `get-object-acl` operation and `s3:GetObjectAcl` permission.

### DIFF
--- a/src/AsyncAwsS3/AsyncAwsS3Adapter.php
+++ b/src/AsyncAwsS3/AsyncAwsS3Adapter.php
@@ -285,7 +285,7 @@ class AsyncAwsS3Adapter implements FilesystemAdapter
     {
         try {
             /** @var string $visibility */
-            $visibility = $this->visibility($source)->visibility();
+            $visibility = $config->get(Config::OPTION_VISIBILITY) ?: $this->visibility($source)->visibility();
         } catch (Throwable $exception) {
             throw UnableToCopyFile::fromLocationTo($source, $destination, $exception);
         }
--- a/src/AwsS3V3/AwsS3V3Adapter.php
+++ b/src/AwsS3V3/AwsS3V3Adapter.php
@@ -394,7 +394,7 @@ class AwsS3V3Adapter implements FilesystemAdapter
     {
         try {
             /** @var string $visibility */
-            $visibility = $this->visibility($source)->visibility();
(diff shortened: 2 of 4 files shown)


src/AsyncAwsS3/AsyncAwsS3Adapter.php
```diff
@@ -285,7 +285,7 @@ public function copy(string $source, string $destination, Config $config): void
     {
         try {
             /** @var string $visibility */
-            $visibility = $this->visibility($source)->visibility();
+            $visibility = $config->get(Config::OPTION_VISIBILITY, $this->visibility($source)->visibility());
```
GromNaN: The call to read the visibility can be omitted when the value is configured.

```suggestion
            $visibility = $config->get(Config::OPTION_VISIBILITY) ?: $this->visibility($source)->visibility();
```

src/AwsS3V3/AwsS3V3Adapter.php
```diff
@@ -394,7 +394,7 @@ public function copy(string $source, string $destination, Config $config): void
     {
         try {
             /** @var string $visibility */
-            $visibility = $this->visibility($source)->visibility();
+            $visibility = $config->get(Config::OPTION_VISIBILITY, $this->visibility($source)->visibility());
```
GromNaN:
```suggestion
            $visibility = $config->get(Config::OPTION_VISIBILITY) ?: $this->visibility($source)->visibility();
```

GromNaN (commented):
This is interesting to suppress a call to S3 api when the visibility can be specified.

1. Is it possible to add a test?
2. Is there other adapters to update to ensure a consistent behavior?

GromNaN (approved):

fnash:
@frankdejonge Is it Okay for you?

frankdejonge:
@fnash can you base this PR on 3.x please?

fnash:
Okay!

But need support for php7.4. Do I make another PR for 2.x?

frankdejonge:
@fnash 2.x is in security fixes only mode as of a couple of months.

frankdejonge:
I can see if I can make an exception, but that exposes me to more people who want an exception ... :/

GromNaN:
That's kind of security issue since the visibility that is specified is not respected.

frankdejonge:
I think you're exaggerating the issue now, but I'll look it I have time for it.

Issue #1006, opened 2019-02-15
FTP adapter and createDir with a dot folder


GregOriol:
### Bug Report

|    Q        |   A
|------------ | ------
| BC Break    | no
| Version     | v1.0.50

#### Summary

With the FTP adapter, it seems that creating a directory inside a dot directory doesn't work.
Server is pure-ftpd on ubuntu.

#### How to reproduce

I'm trying to create a directory inside a dot directory like this `->createDir('.dir1/dir2');` and it fails because in FTP's createActualDirectory method, `$listing = ftp_nlist($connection, '.') ?: [];` returns only non-dot files and directories. So the createDir method thinks .dir1 doesn't exist, and tries to create it, which fails.

It is a problem because it can break in a few other use cases: when a dot file with the same name as the directory exists for example, createActualDirectory will try to create the dot directory and fail too, with a non explicit error.

A few more information:
* this issue happens with pure-ftpd using configuration defaults from ubuntu
* if enabling dot files in the pure-ftpd configuration ("yes" in /etc/pure-ftpd/conf/DisplayDotFiles), it works
* this issue doesn't happen with vsftpd using configuration defaults from ubuntu

frankdejonge:
Hi, I've tried to reproduce this but I'm getting an error trying to create the directory `.dirname` which is not what you describe here. The error message states is not possible to create a directory with that name. So it seems dot-prefixed directory names are prohibited.

frankdejonge:
For now I'm going to close this issue because I think it's expected behaviour. If not, please comment with more information and we can re-open this issue.

GregOriol:
@frankdejonge Did you try it with different servers? The behavior might vary from server implementations I think.

I can't agree with the "expected behaviour": a server can have dot folders, and handling dot files must not fail in a bizarre way.

frankdejonge:
Regardless of what you think of it, I can't make a FTP server behave different, nor can I make the ftp_nlist function return dot prefixed directories. I also didn't see anything in the PHP docs about this described behaviour and I couldn't reproduce it with a standard implementation of FTP. I'm not going to just try all the FTP version out there, this is not my job.

So unless you supply better information or supply a PR that proves it (preferably with a server in docker) this issue is not actionable for me.

GregOriol:
@frankdejonge I actually provided details on which servers I did and dit not experienced the issue!

[...]

PR #1551, opened 2022-09-15
Fix for hidden folders started with dot

Edrard:
Fix for hidden folders started with dot
Issues https://github.com/thephpleague/flysystem/issues/1006


frankdejonge:
@Edrard please add a test case that reproduces the issue.

Edrard:
Good day/ I fixed the test, but I do not know how to reproduce it. ftp_nlist() function have the problem, it does not want to show hidden folders on Linux, and those folders started with dot in front.

frankdejonge:
V1 is no longer supported. Closing the PR.

PR #1554, opened 2022-09-18
fix for createDirectory when prefix is used

oneCentral:
When a prefix is used the createDirectory is incorrect. What is going wrong: 
line 212: The parts are first retrieved including the prefix
line 224: The $directoryPath included the prefix. But the function 'directoryExists' will also put the prefix in front. So check is performed with 2 times the prefix in the path


frankdejonge:
Thank you for raising the issue. I've fixed the issue a little bit different because the proposed solution doesn't create the root  directory. Let me know if the fix works for you :)

frankdejonge:
The fix is released as 3.5.1, thanks again for reporting and providing a description of the issue.

PR #1562, opened 2022-10-04
[FEATURE] Public URL generation.

frankdejonge:


src/AwsS3V3/AwsS3V3AdapterTest.php
```diff
@@ -23,8 +23,10 @@
 use League\Flysystem\Visibility;
 use RuntimeException;
 
+use function file_get_contents;
 use function getenv;
 use function iterator_to_array;
+use function var_dump;
```
flangofas: @frankdejonge I don't think you need this anymore :)

src/UrlGeneration/PublicUrlGenerator.php
```diff
@@ -0,0 +1,16 @@
...
+
+use League\Flysystem\Config;
+use League\Flysystem\UnableToGeneratePublicUrl;
+
+interface PublicUrlGenerator
+{
+    /**
+     * @throws UnableToGeneratePublicUrl
+     */
+    public function publicUrl(string $path, Config $config): string;
```
m-vo, line 15: Have you considered returning a `Psr\Http\Message\UriInterface` here instead of a `string`?

m-vo:
May I ask what's the definition of a "public" URL in this PR? :slightly_smiling_face: 

Is it a URI pointing to a resource that the PHP process is able to read from? Or something that is public for anyone (such as an AWS public URL)?

frankdejonge:
@m-vo it's the URL that a browser or http client can use to download the file.

src/UrlGeneration/PublicUrlGenerator.php
frankdejonge: Too much hassle IMO. The dominant case is to generate URLs for assets, not to manipulate the URL. If people do have those use-cases, they can take the URL string and manipulate it with something like league/uri, but there's no need for Flysystem to force that interface upon its users.

src/AwsS3V3/AwsS3V3AdapterTest.php
frankdejonge: thanks, removed :)

m-vo:
> it's the URL that a browser or http client can use to download the file.

Ok, I was confused by the unit test for the memory adapter, that returns `memory://prefix/path.txt`.

frankdejonge:
@m-vo that's to ensure the test runs quickly (no network).

src/UrlGeneration/PublicUrlGenerator.php
m-vo: Fair point. Looks like we're also only ever dealing with URLs. :+1:

alexander-schranz:
We are in @sulu using some mechanism - If "local" return file via `BinaryFileResponse`, if "external" supporting PublicUrl do `RedirectResponse` to the Public Url, if no Public Url is supported I planned to use a StreamedResponse.

This kind of code would look like the following (untested):

```php
<?php

class DownloadController
{
    public function __construct(private MediaRepository $mediaRepository) {}
  
    public function __invoke(Request $request)
    {
         $media = $this->mediaRepository->findById($request->attributes->getInt($id));
      
         $this->checkIsAllowed($media);

         $path = $media->getPath();
      
         // the following code snippet will not work as an adapter could be wrapped in "PathPrefixedAdapter"
         // and there is currently no way to check instance of the innerAdapter

         if ($adapter instanceof LocalFilesystemAdapter) {
              // use x-sendfile for most efficient streaming of local files

              return new BinaryFileResponse($adapter->getLocalFullPath($path)); // currently getLocalFullPath not supported by LocalFilesystemAdapter
         } elseif ($adapter implements PublicUrlGenerator) {
              // redirect to external storage url

              return new RedirectResponse($adapter->getPublicUrl($path), 302);
         }

         return new StreamedResponse(function() use($adapter) {
              $outputResource = \fopen('php://output');
           
              \stream_copy_to_stream(
                  $adapter->readStream($path),
                  $outputResource
              );
         });
    }
}
```

The problem I see is with the `PathPrefixedAdapter` added lately it would be not possible to `detect` if the inner adapter is `LocalFilesystemAdapter` or `PublicUrlGenerator`. I think the try and error for the `PublicUrlGenerator` feels a little bit strange 🤔 .

And what is your opinion for the `LocalFilesystemAdapter` to add a method `getFullPath` so it can be streamed via x-sendfile.

m-vo:
In @contao we built a similar mechanism. If there **is** a public URL, we simply link to it and otherwise use a `StreamedResponse` like you did in your example. Did you test if there really is a significant benefit of using a `BinaryFileResponse` over streaming?

Testing for the kind of adapter IMHO breaks the abstraction.

alexander-schranz:
@m-vo if you have xsendfile enabled it is a good improvement as php child process is not required to stream the file and nginx is doing it for you. Specially big files would block a pho process along time.

frankdejonge:
@alexander-schranz with this feature, the local filesystem would have a `public_url` setting in the filesystem as the filesystem doesn't know what is exposed or not. This means that you can always set this to the root of the filesystem, even with a specific prefix to allow easier detection. For example:

```php
$rootPath = __DIR__ . '/some/directory';
$filesystem = new Filesystem(new LocalFilesystemAdapter($root), ['public_url' => "@$rootPath"]);

try {
  $url = $filesystem->publicUrl($path);
} catch ($e) {
  return new StreamedResponse(function() use($filesystem, $path) {
     $outputResource = \fopen('php://output');
           
     \stream_copy_to_stream($filesystem->readStream($path), $outputResource);
   });
}

if (str_starts_with($url, '@')) {
  return new BinaryFileResponse(substr($url, 1));
} else {
  return new RedirectResponse($url, 302);
}
```

m-vo:
> even with a specific prefix to allow easier detection

I don't like this to be honest. Then every consumer always has to parse and check the public paths. IMHO things returned from  `publicUrl()` should either return a valid public URL, that I can safely pass to the the (http) client or `null` if there ain't one. 

For me it looks like in the `BinaryFileResponse` example we don't want a public URL but one that the web server can access. This might either be a completely different feature or you would need something to distinguish the consumers, i.e. `$filesystem->publicUrl($path, for: 'foobar')`.

m-vo:
Another thing, that came to my mind is, that `publicUrl()` could/should maybe return an object that can also contain meta data. How would you for instance communicate expiration dates on generated URLs to the consumer?

It might also be needed to pass in options when generating a URL. Think about the above mentioned expiration date or some setting about the content disposition type and so on (S3 presigned URL settings for instance)… Sure, these are provider specific but maybe it could be implemented in an opt-in way.

frankdejonge:
I think I'll go ahead with this implementation for now. When we need more complicated url structures, we can add them lated. For example, private or temporary urls can use the same design but be a different method.

m-vo:
@alexander-schranz Here is another idea to solve the local path problem (we'll probably use that in Contao):

```php
$stream = $storage->stream($path);
$metadata = stream_get_meta_data($stream);

if ('STDIO' === $metadata['stream_type'] && 'plainfile' === $metadata['wrapper_type'] && Path::isAbsolute($localPath = $metadata['uri'])) {
   // Use BinaryFileResponse with $localPath
} else {
   // Use StreamResponse with $stream
}
```

alexander-schranz:
@m-vo interesting. Thx for the update 👍

PR #1565, opened 2022-10-14
[feature] allow `PrefixPublicUrlGenerator` to have multiple prefixes

kbond:
When multiple prefixes are used, the generated URL's prefix is deterministic (the same path used multiple times will always have the same prefix).

### DIFF
--- a/src/UrlGeneration/PrefixPublicUrlGenerator.php
+++ b/src/UrlGeneration/PrefixPublicUrlGenerator.php
@@ -7,17 +7,42 @@ namespace League\Flysystem\UrlGeneration;
 use League\Flysystem\Config;
 use League\Flysystem\PathPrefixer;
 
+use function array_map;
+use function array_values;
+use function count;
+use function fmod;
+use function hash;
+use function hexdec;
+use function ltrim;
+use function mb_substr;
+
 class PrefixPublicUrlGenerator implements PublicUrlGenerator
 {
-    private PathPrefixer $prefixer;
+    /** @var PathPrefixer[] */
+    private array $prefixers;
 
-    public function __construct(string $urlPrefix)
+    /**
+     * @param string|string[] $urlPrefix
+     */
+    public function __construct(string|array $urlPrefix)
     {
-        $this->prefixer = new PathPrefixer($urlPrefix, '/');
+        $this->prefixers = array_map(
+            static fn (string $prefix) => new PathPrefixer($prefix),
+            array_values((array) $urlPrefix)
+        );
     }
 
     public function publicUrl(string $path, Config $config): string
     {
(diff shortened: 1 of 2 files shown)


src/UrlGeneration/PrefixPublicUrlGenerator.php
```diff
@@ -7,17 +7,41 @@
...
+        if (1 === count($this->prefixers)) {
+            return $this->prefixers[0]->prefixPath($path);
+        }
+
+        /**
+         * @source https://github.com/symfony/symfony/blob/294195157c3690b869ff6295713a69ff38b3039c/src/Symfony/Component/Asset/UrlPackage.php#L115
+         */
+        $index = (int) fmod(hexdec(mb_substr(hash('sha256', $path), 0, 10)), count($this->prefixers));
+
+        return $this->prefixers[$index]->prefixPath($path);
```
kbond, line 46: One small issue: a path of `/some/path.txt` could have a different prefix than `some/path.txt` (no `/` prefix). I can account for this if desired.

kbond: (fixed)

frankdejonge:
I don't understand what problem this solves. Can you explain?

kbond:
> Can you explain?

It's my understanding that browsers (under certain conditions) [limit the maximum number of requests](https://stackoverflow.com/questions/985431/max-parallel-http-connections-in-a-browser) that can be made to the same domain in parallel. For a page that has a ton of images all being loaded from your cdn (ie cdn.example.com), these images can be slow to load.

A common solution to this problem is to have multiple cdn subdomains that all point to the same source (ie cdn1.example.com, cdn2.example.com, etc). This PR enables this. It's important that the generated url is deterministic for the browser cache.

Additionally, this feature could be used to effectively _load balance_ your cdn (assuming you have multiple cdn servers that have access to the same set of files).

frankdejonge:
I'm not sure if this belongs in this library. The difficult part about this is that the generator is internal and makes a couple of decisions that people may want to change. For example, the substringing (which is probably a perf optimization) may not provide a good distribution.

kbond:
That's fair. What about a separate `MultiPrefixPublicUrlGenerator` that would be configurable? Perhaps with a callback that generates the `$index` for the path?

If not, totally fine, feel free to close this!

frankdejonge:
I'm going to pass on this one. I don't feel strongly enough to include it in light of perhaps needing to justify/defend the choices made.

frankdejonge:
With the custom prefixer, you don't actually need this as you can provide an alternate implementation. I would be open to a PR that implements one that purely does this distribution and always expects a list of prefixes.

PR #1566, opened 2022-10-16
[feature] add `Filesystem::checksum()`

### LINKED ISSUE #1283: md5 checksum
### Feature Request

|    Q        |   A
|------------ | ------
| New Feature | yes
| BC Break    | yes

#### Scenario / Use-case

we store images on aws s3.
we have a resize cache locally 
i need to decide resource friendly my file is changed on aws s3 or not.

#### Summary

getChecksum($path)
on aws s3 we could use ETAG
on local file system md5_file

if a filesystem does not support md5 then create locally a temp file as default solution
```
$path = sys_get_temp_dir();
$filename = tempnam($path, 'storage');
$handle = fopen($filename, 'wb');
fwrite($handle, $content);
fclose($handle);
```

OR 

at least be able to access File Attributes

kbond:
If an adapter implements `ChecksumProvider`, this method returns the checksum. If not, the checksum is calculated (md5) based on the file contents.

The following adapters support checksums natively:
- Local (uses `md5_file`)
- AsyncAwsS3 (uses `ETag` header value)
- AwsS3V3Adapter (uses `ETag` header value)

Fixes #1283.

### DIFF
--- a/src/ChecksumProvider.php
+++ b/src/ChecksumProvider.php
@@ -1,0 +1,16 @@
+<?php
+
+namespace League\Flysystem;
+
+/**
+ * @author Kevin Bond <kevinbond@gmail.com>
+ */
+interface ChecksumProvider
+{
+    /**
+     * @return string MD5 hash of the file contents
+     *
+     * @throws UnableToGetChecksum
+     */
+    public function checksum(string $path, Config $config): string;
+}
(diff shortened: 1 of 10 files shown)


src/ChecksumProvider.php
```diff
@@ -0,0 +1,16 @@
+<?php
+
+namespace League\Flysystem;
+
+/**
+ * @author Kevin Bond <kevinbond@gmail.com>
```
frankdejonge, line 6: can you remove this? I don't have those in this library anywhere and don't look forward to having to decide what gets source level credits.

frankdejonge:
I like it. I'm wondering if we should prepare for a configurable preferred algo setting for the fallback and the local adapter.

kbond:
I'm thinking the checksum be part of `FileAttributes` in v4.

PR #1583, opened 2022-10-27
fixes for ftp copy and write

t3ran13:
1) It was ignored config for Copy and always set visibility even it was created with the same. It results to permition issue for me
2) It do not have sense to set visibility the same as file has and can lead to permitions issue (during write file)

### DIFF
--- a/src/Ftp/FtpAdapter.php
+++ b/src/Ftp/FtpAdapter.php
@@ -164,7 +164,10 @@ class FtpAdapter implements FilesystemAdapter
             throw UnableToWriteFile::atLocation($path, 'writing the file failed');
         }
 
-        if ( ! $visibility = $config->get(Config::OPTION_VISIBILITY)) {
+        $visibility = $config->get(Config::OPTION_VISIBILITY);
+        if ( ! $visibility
+            || $visibility === $this->visibility($location)->visibility()
+        ) {
             return;
         }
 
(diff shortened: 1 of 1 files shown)


src/Ftp/FtpAdapter.php
```diff
@@ -164,7 +164,10 @@ public function writeStream(string $path, $contents, Config $config): void
             throw UnableToWriteFile::atLocation($path, 'writing the file failed');
         }
 
-        if ( ! $visibility = $config->get(Config::OPTION_VISIBILITY)) {
+        $visibility = $config->get(Config::OPTION_VISIBILITY);
+        if ( ! $visibility
+            || $visibility === $this->visibility($location)->visibility()
+        ) {
```
frankdejonge, line 170: this is replacing one FTP call with another, but in case there is a difference there is not 2 FTP calls. Can you explain why this is needed?

t3ran13: hi @frankdejonge

t3ran13: hi @frankdejonge 
to update file/directory permissions your FTP user needs permitions for that. It meens that some users with copy operation create file 644 but cannot update permitions for this file to the same 644 and got exception. But at the same time we created 644 file on ftp server. 

I reproduce this issue for 2 different docker FTP servers.


may be this is the reason
https://www.quora.com/Why-doesnt-the-CHMOD-command-work-on-all-FTP-SFTP-servers
https://stackoverflow.com/questions/30541941/php-ftp-chmod-not-working-on-file-but-ftp-site-does-why

But in general, if i created 644 file i do not need to do chmod 644 for it

PR #1594, opened 2022-11-18
fix gcloud bucket storage copy when acl not available

### LINKED ISSUE #4: It should be possible not to use the visibility handler/visibility settings
GCP storage buckets now suggest not to use ACL but bucket wide permissions when possible. Unfortunately trying to upload a file to such bucket throws an error:

Cannot insert legacy ACL for an object when uniform bucket-level access is enabled. Read more at https://cloud.google.com/storage/docs/uniform-bucket-level-access

That was a problem for me, I ended up making a copy of the adapter since it was unable to extend it and just commented out the following line:

'predefinedAcl' => $this->visibilityHandler->visibilityToPredefinedAcl($visibility),

it would be good to actually allow null $defaultVisibility, $dafaultAcl with null works when doing uploads.

### LINKED ISSUE #1593: GCloud Adapter Bug - Calling ACL causes issues when bucket has recommended uniform access setup
### Bug Report

| Flysystem Version | latest  |
| Adapter Name      | gcloud |
| Adapter version   | latest   |

#### Summary

The issue here should be addressed:
https://github.com/thephpleague/flysystem-google-cloud-storage/issues/4

#### How to reproduce

If you have a bucket with the uniform access as recommended by Google, we get errors if there is any attempt to access ACL. The adapter needs a way to disable calling acl functionality. 

The issue describes setting the config `visibilirty` to `noPredefinedVisibility`

This does not solve the issue for using `copy`

Is there something I'm missing?

Many thanks.

silverbackdan:
fix: https://github.com/thephpleague/flysystem/issues/1593
fix: https://github.com/thephpleague/flysystem-google-cloud-storage/issues/4

This will allow the copy function to work when using uniform IAM access control as recommended by google

### DIFF
--- a/src/GoogleCloudStorage/GoogleCloudStorageAdapter.php
+++ b/src/GoogleCloudStorage/GoogleCloudStorageAdapter.php
@@ -348,8 +348,12 @@ class GoogleCloudStorageAdapter implements FilesystemAdapter, PublicUrlGenerator
     public function copy(string $source, string $destination, Config $config): void
     {
         try {
-            /** @var string $visibility */
-            $visibility = $this->visibility($source)->visibility();
+            try {
+                /** @var string $visibility */
+                $visibility = $this->visibility($source)->visibility();
+            } catch (UnableToRetrieveMetadata $e) {
+                $visibility = $config->get(Config::OPTION_VISIBILITY, $this->defaultVisibility);
+            }
             $prefixedSource = $this->prefixer->prefixPath($source);
             $options = ['name' => $this->prefixer->prefixPath($destination)];
             $predefinedAcl = $this->visibilityHandler->visibilityToPredefinedAcl($visibility);


silverbackdan:
This is also referenced in an issue from the google bucket repo split here: https://github.com/thephpleague/flysystem-google-cloud-storage/issues/4#issuecomment-1335198537

silverbackdan:
Just a little ping @frankdejonge - I see you may be one of the maintainers responsible for merging. Just had a comment on the issue in the google bucket split again that I've just referenced and this would be a really useful fix if possible.
Or perhaps discussion on a better way to disable ACL for google bucket storage.
I know you've probably got a load of issue on the go though, appreciate your time. Thanks.

frankdejonge:
@silverbackdan Hi, I've created an alternate fix for this by introducing a specific visibility handler. I'll release it shortly.

silverbackdan:
Thanks @frankdejonge - that sounds like a much more complete less hacky approach, thank you.

frankdejonge:
It was released as 3.11.0, you can now use a `new UniformBucketLevelAccessVisibility()` as the `visibilityHandler:` adapter constructor property.

silverbackdan:
Thank you, I haven't tested this yet but I fear an issue still with `UniformBucketLevelAccessVisibility::determineVisibility` as it has a call to `$object->acl()`

Edit: Sorry, I see the catch now and then can see this comes back as Private.. Not sure though because you can set the bucket to have files public accessible at a bucket level and the acl will break it still and return as private. act I don't think should be called at all in this visibility handler IMHO, but I may be wrong.

silverbackdan:
I think as all permissions are at bucket level, we should be able to do something which simply checks if the bucket is allowing public access or not, if it is, it's public, if it isn't it's private.

rwinzhang:
This issue still persist as @silverbackdan mention because it has a call to `$object->acl()`.
My suggestion, we could return `self::NO_PREDEFINED_VISIBILITY` directly on function `UniformBucketLevelAccessVisibility::determineVisibility` because there is no other visibility option on uniform bucket level access storage, cc @frankdejonge 

Here is my PR https://github.com/thephpleague/flysystem/pull/1629

PR #1607, opened 2022-12-08
strip prefix with use start position

dormadekhin:
This fixes the problem when the absolute path is returned. This behavior is typical for WebDavAdapter.

### DIFF
--- a/src/PathPrefixer.php
+++ b/src/PathPrefixer.php
@@ -29,7 +29,7 @@ final class PathPrefixer
     public function stripPrefix(string $path): string
     {
         /* @var string */
-        return substr($path, strlen($this->prefix));
+        return substr($path, (int)strpos($path, $this->prefix) + strlen($this->prefix));
     }
 
     public function stripDirectoryPrefix(string $path): string


frankdejonge:
@dormadekhin what webdav implementation are you using? I've integration tested the webdav component and it doesn't show this behaviour.

dormadekhin:
@frankdejonge, I use League\Flysystem\WebDAV\WebDAVAdapter. Тhis behavior is seen in vendor/league/flysystem-webdav/WebDAVAdapter.php:274
debug 
image

I think we have to consider the position of prefix, because this behaviour is more predictable.

frankdejonge:
Sorry, I meant what is the WebDAV server implementation you’re using.

dormadekhin:
I'm use sabre/dav client in php

dormadekhin:
IIS with WebDav

dormadekhin:
image

frankdejonge:
What happens when you configure the root to include the full url?

dormadekhin:
If you mean the webdav side configuration, we don't have access to change it. We are only consumers. 
could you clarify the question?

dormadekhin:
This change will not affect the current behavior, but will only add handling of exceptional cases. The current behavior is not quite correct, since the prefix position is not taken into account. I think that the implementation should be universal.

frankdejonge:
I mean on the consumer client configuration.

frankdejonge:
I understand you remark regarding universal solutions, however. In this case the problem is not universal, it’s very isolated. In those cases I prefer to not solve it globally but locally in order to make the universal model not bear too much responsibility and keep it as lean as possible and solve the problem as close to the source.

dormadekhin:
> 



> What happens when you configure the root to include the full url?

Sabre/dav hasn't options "root url"
My configuration 
image

dormadekhin:
I cannot agree with you in this situation. Using only the prefix length does not guarantee uniform behavior on different data. I don't see the complexity of the code, on the contrary, it adds stability.

frankdejonge:
The current implementation _is_ correct since a prefix is at the start and you're trying to cater to an infix situation, which is the exception. The prefix length should be enough when properly configured an mappable 1 to 1.

I meant that you could include the URL in the adapter prefix, the one you've configured to `DocsForCustoms/` now.

dormadekhin:
After including the URL in the prefix, the request is not executed
image
request error
image

dormadekhin:
Will there be any additional explanations for this request? I also have a slightly different solution to this problem. Is it possible to implement the ability to pass the Prefixer class to the adapter?

dormadekhin:
@frankdejonge, will there be any solution to this request?

frankdejonge:
I've implemented an alternative fix for this: https://github.com/thephpleague/flysystem/pull/1613

PR #1621, opened 2023-01-05
Chunk s3 delete calls when deleting huge directories with AsyncAws

### LINKED ISSUE #1615: Unable to delete a folder with many files using AsyncAWS
### Bug Report

| Flysystem Version | 3.10.4   |
| Adapter Name      | async-aws|
| Adapter version   | 3.10.3   |

#### Summary

Unable to delete a large folder. Deleting some files by UI then deleting the whole folder works (Maybe some pagination issues?)

(Used Minio)

```
In Response.php line 422:
                                                                                                       
  HTTP 400 returned for "http://127.0.0.1:9000/asset?delete=".                                         
                                                                                                       
  Code:    MalformedXML                                                                                
  Message: The XML you provided was not well-formed or did not validate against our published schema.  
  Type:                                                                                                
  Detail:
```

#### How to reproduce

Create a folder with 5k files and try to delete them

keulinho:
Fixes https://github.com/thephpleague/flysystem/issues/1615

As stated in the linked issue AsyncAws is dumb and does no chunking, in the AsyncAws Adapter we also loop over all objects in a "directory", and create one big delete request for all objects.

This may lead to issues if to many objects should be deleted in one request, as stated in the linked issue. IMHO as AsyncAws already loops through all objects in a "directory" it should also chunk the created delete commands accordingly.

I'm not sure about the exact chunk size we should use, locally trying to delete a dir with ~5.000 objects failed, using a chunk size of 1.000 worked so I used that.

We can also think about making the chunk size configurable, but i'm not sure if that is really needed.

### DIFF
--- a/src/AsyncAwsS3/AsyncAwsS3Adapter.php
+++ b/src/AsyncAwsS3/AsyncAwsS3Adapter.php
@@ -186,7 +186,12 @@ class AsyncAwsS3Adapter implements FilesystemAdapter, PublicUrlGenerator, Checks
             return;
         }
 
-        $this->client->deleteObjects(['Bucket' => $this->bucket, 'Delete' => ['Objects' => $objects]]);
+        foreach (array_chunk($objects, 1000) as $chunk) {
+            $this->client->deleteObjects([
+                'Bucket' => $this->bucket,
+                'Delete' => ['Objects' => $chunk],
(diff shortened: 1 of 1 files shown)


keulinho:
Thanks for merging!

PR #1628, opened 2023-01-17
Fix incorrect argument amount for S3 fileExists()

SergeyPodgornyy:
image
image


SergeyPodgornyy:
@frankdejonge could you please see this bugfix to the code of S3? :pray: This bug causes some problems with the production application I maintain.

Also, it would be really valuable to update sub-repo [thephpleague/flysystem-aws-s3-v3](https://github.com/thephpleague/flysystem-aws-s3-v3) with this fix

SergeyPodgornyy:
Thank you @frankdejonge :bowing_man: Could you also create a new release tag in both repositories to be able to download the fix, please? :pray:

frankdejonge:
It's released as 3.12.2

SergeyPodgornyy:
Thank you

PR #1629, opened 2023-01-18
Fix ACL error on uniform bucket

rwinzhang:


frankdejonge:
Can you supply the error this actually fixes?

rwinzhang:
image

The ACL error occured when copy or move file because it has a call to `$object->acl()`.
My suggestion, we could just return self::NO_PREDEFINED_VISIBILITY directly on function UniformBucketLevelAccessVisibility::determineVisibility because there is no other visibility option on uniform bucket level access storage

silverbackdan:
Fixes #1633 as well

silverbackdan:
Alternative fix PR opened so acl support is not removed. #1634 
Edit, hadn't seen where this update was made, yes, this looks like a better solution and would be a great fix to have released ASAP as I need to use another fork until merged due to this causing a fatal error when copying a file on a bucket which does not have ACL enabled.

rwinzhang:
Hi @frankdejonge, what do you think about the fixes?

PR #1634, opened 2023-01-26
Alternative fix option for ACL error with GoogleCloud adapter on copy

### LINKED ISSUE #1633: Google ACL called causes error during copying
### Bug Report

| Flysystem Version | 3.x   |
| Adapter Name      | GoogleCloud |
| Adapter version   | 3.x   |

#### Summary

This continues to be a bug https://github.com/thephpleague/flysystem/pull/1594
The copy call ends up trying to call ACL which causes an error during copying. I am reimplementing my PR hack fix for now on my own fork.

#### How to reproduce

Try to copy a file when ACL is disabled on a cloud bucket.

silverbackdan:
Alternative to #1629
Fixes #1633

### DIFF
--- a/src/GoogleCloudStorage/GoogleCloudStorageAdapter.php
+++ b/src/GoogleCloudStorage/GoogleCloudStorageAdapter.php
@@ -348,8 +348,12 @@ class GoogleCloudStorageAdapter implements FilesystemAdapter, PublicUrlGenerator
     public function copy(string $source, string $destination, Config $config): void
     {
         try {
-            /** @var string $visibility */
-            $visibility = $this->visibility($source)->visibility();
+            try {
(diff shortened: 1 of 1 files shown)


silverbackdan:
Alternatively we could check `$this->visibilityHander instanceof UniformBucketLevelAccessVisibility` and use the default visibility configured.

There are a number of ways to resolve this issue. Happy to discuss which is best.

I think for no ACL the recommended approach was `new GoogleCloudStorageAdapter(bucket: $bucket, visibilityHandler: new UniformBucketLevelAccessVisibility())`

silverbackdan:
Apologies, I hadn't seen where the file was changed, the other fix looks better.

PR #1640, opened 2023-02-09
Add support for PuTTY ppk files to sftp v3

boserup:
phpseclib v3 supports loading PuTTY ppk files, but the sftp v3 provider assumes the provided private key is always a file, if it does not start with 3 dashes.


frankdejonge:
Do all .ppk files start with PuTTY? Is there a source I can use to verify this?

boserup:
@frankdejonge Yes - actually it is `PuTTY-User-Key-File-` followed by the version (1, 2 or 3): https://github.com/github/putty/blob/master/sshpubk.c#L473-L479

PR #1646, opened 2023-03-03
Add psalm types to DirectoryListing methods

annervisser:


frankdejonge:
I had just pushed the same thing 😅 thanks for this though, the functionality will ship in the next release.

PR #1648, opened 2023-03-09
minor(AsyncAwsS3): add test to ensure top-level directory is excluded

kbond:
Follow-up to #1647.


frankdejonge:
@kbond this seems to fail 😅

kbond:
Damn, ~wasn't sure how to test locally~. The ci doesn't test this adapter?

Figured out how to test locally, working on a fix.

PR #1666, opened 2023-05-09
Use local adapter as new package

ottaviano:
Hi, with this config composer does not install new package `league/flysystem-local`

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -50,18 +50,5 @@
             "name": "Frank de Jonge",
             "email": "info@frankdejonge.nl"
         }
-    ],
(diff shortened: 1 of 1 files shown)


frankdejonge:
[what you're stating is not correct](https://getcomposer.org/doc/05-repositories.md#:~:text=Repositories%20are%20only%20available%20to%20the%20root%20package%20and%20the%20repositories%20defined%20in%20your%20dependencies%20will%20not%20be%20loaded.)

frankdejonge:
image

PR #1683, opened 2023-07-17
[AsyncAwsS3] adapt options for a temporary url same as in the AwsS3V3Adapter

smoench:
In this PR I added the possibility to provide additional options for the presign url. In our case we wanted to provide the `ResponseContentDisposition` but it wasn't possible.

I followed the same approach as in the [AwsS3V3Adapter](https://github.com/thephpleague/flysystem/blob/3.x/src/AwsS3V3/AwsS3V3Adapter.php#L494). With this change there is also no longer the requirement to use the `SimpleS3Client` to get a presigned url.


smoench:
Hey @frankdejonge,
sorry for pinging you here. I know you are busy right now but in case you find a free minute can you may have a look on this PR. Please let me know when there is something missing I can provide/fix :-)
Thanks in advance

PR #1684, opened 2023-07-17
Fix InMemory not overwriting on move

kynx:
This fixes the `InMemoryFilesystemAdapter` not copying or moving directories

### DIFF
--- a/src/InMemory/InMemoryFilesystemAdapter.php
+++ b/src/InMemory/InMemoryFilesystemAdapter.php
@@ -19,7 +19,9 @@ use League\MimeTypeDetection\MimeTypeDetector;
 
 use function array_keys;
 use function rtrim;
+use function strlen;
 use function strpos;
+use function substr;
 
 class InMemoryFilesystemAdapter implements FilesystemAdapter
 {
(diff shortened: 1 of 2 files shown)


frankdejonge:
Can you make this strategy configurable via a constructor option? Otherwise this would be a major BC break and I do not want to cause that.

kynx:
@frankdejonge Uh, OK. Guess the documentation should be updated to reflect the fact that not all adapters behave the same?

kynx:
@frankdejonge Actually I might have got a little carried away with copying and moving directories. My initial problem was with the SFTP adapter not overwriting _files_ when moving after upgrading an old project from 1.x to 3.x. Looking back at the docs there is no mention of copying or moving directories.

I do think it's a bug that some adapters don't overwrite files on move - the docs clearly say they do. If I moved the directory stuff to a different PR targeting the next major, would that work?

kynx:
@frankdejonge I've removed the directory copy / move stuff. This now just brings this adapter in line with the others, overwriting existing files on move.

Please let me know if you still think this needs a constructor arg for specifying this behaviour.

SamMousa:
What's wrong with a new major release that deliberately breaks BC?

tinect:
This looks like I managed it with my PR which got now released.
So this PR could be closed, couldn't it?

PR #1688, opened 2023-07-30
Add docker files to .gitattributes

VincentLanglet:
Hi @frankdejonge, I assume those docker file can be added to the gitattribute file too ?


frankdejonge:
Yes, thank you!

PR #1692, opened 2023-08-03
Add phpunit v10.x support

remi-san:
Allow to use `phpunit` v10 to run tests.

Main thing to change was to make the `dataProvider` functions static.

An abstract test class had to be renamed to `xxxTestCase` as the `Test` suffix for an abstract class triggered a warning.

The config file is not in v10 format, but is compatible with both v9 & v10.

I also had to revert a change in the `composer.json` file to be able to have an installable version (otherwise no `symfony/http-client` version could be resolved for php8.0)

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -26,7 +26,7 @@
         "ext-fileinfo": "*",
         "ext-ftp": "*",
         "microsoft/azure-storage-blob": "^1.1",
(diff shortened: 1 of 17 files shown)


frankdejonge:
Much appreciated!

Issue #1468, opened 2022-05-19
FinfoMimeTypeDetector: incomplete handing of INCONCLUSIVE_MIME_TYPES


marc-mabe:
### Bug Report

|    Q        |   A
|------------ | ------
| BC Break    | yes/no
| Version     | 3.0.7

#### Summary

`FinfoMimeTypeDetector` has defined `INCONCLUSIVE_MIME_TYPES` which contains `text/plain`.
This will be used on `detectMimeType` and if the detected type in this list it falls back to `detectMimeTypeFromPath`.

On the other hand the local file adapter uses `detectMimeTypeFromFile` which does not check for `INCONCLUSIVE_MIME_TYPES` at all.

#### How to reproduce

1. Create a tsv file (tab separated values)
2. Use local file adapter `mimeType()` (which used `FinfoMimeTypeDetector::detectMimeTypeFromFile`)
3. you will get `text/plain`

1. Create the same tsv file in memory adapter
2. Use local file adapter `mimeType()` (which used `FinfoMimeTypeDetector::detectMimeType`)
3. you will get `text/tab-separated-values`

#### Idea how to fix

* `INCONCLUSIVE_MIME_TYPES` should be used in all cases a path exists and fall back to `detectMimeTypeFromPath`
* if fallback `detectMimeTypeFromPath` returns `null` use the previous detected mime type

Something like this:
```php
    public function detectMimeType(string $path, $contents): ?string
    {
            $mimeType = is_string($contents)
                ? $this->detectMimeTypeFromBuffer($contents)
                : null;
    
            if ($mimeType !== null && ! in_array($mimeType, self::INCONCLUSIVE_MIME_TYPES)) {
                return $mimeType;
            }
    
            return $this->detectMimeTypeFromPath($path) ?: $mimeType;
    }

    public function detectMimeTypeFromFile(string $path): ?string
    {
        $mimeType = @$this->finfo->file($path) ?: null;

        if ($mimeType !== null && ! in_array($mimeType, self::INCONCLUSIVE_MIME_TYPES)) {
            return $mimeType;
        }

        return $this->detectMimeTypeFromPath($path) ?: $mimeType;
    }
```

GuySartorelli:
Yup, this is causing us some grief as well. Specifically the lack of
> if fallback detectMimeTypeFromPath returns null use the previous detected mime type

[...]

PR #1710, opened 2023-10-10
Fix: Prefer inconclusive mimetype over null

GuySartorelli:
This fixes the scenario where `detectMimeTypeFromFile()` detects an inconclusive mimetype, so we fallback to `detectMimeTypeFromPath()`, which returns null.

Instead of returning null (which will result in `LocalFilesystemAdapter` throwing a `UnableToRetrieveMetadata` exception), we should return the inconclusive result. Valid but inconclusive is better than an exception.

Related issue:
- https://github.com/thephpleague/flysystem/issues/1468

### DIFF
--- a/src/Local/FallbackMimeTypeDetector.php
+++ b/src/Local/FallbackMimeTypeDetector.php
@@ -46,6 +46,6 @@ class FallbackMimeTypeDetector implements MimeTypeDetector
             return $mimeType;
         }
 
-        return $this->detector->detectMimeTypeFromPath($path);
+        return $this->detector->detectMimeTypeFromPath($path) ?? $mimeType;
     }
 }


GuySartorelli:
@frankdejonge Could I bother you to please take a look at this PR? It's causing us some grief and we'd like to know if you will be accepting this pull request (or one like it), or if we will need to handle this scenario in our own application logic.
We're using a fork with this patch applied at the moment, which isn't ideal.

GuySartorelli:
Thank you!

driesvints:
This PR seems to have broken the Laravel test suite: https://github.com/laravel/framework/pull/49229

frankdejonge:
Sorry about that @driesvints, I've made an effective revert on this while making the PR functionality an opt-in. It's released as 3.23.0

driesvints:
@frankdejonge thank you!

PR #1712, opened 2023-10-13
We should close readStream under any circumstances

yankewei:
We should close readStream under any circumstances

### DIFF
--- a/src/PhpseclibV3/SftpAdapter.php
+++ b/src/PhpseclibV3/SftpAdapter.php
@@ -164,9 +164,12 @@ class SftpAdapter implements FilesystemAdapter
         /** @var resource $readStream */
         $readStream = fopen('php://temp', 'w+');
 
-        if ( ! $connection->get($location, $readStream)) {
+        try {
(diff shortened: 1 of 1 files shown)


src/PhpseclibV3/SftpAdapter.php
```diff
@@ -164,9 +164,12 @@ public function readStream(string $path)
...
+        try {
+            if ( ! $connection->get($location, $readStream)) {
+                throw UnableToReadFile::fromLocation($path);
+            }
+        } finally {
             fclose($readStream);
-            throw UnableToReadFile::fromLocation($path);
         }
 
         rewind($readStream);
```
ChloeSartorelli, line 175: We're rewinding the stream here _after_ closing it... won't that throw an error?

frankdejonge:
@yankewei what situation is this fixing?

yankewei:
> @yankewei what situation is this fixing?

We encountered an issue while using the SFTP driver. Initially, I believed the problem was related to this, but now I can eliminate this possibility; there should be no issue here.

PR #1714, opened 2023-10-19
Fix: avoid deleting on s3 when destination is same as source

optii:
Hi,

An issue we encountered that is sort of an edge case in how the s3 move method works. But if the source is the same as the destination file, the actual file will be deleted. For me this is not expected behavior of the "move" command and is not the same result as the filesystem move would be.

Thanks


frankdejonge:
Please check this alternative: https://github.com/thephpleague/flysystem/pull/1715

frankdejonge:
Will be replaced by #1715

PR #1715, opened 2023-10-19
Prevent copy and move to same destinations as the source.

frankdejonge:


optii:
Looks good to me, erroring works too

frankdejonge:
@optii ended up making it a configurable behaviour with the current behaviour being the default.

maxhelias (approved):

PR #1721, opened 2023-11-02
Allow copy without retaining visibility for adapters and implementations that required fetching visibility for copy and move.

frankdejonge:


shadowhand:
Yes, please!

PR #1729, opened 2023-11-17
New exception to capture NET_SSH2_DISCONNECT_BY_APPLICATION errors

fedeisas:
I'm trying to capture the following SFTP error:
```
SSH_MSG_DISCONNECT: NET_SSH2_DISCONNECT_BY_APPLICATION
Account is locked
```

@webmake also reported this here: https://github.com/thephpleague/flysystem-sftp-v3/issues/14

I'm unsure about the naming, but I think it's a start.

The exception message parsing tries to [reflect this](https://github.com/phpseclib/phpseclib/blob/2.0.14/phpseclib/Net/SSH2.php#L3453).

### DIFF
--- a/src/PhpseclibV2/SftpConnectionProvider.php
+++ b/src/PhpseclibV2/SftpConnectionProvider.php
@@ -174,6 +174,12 @@ class SftpConnectionProvider implements ConnectionProvider
         } elseif ($this->useAgent) {
             $this->authenticateWithAgent($connection);
         } elseif ( ! $connection->login($this->username, $this->password)) {
+            if (UnableToConnect::matches($connection->getLastError())) {
+                throw UnableToConnect::disconnected(
+                    $connection->getLastError()
+                );
+            }
+
             throw UnableToAuthenticate::withPassword();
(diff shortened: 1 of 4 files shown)


fedeisas:
@frankdejonge I'd love your feedback here. Thanks 🙏🏼

frankdejonge:
Ill look at this in more detail tomorrow but overall looks fine

frankdejonge:
After a more thorough look, I want to find an alternative. I acknowledge the need you have, but I'm don't think this is the route to get there. The main reason is the duplication of something internal and private into this codebase. Which means that when there's a new error code, or a new message, or any change of the internals to reflect this differently, it's out of sync. Apart from that, this PR only introduces this error when the login method is used but the same scenario applies to private key and agent based authentication. I'll implement an alternative way to provide the same information but be less coupled to the internals of phpseclib, bases on exposing the last error from the connection. You might need to implement the log message interpretation on your end based on this.

fedeisas:
Thanks @frankdejonge!

PR #1730, opened 2023-11-18
Retain file visibility when copying in LocalAdapter

jnoordsij:
This is a follow-up to both #1723 and #1721, that changes the visibility retainment behavior of `LocalAdapter` in the following way:
- retain the visibility of the copied file by default, matching behavior of various other adapters (e.g. the AWS and FTP adapters)
- allow passing a config key to disable this behavior


frankdejonge:
Much appreciated!

PR #1732, opened 2023-11-18
Add config key constant for retain_visibility

jnoordsij:
While working on #1730 I noticed some inconsistent choice in referring to various config options by string vs. by constant. This MR makes the code more consistent, by adding a constant key for the `retain_visibility` option and replacing various string keys by their constant counterparts.

Note that test files were excluded due to more consistently using string keys and this possibly being a pattern that should be left in place.

### DIFF
--- a/src/AsyncAwsS3/AsyncAwsS3Adapter.php
+++ b/src/AsyncAwsS3/AsyncAwsS3Adapter.php
@@ -318,7 +318,7 @@ class AsyncAwsS3Adapter implements FilesystemAdapter, PublicUrlGenerator, Checks
 
             $visibility = $config->get(Config::OPTION_VISIBILITY);
 
-            if ($visibility === null && $config->get('retain_visibility', true)) {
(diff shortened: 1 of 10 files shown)


rvanlaak:
Landed here after trying to integrate a Cloudflare R2 bucket, which doesn't support `GetObjectAcl` and other ACL related commands when using Flysystem's `move` and `copy` actions.
If my assessment is correct, the `retain_visibility` isn't a adapter configuration option, but is something that has to happen on each write method-call? Or maybe it can't be set through the bundle's configuration.
When passing `retain_visibility` => false as adapter option, the Filesystem's `resolveConfigForMoveAndCopy` still leads to not having that value available, and thus removing `visibility`'s value.

frankdejonge:
@rvanlaak when creating the Filesystem instance, you can set global options there.

rvanlaak:
Thanks @frankdejonge , in the end it turned out to be the bundle missing the configuration option to pass that value along: https://github.com/thephpleague/flysystem-bundle/pull/195

PR #1733, opened 2023-11-28
fix: fix phpseclib security vulnerability

yoeriboven:
https://github.com/advisories/GHSA-jpr7-q523-hx25


maxhelias (approved):

frankdejonge:
The title implies that there is a security vulnerability in this library which this fixes. I just want to state on the record that this is not this works. The library didn't prevent the underlying library to upgrade, so this library doesn't have the vulnerability.

PR #1736, opened 2023-12-22
Bumped aws/aws-sdk-php version to fix CVE-2023-51651

M1V4D1M:
Our corporate security auditing mechanism found the CVE-2023-51651 issue in the aws/aws-sdk-php package, but we cannot fix it because flysystem is heavily based on version 3.220.0.

So I decide to make this pull request to fix this problem.

### DIFF
--- a/src/AwsS3V3/composer.json
+++ b/src/AwsS3V3/composer.json
@@ -12,7 +12,7 @@
         "php": "^8.0.2",
         "league/flysystem": "^3.10.0",
         "league/mime-type-detection": "^1.0.0",
-        "aws/aws-sdk-php": "^3.220.0"
+        "aws/aws-sdk-php": "^3.288.1"
     },
     "conflict": {
         "guzzlehttp/ringphp": "<1.1.1",


frankdejonge:
@mi-vadim Hi, you security auditing mechanism should really only check _your_ dependencies that are installed. You can resolve this by either explicitly stating your own dependency or setting up a `conflict` in your composer.json to prevent earlier versions to be installed.

M1V4D1M:
@frankdejonge I may have misspoken. We have the package - **league/flysystem-aws-s3-v3**, the most current version. 
If you look, there is a package package [https://github.com/thephpleague/flysystem/blob/3.x/src/AwsS3V3/composer.json#L15](https://github.com/thephpleague/flysystem/blob/3.x/src/AwsS3V3/composer.json#L15). 
I can't update **aws/aws-sdk-php** separately, because it's a dependency for the flysystem package. Accordingly, the only possible fix I can see is for someone to update the **aws/aws-sdk-php** version in **league/flysystem-aws-s3-v3** and then update their own version.
Of course, I may not know something, I would appreciate an explanation =)

frankdejonge:
The dependency is not pinned, it accepts anything after this version if it’s in the 3.x range. In your application you can require the version you need in addition to the Flysystem adapter to set a different minimum version to fix your security vulnerability. In addition, your tooling should run on your lock file to see what is installed not what transitive dependencies CAN be installed.

M1V4D1M:
@frankdejonge Thank u and sorry for stupid questions ^^

PR #1737, opened 2023-12-26
ensure error when deleting entire storage or deleting directory by delete method instead of deleteDirectory

tinect:
I saw some custom adapters not managing these inputs. E.g. S3 and local adapter throw these exceptions.

I'd like to see these checks in the AdapterTestUtilities to ensure better quality.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -35,7 +35,7 @@
         "google/cloud-storage": "^1.23",
         "async-aws/s3": "^1.5 || ^2.0",
         "async-aws/simple-s3": "^1.1 || ^2.0",
-        "sabre/dav": "^4.3.1"
+        "sabre/dav": "^4.6.0"
     },
     "conflict": {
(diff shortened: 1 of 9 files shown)


tinect:
@frankdejonge what do you think about the issue?

frankdejonge:
Hi @tinect,

There a bunch of unrelated stuff in this PR, which makes it a bit messy. The package's design tends to not perform these kind of checks, but rather leave it up to the user to do this. The reason for this has multiple reasons. Firstly, this increases the amount of underlying calls significantly, making the default less performant. Secondly, which is kind of related to the first, the case where a directory has the same name as a file is very unlikely, as the majority of files have file extensions that wouldn't be in a directory name.

I'm happy to merge some of your other changes in this PR if you sent them individually, but I'm going to close this one. In case you feel like contributing more to this problem (which I'd be grateful for) please keep it to one thing per PR.

PR #1745, opened 2024-01-26
Added license files for the legacy versions.

limeeugenia:
Hello everyone,

Split packages versions before 3.x do not have a license. Since the later version added one, this PR fixes the issue for people still using 2.x.

Cheers,


frankdejonge:
Sorry, but v2 is no longer supported.

PR #1751, opened 2024-02-03
chore: run quality assurance regularly

tinect:
That way we can tackle problems earlier and faster, right?


frankdejonge:
@tinect thank you for all of the PR's I've merged them and released a new minor version that includes them all.

tinect:
Thank you, @frankdejonge

Issue #1660, opened 2023-04-26
SFTP connections - do we need to disconnect each time?


geebeetoo:
### Question

| Q                 | A       |
|-------------------|---------|
| Flysystem Version | flysystem-sftp-v3  |
| Adapter Name      | example |
| Adapter version   | x.y.z   |

When we submit several jobs via laravel queue, eventually after about 150 jobs we get "unable to access" - would that be because there are too many connections?

Is there such a thing as closing or disconnecting connections after every job?

Thanks

frankdejonge:
The issue does not list the required fields, nor does it share enough info to investigate.

There is a ConnectionProvider interface which allows you to control how a connection is established. You can use that to implement any connection/re-connect logic you may need.

ttruong15:
When the queue worker run it does not kill the established connection.  If there's 4 workers then there will be 4 established connection stay active.  Depending on the server if they have  a set timeout then you next call outside that windows will failed.  So if you try to connect to the server to receive some file you might get something like this 

_`shallow listing Reason: Connection closed by server`_

So my solution is to call disconnect explicitly everytime.
`$sftp = Storage::disk('sftp');`
`$gotSomeFiles = $sftp->files('data');`
`$sftp->getAdapter()->getProvideConnection()->disconnect();`

There's a ping() method you can use.  The ping will check the connection and re-established if necessary but once the job is done the connection still active.  I prefer to use disconnect so I know there's no stale connection floating around taking up resources.

frankdejonge:
@ttruong15 there is an interface you can implement for custom connectivity checking logic: src/PhpseclibV2/ConnectivityChecker.php

ttruong15:
@frankdejonge thanks that does work but it doesn't meet my needs.  That only check if the connection is active or not and then reconnect again.  However that will always leave the connection active.  

The Laravel queue worker when run via supervisor doesn't get kill so it never get to call the __desctruct() of the SSH2 so it can clean up the connection.  I have 4 workers which means there's 4 connections sitting idle wasting resources.  

I want to be able to connect to the server, do what i need to do and disconnect and clean up the connection.  Which means every time I access the server it have to reconnect/login again which I am fine with it.

Anyway, I think I solved my issue.  thanks for your help.,

Btw, if someone want to implement the connectivity checker the config need to be this
`sftp => [ `
`     ... `
`    'connectivityChecker' => new MyConnectivityChecker()`
`]`

hasfoug:
[...]

PR #1758, opened 2024-02-19
Add ability to manually disconnect for Sftp driver

hasfoug:
Exposed ConnectionProvider in SftpAdapter via geter, added disconnect method in SftpConnectionProvider

Related issues: 
https://github.com/thephpleague/flysystem/issues/1660
https://github.com/thephpleague/flysystem/issues/1757

### DIFF
--- a/src/PhpseclibV3/SftpAdapter.php
+++ b/src/PhpseclibV3/SftpAdapter.php
@@ -47,6 +47,11 @@ class SftpAdapter implements FilesystemAdapter
         $this->mimeTypeDetector = $mimeTypeDetector ?? new FinfoMimeTypeDetector();
     }
 
+    public function getConnectionProvider(): ConnectionProvider
+    {
+        return $this->connectionProvider;
+    }
+
     public function fileExists(string $path): bool
     {
         $location = $this->prefixer->prefixPath($path);
--- a/src/PhpseclibV3/SftpConnectionProvider.php
+++ b/src/PhpseclibV3/SftpConnectionProvider.php
@@ -47,6 +47,11 @@ class SftpConnectionProvider implements ConnectionProvider
         $this->connectivityChecker = $connectivityChecker ?? new SimpleConnectivityChecker();
     }
 
+    public function disconnect(): void
+    {
+        $this->connection?->disconnect();
+    }
+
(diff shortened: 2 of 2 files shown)


src/PhpseclibV3/SftpAdapter.php
```diff
@@ -47,6 +47,11 @@ public function __construct(
         $this->mimeTypeDetector = $mimeTypeDetector ?? new FinfoMimeTypeDetector();
     }
 
+    public function getConnectionProvider(): ConnectionProvider
```
frankdejonge, line 50: Flysystem is not a DI, this should be removed.

hasfoug: @frankdejonge But how then i should call disconnect from SftpAdapter if ConnectionProvider is private? Do you suggest that we should rather add a public disconnect method in SftpAdapter as well?

frankdejonge: @hasfoug the connection provider is injected in, so you can get the same instance somewhere else.

ttruong15: @frankdejonge unfortunately the connection provider is injected in but for Laravel the way it's call the connection provider is a local variable and there's no way to access it.

Here's where it's being called:

vendor/laravel/framework/src/Illuminate/Filesystem/FilesystemManager.php

`    /**
     * Create an instance of the sftp driver.
     *
     * @param  array  $config
     * @return \Illuminate\Contracts\Filesystem\Filesystem
     */
    public function createSftpDriver(array $config)
    {
        $provider = SftpConnectionProvider::fromArray($config);

        $root = $config['root'] ?? '/';

        $visibility = PortableVisibilityConverter::fromArray(
            $config['permissions'] ?? []
        );

        $adapter = new SftpAdapter($provider, $root, $visibility);

        return new FilesystemAdapter($this->createFlysystem($adapter, $config), $adapter, $config);
    }`
    
    So here $provider is a local variable so we have no way off accessing the provider.  So if i want to expose the provider in Laravel I have to implement my own Storage system just to override this method to expose the provider. Such a pain.

frankdejonge:
Superseded by #1763

PR #1766, opened 2024-03-14
Node20 actions upgrade

DannyvdSluijs:
This PR updates several `uses` in the GitHub Actions workflow in this repository to no longer run on Node16.

You can checkout https://github.blog/changelog/2024-03-07-github-actions-all-actions-will-run-on-node20-instead-of-node16-by-default/#what-you-need-to-do for more reasoning about the replacements.


frankdejonge:
Thank you!

PR #1767, opened 2024-03-14
Rename ContentMD5 to AddContentMD5 for AWS

alexndlm:

PR #1774, opened 2024-03-18
Added supported 'add_content_md5' option to AwsS3V3Adapter.php

mattbartel:
As of January 2023, the `aws-sdk-php` for S3 now supports a 'add_content_md5` option for multipart uploads. (See [this commit](https://github.com/aws/aws-sdk-php/commit/91e8d4376b6c1805a2a83e4bd75b02aa79ae971c) and the [related issue](https://github.com/aws/aws-sdk-php/issues/2256).) 

On S3 buckets with object locks enabled, setting the option to true asks the SDK to automatically compute and send the required http header as part of the request. Without the header, AWS will return a `400 - Invalid Request` with the message "Content-MD5 OR x-amz-checksum- HTTP header is required for Put Part requests with Object Lock parameters"

### DIFF
--- a/src/AwsS3V3/AwsS3V3Adapter.php
+++ b/src/AwsS3V3/AwsS3V3Adapter.php
@@ -74,6 +74,7 @@ class AwsS3V3Adapter implements FilesystemAdapter, PublicUrlGenerator, ChecksumP
      * @var string[]
      */
     public const MUP_AVAILABLE_OPTIONS = [
+        'add_content_md5',
         'before_upload',
         'concurrency',
         'mup_threshold',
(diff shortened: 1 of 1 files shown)


mattbartel:
Just noting my comment in the other pull request about this: https://github.com/thephpleague/flysystem/pull/1767#issuecomment-2016443380

alexndlm:
I have tested it - it works!

frankdejonge:
Im at a bachelor party right now, will look at this tomorrow

frankdejonge:
@mattbartel I understand adding the option, but removing the other option seems like 1) it's not needed and 2) a BC break for people using that option. Is it required for this? If yes, how so? Or otherwise, can you revert that part of the PR?

mattbartel:
That's what I thought originally too, but when I looked at the SDK code, I couldn't find any code that still referenced the ContentMD5 option, which is why I removed it. It seems like they eliminated it in favor of automatically calculating the hash. I will remove that portion of the PR for now and investigate further, then submit another PR if necessary.

frankdejonge:
I did some investigation on my own, the current option doesn't have an effect on the minimum version we require for the adapter, so the removal is safe.

PR #1779, opened 2024-04-04
Fix: AWS Copy SSE-C Options +  Send all options to copy instead of just params

amenophis:
Hi @frankdejonge,
This is the right MR with all changes.

I tested in with the official S3 SDK, not with AsyncAWS.
Not sure we have to do changes for params with AsyncAWS.

Thanks in advance for review :pray:

### DIFF
--- a/src/AsyncAwsS3/AsyncAwsS3Adapter.php
+++ b/src/AsyncAwsS3/AsyncAwsS3Adapter.php
@@ -73,9 +73,9 @@ class AsyncAwsS3Adapter implements FilesystemAdapter, PublicUrlGenerator, Checks
         'Tagging',
         'WebsiteRedirectLocation',
         'ChecksumAlgorithm',
-        'CopySSECustomerAlgorithm',
-        'CopySSECustomerKey',
-        'CopySSECustomerKeyMD5',
+        'CopySourceSSECustomerAlgorithm',
+        'CopySourceSSECustomerKey',
+        'CopySourceSSECustomerKeyMD5',
     ];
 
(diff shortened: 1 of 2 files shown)


frankdejonge:
The change to send all options broke the test suite, it's reverted.

amenophis:
@frankdejonge Forwarding params only make the copy didn't work from S3 to S3 with SSE-C enabled (the head issued by AWS SDK ObjectCopier ends with a 400 from AWS API)
Could you show me what tests are failling with the change ?

Thanks :pray:

frankdejonge:
@amenophis I found the error, it had to do with how MetadataDirective was handled (which is required for specifying new metadata during copy. I'm not sure how the options you corrected are forwarded, as metadata of otherwise, but you might need to use the config `'MetadataDirective' => 'REPLACE'` for what you're trying to achieve. Hope this helps, 3.27.0 is just released.

amenophis:
I will try this tomorrow.
Thank you for your research!

amenophis:
Hi @frankdejonge 
I checked the fix you have merged, it seems to fix our issue, but there is something that i didn't understand about the MetadataDirective.
I've added it in the adapter configuration with REPLACE value, and if i watch at the options sent to the client->copy() method, The MetadataDirective option in defined twice, at root with COPY value and in 'params' with REPLACE value).

I don't think this is the expected behaviour.

WDYT ?

amenophis:
Hi @frankdejonge 
WDYT about my last message ?

Thanks :pray:

PR #1788, opened 2024-04-27
Make MountManager extendable

thisispiers:


frankdejonge:
Just copy it and change it to your liking.

PR #1794, opened 2024-05-13
Add MongoDB GridFS Adapter

GromNaN:
[GridFS](https://www.mongodb.com/docs/manual/core/gridfs/) is a file system that uses MongoDB as storage. I would like to propose this adapter which uses the [MongoDB PHP Library](https://github.com/mongodb/mongo-php-library/).

Docs PR: https://github.com/thephpleague/flysystem/pull/1795

The [legacy GridFS Flysystem adapter](https://github.com/thephpleague/flysystem-gridfs) had some [significant number of installations](https://packagist.org/packages/league/flysystem-gridfs/stats), even if it used the legacy Mongo extension. It shows there's some interest.

Important notes:
- Directories does not exist in GridFS. I use the same technic as for S3 when explicitly creating a directory: an empty file with a trailing `/` is created. The creation of a file with a name ending in a `/` is therefore prohibited.
- Permissions are not supported by GridFS. When explicitly set, I store the visibility in the `metadata.flysystem_visibility` attribute of the file.
- GridFS' `contentType` attribute [is deprecated](https://github.com/mongodb/specifications/blob/master/source/gridfs/gridfs-spec.md#design-rationale). The attribute `metadata.contentType` is used instead.
- GridFS supports versioning by default. Writing to an existing file creates a new revision. Deleting removes all the revisions. Moving renames all the revisions.
- For GitHub Action, the [mongodb extension](https://pecl.php.net/package/mongodb) is installed and a mongodb server is started using  Docker Compose.
- I added some tests in the shared `FilesystemAdapterTestCase` in order to improve coverage on `mimeType` and `visibility`.

### DIFF
--- a/src/GridFS/GridFSAdapter.php
+++ b/src/GridFS/GridFSAdapter.php
@@ -1,0 +1,413 @@
+<?php
+
+declare(strict_types=1);
+
+namespace League\Flysystem\GridFS;
+
+use League\Flysystem\Config;
+use League\Flysystem\DirectoryAttributes;
+use League\Flysystem\FileAttributes;
+use League\Flysystem\FilesystemAdapter;
+use League\Flysystem\PathPrefixer;
+use League\Flysystem\UnableToCopyFile;
+use League\Flysystem\UnableToCreateDirectory;
+use League\Flysystem\UnableToDeleteFile;
+use League\Flysystem\UnableToMoveFile;
+use League\Flysystem\UnableToReadFile;
+use League\Flysystem\UnableToRetrieveMetadata;
+use League\Flysystem\UnableToSetVisibility;
+use League\Flysystem\UnableToWriteFile;
+use League\MimeTypeDetection\FinfoMimeTypeDetector;
+use League\MimeTypeDetection\MimeTypeDetector;
+use MongoDB\BSON\ObjectId;
+use MongoDB\BSON\Regex;
+use MongoDB\BSON\UTCDateTime;
+use MongoDB\Driver\Exception\Exception;
+use MongoDB\GridFS\Bucket;
+
+class GridFSAdapter implements FilesystemAdapter
+{
+    private const METADATA_DIRECTORY = 'flysystem_directory';
+    private const METADATA_VISIBILITY = 'flysystem_visibility';
+    private const METADATA_MIMETYPE = 'contentType';
+
+    private Bucket $bucket;
+
+    private PathPrefixer $prefixer;
+
+    private MimeTypeDetector $mimeTypeDetector;
+
+    public function __construct(
+        Bucket $bucket,
+        string $prefix = '',
+        ?MimeTypeDetector $mimeTypeDetector = null,
+    ) {
+        $this->bucket = $bucket;
+        $this->prefixer = new PathPrefixer($prefix);
+        $this->mimeTypeDetector = $mimeTypeDetector ?? new FinfoMimeTypeDetector();
+    }
+
+    public function fileExists(string $path): bool
+    {
+        $file = $this->findFile($path);
+
+        return $file !== null;
+    }
+
+    public function directoryExists(string $path): bool
+    {
+        // A directory exists if at least one file exists with a path starting with the directory name
+        $files = $this->listContents($path, true);
+
+        foreach ($files as $file) {
+            return true;
+        }
+
+        return false;
+    }
+
+    public function write(string $path, string $contents, Config $config): void
+    {
+        if (str_ends_with($path, '/')) {
+            throw UnableToWriteFile::atLocation($path, 'file path cannot end with a slash');
+        }
+
+        $filename = $this->prefixer->prefixPath($path);
+        $options = [
+            'metadata' => $config->get('metadata', []),
+        ];
+        if ($visibility = $config->get(Config::OPTION_VISIBILITY)) {
+            $options['metadata'][self::METADATA_VISIBILITY] = $visibility;
+        }
+        if (($mimeType = $config->get('mimetype')) || ($mimeType = $this->mimeTypeDetector->detectMimeType($path, $contents))) {
+            $options['metadata'][self::METADATA_MIMETYPE] = $mimeType;
+        }
+
+        try {
+            $stream = $this->bucket->openUploadStream($filename, $options);
+            fwrite($stream, $contents);
+            fclose($stream);
+        } catch (Exception $exception) {
+            throw UnableToWriteFile::atLocation($path, $exception->getMessage(), $exception);
+        }
+    }
+
+    public function writeStream(string $path, $contents, Config $config): void
+    {
+        if (str_ends_with($path, '/')) {
+            throw UnableToWriteFile::atLocation($path, 'file path cannot end with a slash');
+        }
+
+        $filename = $this->prefixer->prefixPath($path);
+        $options = [];
+        if ($visibility = $config->get(Config::OPTION_VISIBILITY)) {
+            $options['metadata'][self::METADATA_VISIBILITY] = $visibility;
+        }
+        if (($mimetype = $config->get('mimetype')) || ($mimetype = $this->mimeTypeDetector->detectMimeTypeFromPath($path))) {
+            $options['metadata'][self::METADATA_MIMETYPE] = $mimetype;
+        }
+
+        try {
+            $this->bucket->uploadFromStream($filename, $contents, $options);
+        } catch (Exception $exception) {
+            throw UnableToWriteFile::atLocation($path, $exception->getMessage(), $exception);
+        }
+    }
+
+    public function read(string $path): string
+    {
+        return stream_get_contents($this->readStream($path));
+    }
+
+    public function readStream(string $path)
+    {
+        if (str_ends_with($path, '/')) {
+            throw UnableToReadFile::fromLocation($path, 'file path cannot end with a slash');
+        }
+
+        $file = $this->findFile($path);
+
+        if ($file === null) {
+            throw UnableToReadFile::fromLocation($path, 'file does not exist');
+        }
+
+        try {
+            return $this->bucket->openDownloadStream($file['_id']);
+        } catch (Exception $exception) {
+            throw UnableToReadFile::fromLocation($path, $exception->getMessage(), $exception);
+        }
+    }
+
+    public function delete(string $path): void
+    {
+        if (str_ends_with($path, '/')) {
+            throw UnableToDeleteFile::atLocation($path, 'file path cannot end with a slash');
+        }
+
+        // Deleting a file that does not exist is no-op
+        while ($file = $this->findFile($path)) {
+            try {
+                $this->bucket->delete($file['_id']);
+            } catch (Exception $exception) {
+                throw UnableToDeleteFile::atLocation($path, $exception->getMessage(), $exception);
+            }
+        }
+    }
+
+    public function deleteDirectory(string $path): void
+    {
+        foreach ($this->listContents($path, true) as $file) {
+            $this->bucket->delete($file->extraMetadata()['_id']);
+        }
+    }
+
+    public function createDirectory(string $path, Config $config): void
+    {
+        $dirname = $this->prefixer->prefixDirectoryPath($path);
+
+        $options = [
+            'metadata' => $config->get('metadata', []) + [self::METADATA_DIRECTORY => true],
+        ];
+
+        if ($visibility = $config->get(Config::OPTION_VISIBILITY)) {
+            $options['metadata'][self::METADATA_VISIBILITY] = $visibility;
+        }
+
+        try {
+            $stream = $this->bucket->openUploadStream($dirname, $options);
+            fwrite($stream, '');
+            fclose($stream);
+        } catch (Exception $exception) {
+            throw UnableToCreateDirectory::atLocation($path, $exception->getMessage(), $exception);
+        }
+    }
+
+    public function setVisibility(string $path, string $visibility): void
+    {
+        $file = $this->findFile($path);
+
+        if ($file === null) {
+            throw UnableToSetVisibility::atLocation($path, 'file does not exist');
+        }
+
+        try {
+            $this->bucket->getFilesCollection()->updateOne(
+                ['_id' => $file['_id']],
+                ['$set' => ['metadata.' . self::METADATA_VISIBILITY => $visibility]],
+            );
+        } catch (Exception $exception) {
+            throw UnableToSetVisibility::atLocation($path, $exception->getMessage(), $exception);
+        }
+    }
+
+    public function visibility(string $path): FileAttributes
+    {
+        $file = $this->findFile($path);
+
+        if ($file === null) {
+            throw UnableToRetrieveMetadata::mimeType($path, 'file does not exist');
+        }
+
+        return new FileAttributes($path, null, $file['metadata'][self::METADATA_VISIBILITY] ?? null);
+    }
+
+    public function mimeType(string $path): FileAttributes
+    {
+        if (str_ends_with($path, '/')) {
+            throw UnableToRetrieveMetadata::mimeType($path, 'file path cannot end with a slash');
+        }
+
+        $file = $this->findFile($path);
+
+        if ($file === null) {
+            throw UnableToRetrieveMetadata::mimeType($path, 'file does not exist');
+        }
+
+        if ( ! isset($file['metadata'][self::METADATA_MIMETYPE])) {
+            throw UnableToRetrieveMetadata::mimeType($path, 'unknown');
+        }
+
+        return new FileAttributes($path, null, null, null, $file['metadata'][self::METADATA_MIMETYPE]);
+    }
+
+    public function lastModified(string $path): FileAttributes
+    {
+        if (str_ends_with($path, '/')) {
+            throw UnableToRetrieveMetadata::lastModified($path, 'file path cannot end with a slash');
+        }
+
+        $file = $this->findFile($path);
+
+        if ($file === null) {
+            throw UnableToRetrieveMetadata::lastModified($path, 'file does not exist');
+        }
+
+        return new FileAttributes($path, null, null, $file['uploadDate']->toDateTime()->getTimestamp());
+    }
+
+    public function fileSize(string $path): FileAttributes
+    {
+        if (str_ends_with($path, '/')) {
+            throw UnableToRetrieveMetadata::fileSize($path, 'file path cannot end with a slash');
+        }
+
+        $file = $this->findFile($path);
+
+        if ($file === null) {
+            throw UnableToRetrieveMetadata::fileSize($path, 'file does not exist');
+        }
+
+        return new FileAttributes($path, $file['length']);
+    }
+
+    public function listContents(string $path, bool $deep): iterable
+    {
+        $path = $this->prefixer->prefixDirectoryPath($path);
+
+        $pathdeep = 0;
+        $pipeline = [];
+        if ($path !== '') {
+            $pathdeep = substr_count($path, '/');
+            // Exclude files that do not start with the expected path
+            $pipeline[] = ['$match' => ['filename' => ['$regex' => new Regex('^' . preg_quote($path))]]];
+        }
+        // Get the last revision of each file
+        $pipeline[] = ['$sort' => ['filename' => 1, 'uploadDate' => -1]];
+
+        if ($deep === false) {
+            $pipeline[] = ['$addFields' => ['splitpath' => ['$split' => ['$filename', '/']]]];
+            $pipeline[] = ['$group' => [
+                // The same name could be used as a filename and as part of the path of other files
+                '_id' => [
+                    'basename' => ['$arrayElemAt' => ['$splitpath', $pathdeep]],
+                    'isDir' => ['$ne' => [['$size' => '$splitpath'], $pathdeep + 1]],
+                ],
+                // Get the metadata of the last revision of each file
+                'file' => ['$first' => '$$ROOT'],
+                // The "lastModified" date is the date of the last uploaded file in the directory
+                'uploadDate' => ['$max' => '$uploadDate'],
+            ]];
+
+            $files = $this->bucket->getFilesCollection()->aggregate(
+                $pipeline,
+                ['typeMap' => ['root' => 'array', 'document' => 'array', 'array' => 'array']]
+            );
+
+            foreach ($files as $file) {
+                if ($file['_id']['isDir']) {
+                    yield new DirectoryAttributes(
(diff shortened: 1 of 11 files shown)


src/GridFS/.github/FUNDING.yml
```diff
@@ -0,0 +1,2 @@
+github: [frankdejonge]
+custom: "https://ecologi.com/frankdejonge"
```
jmikola: Is this file relevant? It doesn't appear in other providers under `src/`: https://github.com/thephpleague/flysystem/tree/3.x/src

src/GridFS/GridFSAdapter.php
```diff
@@ -0,0 +1,404 @@
...
+            $options['metadata'][self::METADATA_VISIBILITY] = $visibility;
+        }
+        if (($mimetype = $config->get('mimetype')) || $mimetype = $file['metadata'][self::METADATA_MIMETYPE] ?? null) {
+            $options['metadata'][self::METADATA_MIMETYPE] = $mimetype;
+        }
+
+        try {
+            $stream = $this->bucket->openDownloadStream($file['_id']);
+            $this->bucket->uploadFromStream($this->prefixer->prefixPath($destination), $stream, $options);
+        } catch (Exception $exception) {
```
jmikola, line 384: Just confirming: this is intended to catch _any_ exception from PHPC _or_ PHPLIB (noting the PHPLIB exceptions all extend the PHPC interface as well). Basically, anything from this provider gets proxied by a Flysystem exception.

src/GridFS/GridFSAdapter.php
```diff
@@ -0,0 +1,404 @@
...
+    public function move(string $source, string $destination, Config $config): void
+    {
+        $file = $this->findFile($source);
+
+        if ($file === null) {
+            throw UnableToMoveFile::because('file does not exist', $source, $destination);
+        }
+
+        try {
+            $this->bucket->getFilesCollection()->updateMany(
```
jmikola: Noted that this is the same approach used in https://github.com/mongodb/mongo-php-library/pull/1207 ([PHPLIB-1324](https://jira.mongodb.org/browse/PHPLIB-1324)), despite there being [no standard approach documented in the GridFS spec](https://jira.mongodb.org/browse/DRIVERS-2807).

src/GridFS/GridFSAdapter.php
```diff
@@ -0,0 +1,404 @@
...
+
+    public function listContents(string $path, bool $deep): iterable
+    {
+        $path = $this->prefixer->prefixDirectoryPath($path);
+
+        $pathdeep = 0;
+        $pipeline = [];
+        if ($path !== '') {
+            $pathdeep = substr_count($path, '/');
+            // Exclude files that does not start with the expected path
```
jmikola:
```suggestion
            // Exclude files that do not start with the expected path
```

Subject/verb agreement.

src/GridFS/GridFSAdapter.php
```diff
@@ -0,0 +1,404 @@
...
+                        $file['length'],
+                        $file['metadata'][self::METADATA_VISIBILITY] ?? null,
+                        $file['uploadDate']->toDateTime()->getTimestamp(),
+                        $file['metadata'][self::METADATA_MIMETYPE] ?? null,
+                        $file,
+                    );
+                }
+            }
+        } else {
+            // Get the last revision of each file name
```
jmikola:
```suggestion
            // Get the metadata of the last revision of each file
```

For consistency with the earlier pipeline construction. I assume these should be consistent.

src/GridFS/GridFSAdapter.php
```diff
@@ -0,0 +1,404 @@
...
+    }
+
+    public function deleteDirectory(string $path): void
+    {
+        foreach ($this->listContents($path, true) as $file) {
+            $this->bucket->delete($file->extraMetadata()['_id']);
+        }
+    }
+
+    public function createDirectory(string $path, Config $config): void
```
jmikola, line 183: Is this the same approach used in other storage drivers that don't have a native concept of directories (e.g. S3)? I wonder if it'd be sensible to add a boolean metadata field to identify this as a directory, as there is otherwise nothing to distinguish this from an empty file.

I'm thinking more about other tooling that may work with the same GridFS bucket, even if such a field isn't directly usable by this adapter.

src/GridFS/LICENSE
```diff
@@ -0,0 +1,19 @@
+Copyright (c) 2013-2023 Frank de Jonge
```
jmikola:
```suggestion
Copyright (c) 2024-present Frank de Jonge
```

Assuming this was copied from elsewhere.

src/GridFS/composer.json
```diff
@@ -0,0 +1,21 @@
...
+    "autoload": {
+        "psr-4": {
+            "League\\Flysystem\\GridFS\\": ""
+        }
+    },
+    "require": {
+        "php": "^8.0.2",
+        "ext-mongodb": "^1.15",
+        "league/flysystem": "^3.10.0",
+        "mongodb/mongodb": "^1.15"
```
jmikola: Is there something particular in the [1.15.0 release](https://jira.mongodb.org/secure/ReleaseNote.jspa?projectId=12483&version=34330) that is required by this adapter, or was this an arbitrary choice?

src/GridFS/.github/FUNDING.yml
GromNaN: Removed. I don't know how it landed here.

src/GridFS/GridFSAdapter.php
GromNaN: Using this adapter, trailing `/` are forbidden. So a file with a trailing `/` is a directory.

I just added a `metadata.flysystem_directory` as marker.

src/GridFS/GridFSAdapter.php
GromNaN: Yes, flysystem adapters wrap all the exceptions.

src/GridFS/composer.json
GromNaN: Nothing specific. I didn't have the courage to go looking for the oldest version that supported the GridFS features required.

GromNaN: I tested version by version, starting with 1.0, and in the end it was quick because 1.2 seems to be compatible, with https://github.com/mongodb/mongo-php-library/commit/a448b609bdf6e9f0a5fb8a9d0cdd8583b6502245

frankdejonge:
Hi all, I'm out this week but will get on this ASAP to get this out. Thanks for the effort and the interest to contribute. Would there be anybody interested from your end to do continued maintenance in this adapter? And if so, who? Then I know who to rope in with issues.

GromNaN:
Hello Frank, we are a team of currently 3 devs, led by @alcaeus. We use an open [Jira](https://jira.mongodb.org/browse/PHPORM-66?jql=project%20%3D%20PHPORM%20AND%20component%20%3D%20Flysystem) to organise our work and Flysystem is one of the packages we want to maintain.

src/GridFS/composer.json
jmikola: The pedant in me appreciates using the lowest necessary dependency, but we'd probably also be fine to stick with whatever overlaps with Flysystem's PHP language requirement. Up to you.

src/GridFS/GridFSAdapter.php
```diff
@@ -0,0 +1,413 @@
...
+        try {
+            $this->bucket->uploadFromStream($filename, $contents, $options);
+        } catch (Exception $exception) {
+            throw UnableToWriteFile::atLocation($path, $exception->getMessage(), $exception);
+        }
+    }
+
+    public function read(string $path): string
+    {
+        return stream_get_contents($this->readStream($path));
```
frankdejonge: in some PHP environments, the stream opened is not automatically closed, an implementation like to following is needed:

```php
    public function read(string $path): string
    {
        $readStream = $this->readStream($path);
        $contents = stream_get_contents($readStream);
        fclose($readStream);

        return $contents;
    }
```

src/GridFS/GridFSAdapter.php
```diff
@@ -0,0 +1,413 @@
...
+    }
+
+    public function delete(string $path): void
+    {
+        if (str_ends_with($path, '/')) {
+            throw UnableToDeleteFile::atLocation($path, 'file path cannot end with a slash');
+        }
+
+        // Deleting a file that does not exist is no-op
+        while ($file = $this->findFile($path)) {
```
frankdejonge: 1) why is this a while loop?
2) this block may be affected by a race condition when two deletions are happening concurrently. In other implementations we try to delete and ignore the exception if it was caused by a "file not found" situation. Can we inspect the exception to see if occurs and ignore the error when that's the case?

src/GridFS/GridFSAdapter.php
```diff
@@ -0,0 +1,413 @@
...
+
+    public function visibility(string $path): FileAttributes
+    {
+        $file = $this->findFile($path);
+
+        if ($file === null) {
+            throw UnableToRetrieveMetadata::mimeType($path, 'file does not exist');
+        }
+
+        return new FileAttributes($path, null, $file['metadata'][self::METADATA_VISIBILITY] ?? null);
```
frankdejonge: the last constructor parameter is an "extra metadata" field, I'd recommend putting the entire metadata value there

frankdejonge: you can also map everything in one go, defaulting values to null in a central `metadata` method and only check if the value is there in the corresponding getter, this way you map once and check for expected values per case. WDYT?

src/GridFS/GridFSAdapter.php
```diff
@@ -0,0 +1,413 @@
...
+            foreach ($files as $file) {
+                if ($file['_id']['isDir']) {
+                    yield new DirectoryAttributes(
+                        $this->prefixer->stripDirectoryPrefix($path . $file['_id']['basename']),
+                        null,
+                        $file['uploadDate']->toDateTime()->getTimestamp(),
+                    );
+                } else {
+                    $file = $file['file'];
+                    yield new FileAttributes(
```
frankdejonge: this mapping, if extracted into a method could be the implementation of the before mentioned `metadata` method

GromNaN: Good idea, I factorized it with the `mapFileAttributes` function.

src/GridFS/GridFSAdapter.php
GromNaN: The while was necessary to remove all the revisions of the file name. I replaced the code with a "find all revisions" followed by "delete" on each revision. Good point about race-condition, I added a try-catch.

src/GridFS/GridFSAdapter.php
GromNaN: Updated to use `try...finally`

src/GridFS/GridFSAdapter.php
```diff
@@ -0,0 +1,443 @@
...
+        if (str_ends_with($path, '/')) {
+            throw UnableToWriteFile::atLocation($path, 'file path cannot end with a slash');
+        }
+
+        $filename = $this->prefixer->prefixPath($path);
+        $options = [];
+        if ($visibility = $config->get(Config::OPTION_VISIBILITY)) {
+            $options['metadata'][self::METADATA_VISIBILITY] = $visibility;
+        }
+        if (($mimetype = $config->get('mimetype')) || ($mimetype = $this->mimeTypeDetector->detectMimeTypeFromPath($path))) {
```
jmikola, line 115: I noticed this was done in `write()` as well, but what's the point of inferring a mime type from the destination filename? Shouldn't this be inferred from the actual contents, like [file(1)](https://man7.org/linux/man-pages/man1/file.1.html)?

Maybe this doesn't matter at all since `detectMimeTypeFromPath()` is only fall-back behavior if an explicit `mimetype` option isn't specified. I'm assuming the Config object is a container for per-operation options and not some common configuration for the entire library.

src/GridFS/GridFSAdapter.php
```diff
@@ -0,0 +1,443 @@
...
+        }
+
+        $file = $this->findFile($path);
+
+        if ($file === null) {
+            throw UnableToReadFile::fromLocation($path, 'file does not exist');
+        }
+
+        try {
+            return $this->bucket->openDownloadStream($file['_id']);
```
jmikola: Why not [`openDownloadStreamByName()`](https://www.mongodb.com/docs/php-library/current/reference/method/MongoDBGridFSBucket-openDownloadStreamByName/)?

src/GridFS/GridFSAdapter.php
```diff
@@ -0,0 +1,443 @@
...
+    public function createDirectory(string $path, Config $config): void
+    {
+        $dirname = $this->prefixer->prefixDirectoryPath($path);
+
+        $options = [
+            'metadata' => $config->get('metadata', []) + [self::METADATA_DIRECTORY => true],
+        ];
+
+        if ($visibility = $config->get(Config::OPTION_VISIBILITY)) {
+            $options['metadata'][self::METADATA_VISIBILITY] = $visibility;
```
jmikola, line 192: What's the purpose of this option? I don't follow why it's only injected into `metadata` when truthy. I noted this same logic appears earlier in the `write` and `writeStream` methods.

It seems contrary to the logic in `setVisibility()` below, which could end up storing a true _or_ false value to the field.

src/GridFS/GridFSAdapter.php
```diff
@@ -0,0 +1,443 @@
...
+
+        if ($file === null) {
+            throw UnableToRetrieveMetadata::mimeType($path, 'file does not exist');
+        }
+
+        if ( ! isset($file['metadata'][self::METADATA_MIMETYPE])) {
+            throw UnableToRetrieveMetadata::mimeType($path, 'unknown');
+        }
+
+        return new FileAttributes($path, null, null, null, $file['metadata'][self::METADATA_MIMETYPE]);
```
jmikola: Should this use `mapFileAttributes()`?

src/GridFS/GridFSAdapter.php
```diff
@@ -0,0 +1,443 @@
...
+
+        $pathdeep = 0;
+        $pipeline = [];
+        if ($path !== '') {
+            $pathdeep = substr_count($path, '/');
+            // Exclude files that do not start with the expected path
+            $pipeline[] = ['$match' => ['filename' => new Regex('^' . preg_quote($path))]];
+        }
+        // Get the last revision of each file
+        $pipeline[] = ['$sort' => ['filename' => 1, 'uploadDate' => -1]];
```
jmikola: Is `listContents()` always expected to sort on the filename? Just curious since this isn't mentioned in the comment.

src/GridFS/GridFSAdapter.php
```diff
@@ -0,0 +1,443 @@
...
+                // The same name could be used as a filename and as part of the path of other files
+                '_id' => [
+                    'basename' => ['$arrayElemAt' => ['$splitpath', $pathdeep]],
+                    'isDir' => ['$ne' => [['$size' => '$splitpath'], $pathdeep + 1]],
+                ],
+                // Get the metadata of the last revision of each file
+                'file' => ['$first' => '$$ROOT'],
+                // The "lastModified" date is the date of the last uploaded file in the directory
+                'uploadDate' => ['$max' => '$uploadDate'],
+            ]];
```
jmikola, line 305: Does this pipeline correctly filter out documents _above_ the target directory?

Consider:

```
> db.foo.insertOne({filename:'/a/file.txt'})
> db.foo.insertOne({filename:'/a/b/file.txt'})
> db.foo.insertOne({filename:'/a/b/c/file.txt'})
> db.foo.aggregate([
  {$addFields:{splitpath:{$split:['$filename','/']}}},
  {$group:{
    _id:{
      basename:{$arrayElemAt:['$splitpath',3]},
      isDir:{$ne:[{$size:'$splitpath'},4]}
    },
    file: {'$first': '$$ROOT'}
  }}
])
```

The above pipeline uses `$pathdeep` of 3 and yields the following:

```
[
  {
    _id: { isDir: true },
    file: {
      _id: ObjectId("664b625ca81bc8b8d615421c"),
      filename: '/a/file.txt',
      splitpath: [ '', 'a', 'file.txt' ]
    }
  },
  {
    _id: { basename: 'file.txt', isDir: false },
    file: {
      _id: ObjectId("664b6260a81bc8b8d615421d"),
      filename: '/a/b/file.txt',
      splitpath: [ '', 'a', 'b', 'file.txt' ]
    }
  },
  {
    _id: { basename: 'c', isDir: true },
    file: {
      _id: ObjectId("664b6263a81bc8b8d615421e"),
      filename: '/a/b/c/file.txt',
      splitpath: [ '', 'a', 'b', 'c', 'file.txt' ]
    }
  }
]
```

I was consistent about using a leading `/` in the filenames here, but I suppose that could vary in practice (especially if Flysystem isn't used exclusively for IO on the bucket).

src/GridFS/GridFSAdapter.php
```diff
@@ -0,0 +1,443 @@
...
+    }
+
+    public function listContents(string $path, bool $deep): iterable
+    {
+        $path = $this->prefixer->prefixDirectoryPath($path);
+
+        $pathdeep = 0;
+        $pipeline = [];
+        if ($path !== '') {
+            $pathdeep = substr_count($path, '/');
```
jmikola, line 288: Does this function behave consistently irrespective of a trailing `/` in `$path`?

src/GridFS/GridFSAdapter.php
```diff
@@ -0,0 +1,443 @@
...
+use MongoDB\GridFS\Bucket;
+use MongoDB\GridFS\Exception\FileNotFoundException;
+
+/**
+ * @phpstan-type GridFile array{_id:ObjectId, length:int, chunkSize:int, uploadDate:UTCDateTime, filename:string, metadata?:array{contentType?:string, flysystem_visibility?:string}}
+ */
+class GridFSAdapter implements FilesystemAdapter
+{
+    private const METADATA_DIRECTORY = 'flysystem_directory';
+    private const METADATA_VISIBILITY = 'flysystem_visibility';
```
jmikola, line 36: What's the purpose of this metadata field? I only see it being set during write operations and incorporated into returned FileAttributes or DirectoryAttributes.

src/GridFS/GridFSAdapter.php
GromNaN: Good point, I've changed it.

src/GridFS/GridFSAdapter.php
GromNaN: I don't want to store visibility unless it's explicitly set. `setVisibility` requires a value. For `write`, `writeStream`, `createDirectory` and `copy`, this is an option.

@frankdejonge Since there's no concept of file-level permission in GridFS, would it be better to disable the functionality altogether, as is done for WebDAV?

src/GridFS/GridFSAdapter.php
GromNaN: I don't think there's any specification on this point. The idea here is to make full use of [the index](https://github.com/mongodb/mongo-php-library/blob/cbc8104c0b2c32b7cf572ff759324c872e8dc63a/src/GridFS/CollectionWrapper.php#L324) on `filename` and `uploadDate`.

src/GridFS/GridFSAdapter.php
GromNaN: There is a `$match` stage to filter-out files that are not in the specified directory.
```ts
db.foo.aggregate([
  {$match:{filename:/^\/a\/b\/c\//}},
  {$addFields:{splitpath:{$split:['$filename','/']}}},
  {$group:{
    _id:{
      basename:{$arrayElemAt:['$splitpath',3]},
      isDir:{$ne:[{$size:'$splitpath'},4]}
    },
    file: {'$first': '$$ROOT'}
  }}
])
```

```ts
[
  {
    _id: { basename: 'c', isDir: true },
    file: {
      _id: ObjectId("664b706bf6dc6882f4ae8b93"),
      filename: '/a/b/c/file.txt',
      splitpath: [ '', 'a', 'b', 'c', 'file.txt' ]
    }
  }
]
```

src/GridFS/GridFSAdapter.php
GromNaN: `PathPrefixer` adds the trailing slash: https://github.com/thephpleague/flysystem/blob/0487c25a61e9b17ce82559af9bc1fc7934ee53b0/src/PathPrefixer.php#L48
So we always have a leading and a trailing slash.

src/GridFS/GridFSAdapter.php
jmikola: Noted. Thanks for adding a comment.

src/GridFS/GridFSAdapter.php
GromNaN: The content can only be used when the file is created with `write`. The methods `writeStream` and `copy` can't access the contents of the file, only its name.
We could also use fallback only when reading, but in the same way we don't know the contents of the file when the metadata are read.

Yes, `Config` is set per operation and contains the metadata `mimetype` and `visibility`.

src/GridFS/GridFSAdapter.php
GromNaN: Related to https://github.com/thephpleague/flysystem/pull/1794#discussion_r1606864392

src/GridFS/GridFSAdapter.php
jmikola: Thanks. Missed that when testing locally.

jmikola (approved):
LGTM. I'll defer to you both to decide whether to remove the the visibility option (https://github.com/thephpleague/flysystem/pull/1794/files#r1606938015).

frankdejonge:
Looks like the test case `GridFSAdapterTest::move_all_revisions` is failing.

GromNaN:
> Looks like the test case `GridFSAdapterTest::move_all_revisions` is failing.

Where do you see that? It seems to work in GitHub Actions: https://github.com/thephpleague/flysystem/actions/runs/9161336038/job/25242807530?pr=1794

frankdejonge:
@GromNaN this is an example; https://github.com/thephpleague/flysystem/actions/runs/9180056390/job/25244027378?pr=1794 might be because of eventual consistency in MongoDB? Is that a thing? Previous run had more errors on the same test case.

GromNaN:
Ok, I suspect that it's because 2 files are created in the same millisecond and revisions are sorted using this date.

GromNaN:
I updated the `readPreference`, so it should be OK.

frankdejonge:
@GromNaN still one failing, is there another way to deal with this or to configure it? https://github.com/GromNaN/flysystem/actions/runs/9181466558/job/25248713035

GromNaN:
It seems that my previous guess was correct but the fix incorrectly implemented. I added `usleep(1000)` (1ms) between each insertion of a revision to have a different `uploadDate`.

PR #1795, opened 2024-05-14
Add docs for GridFS adapter

GromNaN:
Docs for https://github.com/thephpleague/flysystem/pull/1794

### DIFF
--- a/_data/menu_v2.yml
+++ b/_data/menu_v2.yml
@@ -26,7 +26,7 @@ Official Adapters:
     AsyncAws S3: '/docs/adapter/async-aws-s3/'
     Azure Blob Storage: '/docs/adapter/azure-blob-storage/'
     Google Cloud Storage: '/docs/adapter/google-cloud-storage/'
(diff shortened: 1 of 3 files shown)


docs/adapter/mongodb-gridfs.md
```diff
@@ -0,0 +1,69 @@
...
+permalink: /docs/adapter/mongodb-gridfs/
+redirect_from:
+  - /v2/docs/adapter/mongodb-gridfs/
+  - /docs/adapter/gridfs/
+---
+
+## Installation
+
+```bash
+composer require league/mongodb-gridfs:^3.0
```
frankdejonge: this misses the flysystem infix. Come to think of it, we could also name it `league/flysystem-gridfs` in line with the v1 adapter and unabandon it. WDYT?

GromNaN: Ok to keep the old name. Updated.

PR #1816, opened 2024-09-09
Fixed int type checking error when custom ftp port is set

rahulhaque:
Using in Laravel 10 and when FTP_PORT is set in `.env` file, it throws following error. Other drivers may have this type of issue but haven't checked yet.
Screenshot from 2024-09-09 15-55-32


rahulhaque:
You close the pull request but the issue exists. Need to dig a little bit more because when `./config/filesystems.php` does not have `port` set for ftp, it throws below error. I think this is also related #1812 as he removed the variable type.
image

rahulhaque:
Please, have a look at the [Final fix](https://github.com/thephpleague/flysystem/pull/1816/commits/bfb65bfee994cdab892625507cb96e796e4d3683) to see if the changes suits your taste.

PR #1822, opened 2024-10-11
Fix ACLs for copy method

raczekm:


Korbeil:
Considering #1876 I think we could close this pull request.

PR #1831, opened 2024-11-20
Add ability to delete multiple S3 objects

aneeskhan47:
Hi,

This PR enables you to delete multiple objects from a bucket using a single HTTP request. If you know the object keys that you want to delete, then this operation provides a suitable alternative to sending individual delete requests, reducing per-request overhead.

I have modified the existing delete method to support both single-string path and array of paths.

AWS S3 docs: https://docs.aws.amazon.com/AmazonS3/latest/API/API_DeleteObjects.html

@frankdejonge

### DIFF
--- a/src/AwsS3V3/AwsS3V3Adapter.php
+++ b/src/AwsS3V3/AwsS3V3Adapter.php
@@ -221,15 +221,31 @@ class AwsS3V3Adapter implements FilesystemAdapter, PublicUrlGenerator, ChecksumP
         return $resource;
     }
 
-    public function delete(string $path): void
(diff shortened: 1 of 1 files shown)


aneeskhan47:
@frankdejonge any insights on this :) ?

aneeskhan47:
@frankdejonge 

Alt Text

frankdejonge:
The ability to delete multiple object could land, but not in V3 and not only for S3. So this PR will not be merged.

aneeskhan47:
Thank you for your attention

PR #1847, opened 2025-01-24
FTPAdapter: Use more precise time for 'timestampsOnUnixListings' option

themark147:
What about using -T option for listing? For exact definition of time?

definition of -T option:

```
-T      When printing in the long (-l) format, display complete time information for the file, including month, day, hour, minute, second, and year.  The -D option gives even more control over the output
             format.  This option is not defined in IEEE Std 1003.1-2008 (“POSIX.1”).

```
definition of -l is quite weird and changing over time (after 6 months)

```
If the modification time of the file is more than 6 months in the past or future, and the -D or -T are not specified, then the year of the last modification is displayed in place of the hour and minute
     fields.
```

what do you think about that suggestion? thanks 😊

### DIFF
--- a/src/Ftp/FtpAdapter.php
+++ b/src/Ftp/FtpAdapter.php
@@ -356,7 +356,7 @@ class FtpAdapter implements FilesystemAdapter
             yield from $this->listDirectoryContentsRecursive($path);
         } else {
             $location = $this->prefixer()->prefixPath($path);
-            $options = $deep ? '-alnR' : '-aln';
+            $options = $deep ? '-alnRT' : '-alnT';
             $listing = $this->ftpRawlist($options, $location);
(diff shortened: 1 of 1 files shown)


themark147:
this is how look like files without -T option

```
-rw-r--r--@ 1 markuss  staff  0 May 17  2018 file_from_2018.csv
-rw-r--r--@ 1 markuss  staff  0 Jan 24 14:56 file_from_2025_today.csv
```

this is with -T option

```
-rw-r--r--@ 1 markuss  staff  0 May 17 12:48:00 2018 file_from_2018.csv
-rw-r--r--@ 1 markuss  staff  0 Jan 24 14:56:14 2025 file_from_2025_today.csv
```

themark147:
FTP dont support -T option

PR #1848, opened 2025-01-26
[GCS] Add option to stream reads

rickclephas:
Adds a `streamReads` option to the `GoogleCloudStorageAdapter` (similar to the `AwsS3V3Adapter`).

It makes sure that `readStream` will actually stream the read instead of downloading the whole file.
When set to `true` the behaviour will be the same as with the official [GCS StreamWrapper](https://github.com/googleapis/google-cloud-php/blob/d19f3cfb69a7799713ff2c9c6283a2d4870fc884/Storage/src/StreamWrapper.php#L270).

To preserve backwards compatibility the default value is set to `false`, this could be changed to `true` in a future version.


frankdejonge:
@rickclephas if you could rebase on latest 3.x, that will probably make the tests green again.

PR #1858, opened 2025-03-11
Fix minor typo on Path Prefixing Adapter Decorator page

elazar:
One of the Usage examples is missing an "i" in the "PathPrefixing" namespace segment.


frankdejonge:
@elazar much appreciated!

PR #1862, opened 2025-03-27
AsyncAws S3: URL-encode the CopySource argument 

stevenrombauts:
We ran into an issue moving objects that have non-ASCII characters in their name. AWS S3 expects the  `CopySource` argument to be URL-encoded. For example:

```php
$client = new AsyncAws\S3\S3Client([
    'accessKeyId' => '...',
    'accessKeySecret' => '...',
    'region' => 'eu-west-1',
]);

$adapter = new League\Flysystem\AsyncAwsS3\AsyncAwsS3Adapter($client, 'urlencode-test');

$copyFrom = 'buıld_Çöü.json';
$filesystem->write($copyFrom, '{"foo": "bar"}');
$filesystem->copy($copyFrom, 'buıld_Çöü-copy.json');
```

will result in this error:

```
PHP Fatal error:  Uncaught AsyncAws\Core\Exception\Http\ClientException: HTTP 404 returned for "https://urlencode-test.s3.eu-west-3.amazonaws.com/bu%C4%B1ld_%C3%87%C3%B6%C3%BC-copy.json".

Code:    NoSuchKey
Message: The specified key does not exist.
``` 

URL-encoding the `CopySource` argument makes it work fine. 

According to the [CopyObject docs](https://docs.aws.amazon.com/AmazonS3/latest/API/API_CopyObject.html#API_CopyObject_RequestParameters) the `x-amz-copy-source` header, which is set to the value of the `CopySource` argument in the S3 client, must be URL-encoded. 

Note: [this older PR](https://github.com/thephpleague/flysystem/pull/1302/commits) previously removed the `rawurlencode()` method, so maybe URL-encoding should be made configurable instead?

### DIFF
--- a/src/AsyncAwsS3/AsyncAwsS3Adapter.php
+++ b/src/AsyncAwsS3/AsyncAwsS3Adapter.php
@@ -357,7 +357,7 @@ class AsyncAwsS3Adapter implements FilesystemAdapter, PublicUrlGenerator, Checks
             'ACL' => $this->visibility->visibilityToAcl($visibility ?: 'private'),
             'Bucket' => $this->bucket,
             'Key' => $this->prefixer->prefixPath($destination),
-            'CopySource' => $this->bucket . '/' . $this->prefixer->prefixPath($source),
+            'CopySource' => rawurlencode($this->bucket . '/' . $this->prefixer->prefixPath($source)),
         ];
 
         try {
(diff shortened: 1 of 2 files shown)


frankdejonge:
@stevenrombauts could you add a test case with those character? That way we can prevent a regression.

stevenrombauts:
@frankdejonge Certainly, I've added that test now.

cfoehrdes:
We ran into the same issue today where we call copy for an S3 file with some special chars in the source path (`Zwei Compañeros/some-file.jpg`) and it fails with a 404 call to the source object. I applied the fix from this PR locally it also solved the issue for us.

PR #1864, opened 2025-04-17
Add option to disconnect on destruct

qdequippe:
When using flysystem in worker (never stopped) the connection stay active. Add an option to disconnect on object destruction.

### DIFF
--- a/src/PhpseclibV3/SftpAdapter.php
+++ b/src/PhpseclibV3/SftpAdapter.php
@@ -41,6 +41,7 @@ class SftpAdapter implements FilesystemAdapter
         ?VisibilityConverter $visibilityConverter = null,
         ?MimeTypeDetector $mimeTypeDetector = null,
         private bool $detectMimeTypeUsingPath = false,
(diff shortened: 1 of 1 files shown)


qdequippe:
thanks @frankdejonge

bendavies:
is there a way to configure this option when using the bundle?

bendavies:
i don't think so, i'll add

bendavies:
for now, a compiler pass with something like this should do it:

```php
    public function process(ContainerBuilder $container): void
    {
        $jencapStorageDefinition = $container->getDefinition('flysystem.adapter.foo.storage');

        if ($jencapStorageDefinition->getClass() === SftpAdapter::class) {
            $jencapStorageDefinition->setArgument('$disconnectOnDestruct', true);
        }
    }
```

maxhelias:
You can submit a PR on the bundle to support this option in the adapter here : https://github.com/thephpleague/flysystem-bundle/blob/3.x/src/Adapter/Builder/SftpAdapterDefinitionBuilder.php I'll be happy to merge it 😉

PR #1868, opened 2025-04-25
AsyncAws S3: Replace special characters by XML entity codes

stevenrombauts:
When calling `deleteDirectory()`, we want to make sure that special characters (`'`, `"`, `&`, `<` and `>`) are replaced by their XML entity code. If not, the following code will result in the error below:

```php
$client = new AsyncAws\S3\S3Client([
    'accessKeyId' => '...',
    'accessKeySecret' => '...',
    'region' => 'eu-west-1',
]);

$adapter = new League\Flysystem\AsyncAwsS3\AsyncAwsS3Adapter($client, 'urlencode-test');

$filesystem->write('dir-to-delete/\'"&<>.json', '{"foo": "bar"}');
$filesystem->deleteDirectory('dir-to-delete');
```

This will result in:

```
Warning: DOMDocument::createElement(): unterminated entity reference         <>.json in flysystem/vendor/async-aws/s3/src/ValueObject/ObjectIdentifier.php on line 124

Fatal error: Uncaught AsyncAws\Core\Exception\Http\ClientException: HTTP 400 returned for "https://urlencode-test.s3.eu-west-3.amazonaws.com/?delete=".

Code:    UserKeyMustBeSpecified
Message: User key must be specified.
```

Also see AWS docs here: https://docs.aws.amazon.com/AmazonS3/latest/userguide/object-keys.html#object-key-xml-related-constraints

### DIFF
--- a/src/AsyncAwsS3/AsyncAwsS3Adapter.php
+++ b/src/AsyncAwsS3/AsyncAwsS3Adapter.php
@@ -188,7 +188,7 @@ class AsyncAwsS3Adapter implements FilesystemAdapter, PublicUrlGenerator, Checks
             foreach ($result->getContents() as $item) {
                 $key = $item->getKey();
                 if (null !== $key) {
-                    $objects[] = new ObjectIdentifier(['Key' => $key]);
+                    $objects[] = $this->createObjectIdentifierForXmlRequest($key);
                 }
             }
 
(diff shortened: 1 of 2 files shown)


maxhelias (approved):

frankdejonge:
@stevenrombauts can you add a test case that verifies the error so we prevent regressions? The rest looks good 👍

stevenrombauts:
@frankdejonge Do you mean to verify the exception that can be thrown by this [new code](https://github.com/thephpleague/flysystem/pull/1868/files#diff-4eededd8a94ab09a2848ac6bc341c08c26fc522caa1550eae474777ca3ed94eeR534-R536), or do you mean something else?

frankdejonge:
@stevenrombauts yes, verify that the guard works. Essentially a test that fails when the check is removed.

stevenrombauts:
@frankdejonge Thanks, got it. I've added that test now, let me know if that works!

PR #1873, opened 2025-05-21
LocalFilesystemAdapter. Call clearstatcache in several methods

### LINKED ISSUE #1826: LocalFilesystemAdapter::fileSize(),fileExists(),directoryExists and possible other method uses cached values
### Bug Report

| Flysystem Version | 3.29.1   |
| Adapter Name      | LocalFilesystemAdapter |
| Adapter version   | 3.29.0  |

#### Summary

Repeat calls to `fileSize()` will return the same value if the file has been modified(by another process) between calls as the value is cached and not cleared by calling `clearstatcache()`.

The is more or less the same issue that has previously been fixed in #1035 for file size.

The same issue is present if another process deletes a file or a directory, then concurrent calls to `fileExists()` or `directoryExists()` will be cached if it has existed on disk during the process. So it will continue to return true, even if the file/director has been deleted on disk.

Looking at [https://www.php.net/manual/en/function.clearstatcache.php](clearstatcache)
Then this could be an issue for some of the other methods as well `lastModified()` and possible others

In v1 `fileSize()`, `getTimestamp()` was part of the `getMetaData()` method which called `clearstatcache()`
And the way to differentiate between a file and directory was to check the type inside the metadata

#### How to reproduce

Belows simulates what happens inside `LocalFileSystemAdapterfileSize::fileSize()` when another process writes to a file

FileSize:
```php
$file = 'fileSize.txt';
touch($file);
for($i = 0; $i < 5; $i++)
{
    clearstatcache();
    $beforeSize = filesize($file);

    exec("echo 1 >> {$file}"); // This simulates another process that writes to the file

    //clearstatcache(); // Uncomment to get the change in filesize
    $afterSize = filesize($file);

    if($beforeSize === $afterSize)
     {
        $strMessage = 'No change in file size.';
     }
    else
    {
        $strMessage = 'File size changed.';
    }

    echo "{$strMessage} Before: {$beforeSize} After: {$afterSize}".PHP_EOL;

    sleep(1);
}
```
FileExists:
```php
$file = 'filesize.txt';
touch($file);

$isFileBefore = is_file($file);

exec("rm -f {$file}");

$isFileAfter = is_file($file);

echo 'Before: '.($isFileBefore ? 'Exists' : 'Does Not Exist').' After: '.($isFileAfter ? 'Exists' : 'Does Not Exist').PHP_EOL;

clearstatcache();

echo "Clear stat cache".PHP_EOL;
echo "File exists: ".(is_file($file) ? 'Yes' : 'No').PHP_EOL;
```

pvandommelen:
Php's statcache caches the results of some file functions like `is_file` or `filesize`.

For the `visibility()` method, a `clearstatcache` is already used to clear this cache, but this is not the case for several other methods. 

Fixes #1826

### DIFF
--- a/src/Local/LocalFilesystemAdapter.php
+++ b/src/Local/LocalFilesystemAdapter.php
@@ -337,14 +337,14 @@ class LocalFilesystemAdapter implements FilesystemAdapter, ChecksumProvider
     public function fileExists(string $location): bool
     {
         $location = $this->prefixer->prefixPath($location);
-
+        clearstatcache();
         return is_file($location);
     }
 
     public function directoryExists(string $location): bool
     {
         $location = $this->prefixer->prefixPath($location);
-
+        clearstatcache();
         return is_dir($location);
     }
 
@@ -423,6 +423,7 @@ class LocalFilesystemAdapter implements FilesystemAdapter, ChecksumProvider
     public function lastModified(string $path): FileAttributes
     {
         $location = $this->prefixer->prefixPath($path);
+        clearstatcache();
         error_clear_last();
         $lastModified = @filemtime($location);
 
@@ -436,6 +437,7 @@ class LocalFilesystemAdapter implements FilesystemAdapter, ChecksumProvider
     public function fileSize(string $path): FileAttributes
     {
         $location = $this->prefixer->prefixPath($path);
+        clearstatcache();
         error_clear_last();
 
         if (is_file($location) && ($fileSize = @filesize($location)) !== false) {


frankdejonge:
Can you add a configuration option to opt-in to this behaviour?

pvandommelen:
Thanks for the quick reply! Do we really have to have a configuration option for this?

`clearstatcache` is already used in the visibility method, which would probably also be affected by such an option. Something similar is also implemented in `ensureDirectoryExists`.

As mentioned in the issue, this commit would mostly reapply an older fix (https://github.com/thephpleague/flysystem/commit/6f2211c3abf55984dfcb1a1539a362d2e14bb2ee). Though that's on a much older release.

If you'd still like a configuration option, I can add one. But I think we should then move the existing `clearstatcache` calls behind that check.

shyim:
Would you merge an PR to make this opt-in @frankdejonge?

This change makes stat-cache useless, and tbh when you call exec, the caller should clear stat-cache not the filesystem abstraction on EVERYTHING 😅

frankdejonge:
@shyim hi, yes that would work! 👍

pvandommelen:
No objections from my side to make it configurable.

I would maybe consider this option being opt-out instead of opt-in. Mostly because having no caching would be a reasonable default and there have even been calls to remove the statcache in php itself entirely (https://externals.io/message/115912).

Just to clarify our use case since it is different from the one in the related issue (calling `exec`). We run a file exists check in a while loop as a low-tech method to detect the status of an external process. For this it is important that there is no cache involved.

PR #1877, opened 2025-09-05
Fix: Make fetching metaData and the download and response functions work with SSE-C in AwsS3V3

jegelkraut:
The S3 Command "HeadObject" that is used to fetch a files MetaData(mimeType, size, ...) does not work right now when SSE-C is used. This is because when SSE-C is enabled "HeadObject" requires the SSEC headers(SSECustomerAlgorithm, SSECustomerKey) to be present. Since fetchFileMedata is also used in the download and response functions, it is not possible right now to download a file when SSE-C is enabled. This pull request fixes this in that it adds the options also to the "HeadObject" command as it is already done in the readObject function (aka "GetObject" command).


eltouf:
Hello @frankdejonge 

PR #1877 fixes a bug that is currently affecting our use of Flysystem. Would it be possible to merge it into the main branch?

Thank you.

frankdejonge:
Released as 3.32.0

PR #1880, opened 2025-10-02
feat(gcs): support filename prefix filtering in listContents

### LINKED ISSUE #1879: Support prefix filtering on listContents() for file name prefixes (not only “directory” prefixes)
Currently, the GoogleCloudStorageAdapter::listContents() method automatically appends a / to the provided path to build the prefix option for the underlying Bucket::objects() call:

`if (! empty($prefixedPath)) {
    $options = ['prefix' => sprintf('%s/', rtrim($prefixedPath, '/'))];
}`

This means that listContents('incoming') correctly lists objects under incoming/, but calling:
`listContents('incoming/prefix_file_', false)`

does not list objects whose names start with prefix_file_, because the adapter is forcing a “directory-like” behavior by appending /.

This is a limitation, because Google Cloud Storage natively supports prefix filtering for arbitrary filename prefixes, not just directory prefixes.

**Why this matters**
	•	This matches GCS’s native capabilities (prefix filtering is powerful & efficient server-side).
	•	It avoids listing entire buckets and filtering client-side for common patterns.
	•	It allows Flysystem users to use listContents() for real-world scenarios with file prefixes at scale.

mouize:
**Summary**

This PR adds support for arbitrary filename prefix filtering to the GoogleCloudStorageAdapter::listContents() method.

Currently, the adapter forces a trailing / on the prefix, which prevents listing objects by filename prefix.
Ex: my_folder/my_file_prefixes_

This PR removes the forced /, allowing both directory and filename prefix listing as supported by Google Cloud Storage natively.

Why
	•	GCS natively supports prefix filtering server-side.
	•	Current adapter prevents efficient listing of large buckets by filename prefix.
	•	This makes Flysystem closer to native GCS behavior.

Fixes #1879


frankdejonge:
Hi, this would be a breaking change on the behavioural contract so will not be merged as it. What _is_ possible, is to add a `listContentsAsPrefix` option (default to false) to allow people to opt-in to this behaviour.

mouize:
> Hi, this would be a breaking change on the behavioural contract so will not be merged as it. What _is_ possible, is to add a `listContentsAsPrefix` option (default to false) to allow people to opt-in to this behaviour.

Right ! Thx for the review. Just pushed a new update

PR #1883, opened 2025-11-10
add error_clear_last() to rename and copy

Phillip85:
All other methods which are using the error_get_last() function are already using error_clear_last().
This two methods are showing old errors when a unknown error happens:

```
Unable to move file from xxx/x/x.xml to xxx/x/x.xml.20251110082848.done, because file_get_contents(/var/www/html/.env): Failed to open stream: No such file or directory
```


frankdejonge:
Released as 3.30.2

PR #1885, opened 2025-12-02
feat: Add SFTP connection reset method

MizouziE:
## Adds
- A `League\Flysystem\PhpseclibV3\SftpConnectionProvider::resetConnection()` method which leverages existing public `disconnect()` and private `setupConnection()` methods.
- Test to assert that the initial connection from `provideConnection()` differs from that returned by `resetConnection()` and that subsequent calls to `provideConnection()` return the re-initiated connection.

## Reasoning
An application I work on has a Laravel/Symfony Command which is used to diagnose various connections to SFTP servers and we would prefer to keep the logs for each test action separate. Currently we need to flush and re-establish the connection for each step. This addition would allow us to configure the connection once and reset it for each test action.

### DIFF
--- a/src/PhpseclibV3/SftpConnectionProvider.php
+++ b/src/PhpseclibV3/SftpConnectionProvider.php
@@ -91,6 +91,13 @@ class SftpConnectionProvider implements ConnectionProvider
         }
     }
 
+    public function resetConnection(): SFTP
+    {
+        $this->disconnect();
+
+        return $this->setupConnection();
+    }
+
     private function setupConnection(): SFTP
     {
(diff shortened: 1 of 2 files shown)


SamMousa:
Can't you just call `->disconnect()` and then `->provideConnection()`? The implementation does almost the same as your private function `setupConnection` does almost the same as `provideConnection()` except no retries and no connectivity check.

I'd think extending the public API for test cases somewhere else that have alternatives is not ideal.

MizouziE:
Thanks for the feedback @SamMousa , to be honest it helped when I went back to assess the thing that caused me to make this PR in the first place and it was easier to implement the reset than I'd originally imagined!

PR #1892, opened 2026-02-14
Move league/flysystem-local to suggests

GautamMKGarg:
Removing flysystem-local from the required list, because if we are using different adaptor like "league/flysystem-sftp-v3", then we don't necessarily require "local" adaptor.


frankdejonge:
It's shipped by default, just don't use it, nothing is wrong.

PR #1893, opened 2026-03-16
Add support for multiple host fingerprints

sekiesel:
Previously, only a single host fingerprint string was accepted for server verification.
During rollouts the servers alternate between old and new host keys, which could lead to connection failures.

This change updates the behavior such that either a single string or an array of strings is accepted, allowing multiple valid fingerprints to be configured. The connection is accepted if any match is found.

This maintains full backwards compatibility with existing single fingerprint configurations.

on-behalf-of: @e-solutions-GmbH <info@esolutions.de>

### DIFF
--- a/src/PhpseclibV3/SftpConnectionProvider.php
+++ b/src/PhpseclibV3/SftpConnectionProvider.php
@@ -126,7 +126,7 @@ class SftpConnectionProvider implements ConnectionProvider
             : [$this->hostFingerprint];
 
         foreach ($expectedFingerprints as $expectedFingerprint) {
-            if (0 !== strcasecmp($expectedFingerprint, $fingerprint)) {
+            if (0 === strcasecmp($expectedFingerprint, $fingerprint)) {
                 return;
             }
         }
--- a/src/PhpseclibV3/SftpConnectionProvider.php
+++ b/src/PhpseclibV3/SftpConnectionProvider.php
@@ -39,7 +39,7 @@ class SftpConnectionProvider implements ConnectionProvider
         private bool $useAgent = false,
         private int $timeout = 10,
         private int $maxTries = 4,
-        private ?string $hostFingerprint = null,
+        private string|array|null $hostFingerprint = null,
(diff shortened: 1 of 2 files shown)


sekiesel:
Hello @frankdejonge, are there any chances we could merge this PR into the main branch? Thank you.

frankdejonge:
Hi @sekiesel this makes sense, I'll get this merged and released tonight.

src/PhpseclibV3/SftpConnectionProvider.php
```diff
@@ -121,10 +121,17 @@ private function checkFingerprint(SFTP $connection): void
...
 
         $fingerprint = $this->getFingerprintFromPublicKey($publicKey);
+        $expectedFingerprints = is_array($this->hostFingerprint)
+            ? $this->hostFingerprint
+            : [$this->hostFingerprint];
 
-        if (0 !== strcasecmp($this->hostFingerprint, $fingerprint)) {
-            throw UnableToEstablishAuthenticityOfHost::becauseTheAuthenticityCantBeEstablished($this->host);
+        foreach ($expectedFingerprints as $expectedFingerprint) {
+            if (0 !== strcasecmp($expectedFingerprint, $fingerprint)) {
```
frankdejonge: this should be `===` instead of `!==`, this is checking the inverse, so it break the control mechanism.

frankdejonge: @sekiesel can you look into this? CI is failing, I think I've spotted the error.

frankdejonge: Just as a side note, this would quite easily be caught if you had run the tests yourself before submitting the PR.

sekiesel: Good catch. I've updated the PR accordingly. However, I did not manage to setup the project locally, which is why the error has not been caught earlier.

frankdejonge:
Released as 3.33.0

PR #1905, opened 2026-07-01
Honor an explicit ACL override when copying in the AwsS3V3 adapter

### LINKED ISSUE #1904: AWS S3 adapter copy() fails when accessing bucket in a different account
### Bug Report

| Flysystem Version | 3.34.0  |
| Adapter Name      | league/flysystem-aws-s3-v3 |
| Adapter version   | 3.34.0   |

#### Summary

Since April 27, 2023 (per [S3 User Guide](https://docs.aws.amazon.com/AmazonS3/latest/userguide/WhatsNew.html)) new S3 buckets are created with "Bucket Owner Enforced" enabled, and ACLs disabled.

If you attempt to set an ACL on such a bucket, _including at time of object creation_, the behavior depends on whether you are a principal within the owning account. If you are, the call is silently ignored (although I've only tested with creation, not an explicit PutAcl). If you are a principal in another account, you will get an ACL error, even if you have permission to write to the bucket.

The Flysystem AWS S3 adaptor's `copy()` function explicitly sets an ACL to control visibility, triggering this behavior:

```
        try {
            $this->client->copy(
                $this->bucket,
                $this->prefixer->prefixPath($source),
                $this->bucket,
                $this->prefixer->prefixPath($destination),
=======>        $this->visibility->visibilityToAcl($visibility ?: 'private'),
                $options,
            );
        } catch (Throwable $exception) {
            throw UnableToCopyFile::fromLocationTo($source, $destination, $exception);
        }
```

By comparison, the `upload()` function allows overriding the ACL; passing the `bucket-owner-full-control` ACL allows the call to succeed (see related issue #1874, in particular my comment showing work-around).

As it stands, there is no way to run programs in multiple accounts that copy objects within a shared bucket.

Allowing ACLs on the target bucket is not a solution: the only choices that this function allows are `private` and `public-read`. The former would prevent sharing between accounts, the latter would expose the object to the entire world (ex a bucket policy preventing that).

#### How to reproduce

Prerequisites:

* Two AWS accounts. For this report, the account IDs are 123456789012 and 210987654321; your account IDs will vary.

Steps to reproduce:

* Create a bucket "example" in account 210987654321 (using the account/region namespacing feature).  Use default "Bucket Owner Enforced" enabled, and ACLs disabled 

* Add a bucket policy to this bucket that grants full access to account 123456789012:

  ```
  {
      "Version": "2012-10-17",
      "Statement": [
          {
              "Effect": "Allow",
              "Principal": {
                  "AWS": "arn:aws:iam::123456789012:root"
              },
              "Action": "s3:*",
              "Resource": [
                  "arn:aws:s3:::example-210987654321-us-east-2-an",
                  "arn:aws:s3:::example-210987654321-us-east-2-an/*"
              ]
          }
      ]
  } 
  ```

* Start a session using `php artisan tinker` in a project that has this filesystem installed, with credentials for a user in account 123456789012 that also has full S3 permissions.

* Create an instance of the adaptor:

  ```
  $client = new Aws\S3\S3Client([]);

  $adapter = new League\Flysystem\AwsS3V3\AwsS3V3Adapter($client, 'example-210987654321-us-east-2-an');

  ```

* Attempt to write a file to the bucket, letting the adaptor pick an ACL. This will fail (and is intended to demonstrate the base problem of cross-bucket access):

  ```
  $filesystem->write('foo.txt', 'blah blah blah');
  ```

  ```
  League\Flysystem\UnableToWriteFile  Unable to write file at location: foo.txt. Error executing "PutObject" on "https://example-210987654321-us-east-2-an.s3.us-east-2.amazonaws.com/foo.txt"; AWS HTTP error:
  AccessControlListNotSupported (client): The bucket does not allow ACLs.
  ```

* You can work-around this problem by explicitly providing the `bucket-owner-full-control` ACL:

  ```
  $filesystem->write('foo.txt', 'blah blah blah', [ 'params' => ['ACL' => 'bucket-owner-full-control']]);
  ```

* Verify that the object has been written to S3 (or trust the `null` return value).

* Attempt to copy that object to a different name. This will fail with a more generic error:

  ```
  $filesystem->copy('foo.txt', 'bar.txt');
  ```

  ```
  League\Flysystem\UnableToCopyFile  Unable to copy file from foo.txt to bar.txt.
  ```

* Attempt the same work-around of specifying `bucket-owner-full-control`. This fails with the same error (because you can't actually override the ACL in this call):

  ```
  $filesystem->copy('foo.txt', 'bar.txt', [ 'params' => ['ACL' => 'bucket-owner-full-control']]);
  ```

#### Recommendation

Allow an ACL override, as is done with the `upload()` function:

```
$acl = $options['params']['ACL'] ?? $this->determineAcl($config);
```

lazerg:
The AwsS3V3 adapter's `copy()` always derived the ACL from the resolved visibility and ignored an explicit `ACL` passed through the config, so there was no way to override it the way `upload()` already allows. This makes `copy()` honor an `ACL` from the options (e.g. `bucket-owner-full-control`), which is needed to copy objects on buckets with ACLs disabled or across accounts.

Fixes #1904

### DIFF
--- a/src/AwsS3V3/AwsS3V3Adapter.php
+++ b/src/AwsS3V3/AwsS3V3Adapter.php
@@ -441,6 +441,7 @@ class AwsS3V3Adapter implements FilesystemAdapter, PublicUrlGenerator, ChecksumP
 
         $options = $this->createOptionsFromConfig($config);
         $options['MetadataDirective'] = $config->get('MetadataDirective', 'COPY');
+        $acl = $options['params']['ACL'] ?? $this->visibility->visibilityToAcl($visibility ?: 'private');
 
         try {
             $this->client->copy(
@@ -448,7 +449,7 @@ class AwsS3V3Adapter implements FilesystemAdapter, PublicUrlGenerator, ChecksumP
                 $this->prefixer->prefixPath($source),
                 $this->bucket,
                 $this->prefixer->prefixPath($destination),
-                $this->visibility->visibilityToAcl($visibility ?: 'private'),
+                $acl,
                 $options,
             );
         } catch (Throwable $exception) {
--- a/src/AwsS3V3/AwsS3V3AdapterTest.php
+++ b/src/AwsS3V3/AwsS3V3AdapterTest.php
@@ -459,6 +459,26 @@ class AwsS3V3AdapterTest extends FilesystemAdapterTestCase
         });
     }
 
+    /**
+     * @test
+     */
+    public function copying_a_file_with_an_explicit_acl(): void
+    {
+        $adapter = $this->adapter();
+        $prefixer = new PathPrefixer(static::$adapterPrefix);
+        $prefixedPath = $prefixer->prefixPath('destination.txt');
+
+        $adapter->write('source.txt', 'contents', new Config());
+        $adapter->copy('source.txt', 'destination.txt', new Config(['ACL' => 'bucket-owner-full-control']));
+
+        $arguments = ['Bucket' => getenv('FLYSYSTEM_AWS_S3_BUCKET'), 'Key' => $prefixedPath];
+        $command = static::$s3Client->getCommand('GetObjectAcl', $arguments);
+        $response = static::$s3Client->execute($command)->toArray();
+        $permission = $response['Grants'][0]['Permission'];
+
+        self::assertEquals('FULL_CONTROL', $permission);
+    }
+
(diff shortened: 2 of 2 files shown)


lazerg:
@frankdejonge not sure why but CI never even registered a single check on this one across two sync passes, could you take a look when you're free?

frankdejonge:
Published as 3.35.2, thank you!

PR #1908, opened 2026-07-07
Add multiple fingerprint formats

ghazalfaraj:
Currently the way that the library determines the provider fingerprint type as either md5 or sha512 is a bit limited. We updated the code so the library should recognize and work with multiple fingerprint formats.


ghazalfaraj:
@frankdejonge Could you please review this pr?

ghazalfaraj:
@frankdejonge  Kind reminder to review this pr please?

PR #1919, opened 2026-08-17
Fix: create SFTP root directory when missing (#1832)

### LINKED ISSUE #1832: Root folder is not created when missing (sftp)
### Bug Report

| Flysystem Version | 3.29.1   |
| Adapter Name      | SftpAdapter |
| Adapter version   | 3.29.0   |

#### Summary

Root folder is not created when is missing and UnableToWriteFile exception is thrown 

Note: It is created though if the path contains a prefix.

#### How to reproduce

```php
use League\Flysystem\Config;
use League\Flysystem\PhpseclibV3\SftpAdapter;
use League\Flysystem\PhpseclibV3\SftpConnectionProvider;

$provider = SftpConnectionProvider::fromArray([
    'driver' => 'sftp',
    'host' => 'domain',
    'username' => 'username',
    'password' => 'password',
]);


$adapter = new SftpAdapter(
    connectionProvider: $provider,
    root: 'uploads'
);

// Make sure that "uploads" folder is missing on remote server before each test

// ❌ Case 1: UnableToWriteFile exception is thrown
// I expect to work and file to be uploaded at /uploads/file.txt
$adapter->write('file.txt', 'content', new Config());


// ✅ Case 2: File is uploaded at /uploads/parent_folder/file.txt even if the "uploads" folder doesn't exist
$adapter->write('parent_folder/file.txt', 'content', new Config());

wakqasahmed:
Fixes #1832

## Summary

`SftpAdapter::write()` (both `PhpseclibV2` and `PhpseclibV3` variants) only creates missing directories via `ensureParentDirectoryExists()`, which computes `dirname($path)`. When writing directly to the adapter's root (e.g. `write('file.txt', ...)`), `dirname()` returns `'.'`, and the method returned early without creating anything — so if the configured root directory itself didn't exist on the remote server, the write failed with `UnableToWriteFile`.

Writing to a subdirectory (e.g. `write('parent_folder/file.txt', ...)`) worked fine because `makeDirectory()` is called with `mkdir(..., recursive: true)`, which implicitly creates the root along with the subdirectory.

### Fix

- Stored the configured `$root` as an adapter property.
- When `ensureParentDirectoryExists()` detects the write target is at adapter root (`dirname === '' || '.'`), it now also ensures the root directory itself exists (via the existing `makeDirectory()` path, reusing the same directory-visibility handling), matching the behavior subdirectories already had.
- Guarded on `$root !== ''` so adapters constructed with an empty root (i.e. no root restriction, using the connection's current working directory) keep their existing behavior and don't attempt to `is_dir('')`/`mkdir('')` against an empty path.
- Applied the same fix to both `League\Flysystem\PhpseclibV2\SftpAdapter` (deprecated but still shipped) and `League\Flysystem\PhpseclibV3\SftpAdapter` for consistency.

## Test plan

- Added `writing_a_file_creates_the_root_directory_when_it_is_missing()` to both `SftpAdapterTest` suites (`PhpseclibV2` and `PhpseclibV3`), following the existing test patterns in this suite (real SFTP connection via `StubSftpConnectionProvider`/`SftpStub` against the project's docker-based SFTP test server, `useAdapter()` for a one-off adapter instance, cleanup via `deleteDirectory()`).
- Test constructs an adapter pointed at a freshly-generated, not-yet-existing root path under `/upload`, asserts the root does not exist, writes a file directly to the adapter root, then asserts both the root directory and the file now exist.

**Disclosure: I could not run PHPUnit locally** — this environment has no PHP runtime available, so these tests are unverified locally. They're written to match the existing suite's conventions (same base class, same connection provider/stub, same `useAdapter`/cleanup pattern used elsewhere in this file) and should be validated by CI, which already has phpseclib3 and phpseclib and the SFTP docker test server configured for this test group (`@group sftp`).

### DIFF
--- a/src/PhpseclibV2/SftpAdapter.php
+++ b/src/PhpseclibV2/SftpAdapter.php
@@ -56,7 +56,7 @@ class SftpAdapter implements FilesystemAdapter
 
     public function __construct(
         ConnectionProvider $connectionProvider,
-        string $root,
+        private string $root,
         ?VisibilityConverter $visibilityConverter = null,
         ?MimeTypeDetector $mimeTypeDetector = null,
         private bool $detectMimeTypeUsingPath = false,
@@ -115,12 +115,17 @@ class SftpAdapter implements FilesystemAdapter
     {
         $parentDirectory = dirname($path);
 
+        /** @var string $visibility */
+        $visibility = $config->get(Config::OPTION_DIRECTORY_VISIBILITY);
+
         if ($parentDirectory === '' || $parentDirectory === '.') {
+            if ($this->root !== '') {
+                $this->makeDirectory('', $visibility);
+            }
+
             return;
         }
 
-        /** @var string $visibility */
-        $visibility = $config->get(Config::OPTION_DIRECTORY_VISIBILITY);
         $this->makeDirectory($parentDirectory, $visibility);
     }
 
--- a/src/PhpseclibV2/SftpAdapterTest.php
+++ b/src/PhpseclibV2/SftpAdapterTest.php
@@ -207,6 +207,26 @@ class SftpAdapterTest extends FilesystemAdapterTestCase
     /**
      * @test
      */
+    public function writing_a_file_creates_the_root_directory_when_it_is_missing(): void
+    {
+        $root = '/upload/missing-root-' . bin2hex(random_bytes(4));
+        $adapter = $this->useAdapter(new SftpAdapter(static::connectionProvider(), $root));
+
+        try {
+            $this->assertFalse($adapter->directoryExists(''), 'Precondition: root directory should not exist yet.');
+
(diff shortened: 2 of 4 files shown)


wakqasahmed:
Hi @frankdejonge — noticed this PR doesn't have a reviewer assigned yet — it's been about 4 days, CI is green and it's mergeable. Would you (or whoever's best placed) be able to take a look when you get a chance, or point me to who should? Thanks!

wakqasahmed:
Hi @frankdejonge — another follow-up on this one, 7 days in and still unassigned. Let me know if it needs anything from me.

frankdejonge:
Hi @wakqasahmed, I have not had a spot in my calendar to go over PRs. As you're probably aware, this is not a funded project. Any time is dedicated should be considered a gift, not something that people are entitled to.

